PR #59, opened 2015-01-22
Optimized code for BIP32 and BIP38

GeorgeKimionis:


NicolasDorier:
Seems good, however can you change the coding convention from _camel to _Upper ? Using it since I use it in vs code snippets.

Also,

> public ExtKey Derive(int index, bool hardened) 
{ 
if(nChild < 0) 
throw new ArgumentOutOfRangeException("index", "the index can't be negative"); 

is clearly a bug on my part, I wanted to compare index, not nChild.

GeorgeKimionis:
Since this is the notation used for code snippets we'll stick to that. The `index` glitch was fixed.

PR #61, opened 2015-01-29
Changes in Script.cs

GeorgeKimionis:
Fixed ToString() override.
Added null protection for lastOpcode in ToOps().
Optimized code.


NicolasDorier:
why the fuck did I wrote that 

>if(scriptHashParams != null)
  return scriptHashParams;
return null;

PR #67, opened 2015-03-24
Add unit test on private key creation from bytes.

bitcoinbrisbane:
Needed to add xunit nuget package too to my solution.

### DIFF
--- a/NBitcoin.Tests/NBitcoin.Tests.csproj
+++ b/NBitcoin.Tests/NBitcoin.Tests.csproj
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\packages\xunit.runner.visualstudio.2.0.0\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\packages\xunit.runner.visualstudio.2.0.0\build\net20\xunit.runner.visualstudio.props')" />
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


NicolasDorier:
cool, does this package means someone does not have to install the xunit test runner manually on their visual studio ?

DeathAndTaxes:
That is correct.  Nuget added support for installing tools and extensions via solution level nuget references.  Manually installing the vsix plugin is still an option (installed via Extensions in Tool menu) but the package will accomplish the same thing and with nuget package restore it makes it easier to get others on the same build quickly.

Two caveats:

* xunit recommends uninstalling the runner extension if using the nuget package.
> Important note: If you've previously installed the xUnit.net Visual Studio Runner VSIX (Extension), you must uninstall it first. The Visual Studio runner is only distributed via NuGet now. To remove it, to go Tools > Extensions and Updates. Scroll to the bottom of the list, and if xUnit.net is installed, uninstall it. This will force you to restart Visual Studio.

Despite that warning it still worked for me even without uninstalling but I believe leaving the extension manually installed prevents nuget from updating the runner when the nuget package (and reference in solution) is updated.  So pick either using the manually installed extension or 

* The package only has scope to the installed project. The manually installed vsix plugin gives you runner support for all projects but the nuget package only provides support for the project that references it.  I don't see this as much of an issue as getting xunit core support requires nuget package restore anyways but if you notice that after switching to the nuget package another solution (which uses xunit but doesn't have the xunit.runner.visualstudio package) doesn't have runner support then you will need to add the runner package to that solution as well.

Xunit seem to favor using nuget over the vsix extension.  
http://xunit.github.io/docs/running-tests-in-vs.html

BTW I am more familiar with NUnit, Nuget is recommended over manually installing the plugin for runner support in NUnit is now as well.

NicolasDorier:
ok that's cool, I merged it

bitcoinbrisbane:
As DeathAndTaxes said, Nuget is the way everything is going.  .Net next is all delivered by Nuget.

NicolasDorier:
Well, I know well about nuget, but I admit I was surprised to see a vs extension delivered as a nuget package, but I admit, this is awesome idea.

bitcoinbrisbane:
Its why open source / xp programming is great, learn new things all the time.

PR #69, opened 2015-04-14
Change default maximum OP_RETURN size to 80 bytes

satwo:
... to maintain compatibility with Bitcoin Core v0.10. See: https://github.com/bitcoin/bitcoin/pull/5286/files


NicolasDorier:
Well, I did that before, but realized that the 80 bytes is on the master but not on the 0.10, which you can see on the commit tagged 0.10:

[standard.h](https://github.com/bitcoin/bitcoin/blob/047a89831760ff124740fe9f58411d57ee087078/src/script/standard.h)

satwo:
Whoops! My fault for not realizing that. I'll withdraw the pull request.

PR #70, opened 2015-04-18

Glyco89:
Adding of GetUnspentAsync method to allow recovery of all unspent outputs belonging to a Bitcoin address via Blockr.io
Result is a dictionnary<TxIn, Money> object


Glyco89:
oups sorry, not the good one

PR #71, opened 2015-04-18

Glyco89:
Adding of GetUnspentAsync method to allow recovery of all unspent outputs belonging to a Bitcoin address via blockr.io
Result is a list of Coin

### DIFF
--- a/NBitcoin/BlockrTransactionRepository.cs
+++ b/NBitcoin/BlockrTransactionRepository.cs
@@ -82,8 +82,8 @@ namespace NBitcoin
 			return Task.FromResult(false);
 		}
 		
-		public async Task<Dictionary<TxIn, Money>> GetUnspentAsync(string Address)
-	        {
+		public async Task<List<Coin>> GetUnspentAsync(string Address)
+        	{
 	            while (true)
 	            {
 	                using (HttpClient client = new HttpClient())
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Thanks,
Can you stash the commits ? also the GetUnspent does not work on testnet right now, can you refactor the blockr url, so you use the same logic in both, GetUnspentAsync and GetAsync for building the url ?

Glyco89:
How is it possible to stash the commits on github? I am new to Git and I
didn't find it. I delete the repository and recreate a fork from yours but
the trace of the commits are still present...

"also the GetUnspent does not work on testnet right now" I found why: it
should be new Script(*Encoders.Hex.DecodeData*(element["script"].ToString()))
instead of new Script(element["script"].ToString()).

"can you refactor the blockr url, so you use the same logic in both,
GetUnspentAsync and GetAsync for building the url ?" yes I will do it.

Best,

NicolasDorier:
> also the GetUnspent does not work on testnet right now

I was refering to the fast you used http://btc.blockr.io/ even when network == Network.TestNet.

you should not have closed the pull request, I just wanted you update your branch to clean the commit messages.

take a look on [changing multiple commit message / Squashing](http://git-scm.com/book/it/v2/Git-Tools-Rewriting-History)

It is "git rebase -i HEAD~4"
The interactive mode is easy (VIM). Hit : "i" then you are in edit mode, edit as they explain on the link I gave you. Once you have finished hit "ESCAPE, then :wq ENTER".

Then edit commit messages if you want, if you don't "ESCAPE, then :wq ENTER".

PR #78, opened 2015-05-27
Replace Mono.Nat with Open.NAT

lontivero:
@NicolasDorier

I removed the Mono.Nat code and replaced it with the Open.NAT nuget package. Could you take a look at that? You will see I deleted a lot of code from UPnPLease class because Open.NAT already implements what was implemented by you in UPnPLease.cs

Let me know if you find any error, problem or suggestion.


NicolasDorier:
Is it possible to fix the indentation so the diff is easier to follow ? More important, you are using `await` without  `.ConfigureAwait(false)` this has the result to make deadlock if `.Result` is used in a ASP.NET or WPF context.

NicolasDorier:
You forgot some Await, no worry, I'll fix that. I'm testing it.

NicolasDorier:
as we talked about, I should remove dependency to Mono.Nat and let people decide for themselves what library to take. I close this PR, not because it does not work, but because I agreed with you that UPNP lib have not its place in a Bitcoin Library.

PR #80, opened 2015-05-29
Money.TryParse using decimal.TryParse

lontivero:
This PR changes the internal implementation details of Money.TryParse. It uses decimal.TryParse for persing string containing the amount. 

Important:
* it uses **InvariantCulture** in order to prevent the usage of numbers formatted with specific culture  (as before).
* it **prevents 63 bit overflow** for possitive/negative values allowed (as before).
* leading and trailing spaces are allowed (as before)

New:
* it allows now numbers like .5 and -.5

Tests passed.

### DIFF
--- a/NBitcoin/Money.cs
+++ b/NBitcoin/Money.cs
@@ -1,10 +1,7 @@
-﻿using NBitcoin.DataEncoders;
+﻿using System.Globalization;
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Numerics;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace NBitcoin
 {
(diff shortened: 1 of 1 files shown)


NicolasDorier:
sounds good to me, have you checked the tests pass ? there is lots of test on Money parsing and formatting, I think they are broken now. (we might have to change the tests)
Should be in the "Core" trait.

Have you also checked that it compile portable libraries ?

NicolasDorier:
also can we have a test for the overflow case ? (if you don't have time, I can do it)

lontivero:
> There is lots of test on Money parsing and formatting, I think they are broken now. (we might have to change the tests).

I used the existing UT to be sure the behaviour is the same that the old implementation. we don't have to change the UTs. 

> Also can we have a test for the overflow case ?

Can you tell me more about this test? I don't fully understand. 

> Have you also checked that it compile portable libraries?

No, I didn't check that.

NicolasDorier:
> No, I didn't check that

Don't worry, I'll do.

for the overflow case, I mean a test going into

> if (Math.Abs(integralPart) > maxAllowedValue)

But don't worry, I'll check that.

NicolasDorier:
I've changed it a bit, so I check overflow also in the decimal ctor of Money.

lontivero:
Well, I didn't want to change the previous code behaviour in any way and for that reason I did it as I did it (Even when it was clearly wrong). 

The number 9223372036854775807 is the value of long.MaxValue. It is no the same for long.MinValue. So, the code should be just a bit different:

Replace this:
```
return Math.Abs(satoshis) <= 9223372036854775807m;
```

By this:
```
return long.MinValue >= satoshies &&  satoshies <= long.MaxValue;
```

NicolasDorier:
nice catch fixing it

NicolasDorier:
well, actually the code was correct. Interestingly.

```
9223372036854775808 == 9223372036854775808
```
BUT
```
-(-9223372036854775808) = -9223372036854775808 
```
WTF of the day for me

lontivero:
That´s because by default math operation are unchecked, as you know. Then:

```
unchecked(-(-9223372036854775808) == -9223372036854775808) // True
  checked(-(-9223372036854775808) == -9223372036854775808) // OverflowException
```

long.MaxValue = 9223372036854775807
binary = 0111111111111111111111111111111111111111111111111111111111111111

long.MaxValue + 1 =  9223372036854775807 + 1 = -9223372036854775808 = long.MinValue (overflow)
binaty = 1111111111111111111111111111111111111111111111111111111111111111

-long.MinValue =  -1 * -9223372036854775808 = 9223372036854775808 =  long.MaxValue + 1 (overflow again) = -9223372036854775808
binaty = 1111111111111111111111111111111111111111111111111111111111111111

So, yes, it is right
-(-9223372036854775808) == -9223372036854775808 // this is correct because the overflows

NicolasDorier:
thanks for the explanation, now Money is safe from overflow ! :)

lontivero:
Wait a minute. It is not safe yet. You need to check in Money(ulong) because there is a problem there. In fact, take a look at the following test:

	[Fact]
	public void Overflow()
	{
		var money = new Money(9223372036854775808uL);
		var str = money.ToString(); // --92233720368.-54775808
	}

I´m going to continue testing this class bacause I think that even when is not a complex class, it is an important one.

lontivero:
@NicolasDorier I am going to send a PR with these fixes. Operator with doubles and ulongs are not safe neither. 

Also, I have a fix for the inherited code in Money.ToString() method that, as I show you (--92233720368.-54775808), has a problem too.

Please wait for my PR.

NicolasDorier:
can you take a look to my recent commits ? I fixed the -- problem by forbidding long.MinValue

lontivero:
Some other findings. Money loses money in division operation. These methods perform integer division!

    public static Money operator /(long left, Money right)
    {
        return Money.Satoshis(checked(left / right._Satoshis));
    }
    public static Money operator /(Money right, long left)
    {
        return Money.Satoshis(checked(right._Satoshis / left));
    }

look this test:

    [Fact]
    public void IntegerDivision()
    {
        var total = new Money(1.0m, MoneyUnit.BTC);
        var op = (total/300000) * 300000;
        Assert.Equal(total.Satoshi, op.Satoshi);
    }

>Test 'NBitcoin.Tests.util_tests.IntegerDivision' failed: Assert.Equal() Failure
>Expected: 100000000
>Actual:   99900000
>	util_tests.cs(282,0): at NBitcoin.Tests.util_tests.IntegerDivision()

100.000 satoshis lost.

lontivero:
The Movey class should be inmutable, in fact, every single operation return a new instance (what is correct) however it provides operators that change its internal state:

* public static Money operator ++(Money left) and,
* public static Money operator --(Money left)

The following test fails:

    [Fact]
    public void Overflow()
    {
        var money = new Money(long.MinValue+1);
        Assert.Throws<OverflowException>(() => money--);
    }

There are two problem here:

* It doesn't throw OverflowException but even more important, 
* money's instance internal state is invalid (long.MinValue)

I will fix these bugs but I really believe that money class has some operators that make no sense and are dangerous for different reasons: /, ++, --

PR #82, opened 2015-06-02
Change encoders and uint128/256

lontivero:
@NicolasDorier, this PR contains changes in:

## Encoders

* EncodeData method signature changed. Basically, every time we deal with arrays we can see the same pattern all around the .NET libraries: (byte[] data, int offset, int count). That pattern is correct and flexible.
 
```
string EncodeData(byte[] data, int length)  // old version
string EncodeData(byte[] data, int offset, int count) // new version
```

### ASCIIEncoder 

New implementation uses the .NET _Encoding.ASCII_ class ( _GetString_ / _GetBytes_ methods)

### Base58Encoder

New implementation uses _Buffer.BlockCopy_ to manipulate (extract/concat) arrays of bytes instead of Linq. I performed a simple test and the improvement in performance is good (31ms encoding 10 arrays[40])

###   Base64Encoder

Tiny change (only the method signature)

### HexEncoder

I simplified the encoding code using a lookup table what made it easier to understand/mantain and it performs very well. I didn't measured it. However, given it uses a lookup table, it should perform better than the previous algorithm.

## unit128 / uint256

I refactored them in order to eliminate duplicated code. Also, all byte-per-byte operations were replaced by uint32 operations (simplicity and performance). As a result, these classes are ~50% smaller than before.

## Others

* I fixed a broken test. It wasn't failing but it wasn't really testing neither.  

## Tests

All existing tests are green.


NicolasDorier:
I will test that.
However, I don't want to use the StringBuilder. I did in the past, but since EncodeData is often used, I runned into performance and memory problems which resulted into a very busy GC.

lontivero:
Ok, don't worry about the StringBuilder issue, I can test the performance/memory behaiour and a make a benchmark for both the original and the new implementations if you want.

NicolasDorier:
for the StringBuilder part ? you can do it, but I am very sure about what I am saying.

lontivero:
np. I've removed the **StringBuilder**. The resulting algorithm looks very very similar to the previous one but I think it is an improvement because **the original was O(n^2)** and **the new one is O(n)**. 

That is because IsDigit() uses an IndexOf twice for each byte.

NicolasDorier:
Not sure I understand what you mean, commit 2a3daff7 was using a StringBuilder initially, as far as I can see, as you are doing. I don't see it is O(n^2)

lontivero:
IsDigit uses IndexOf that is O(n/2).
IsDigit is invoked twice for each byte: 2n O(n/2) = O(n^2)

PR #84, opened 2015-06-03
Minor cleanup & Fix code 

lontivero:
@NicolasDorier, this PR has minor changes similar to the previous once and also fix a problem in Utils.ToUnit32(). If you run the UTs (before merging this PR) you will see several tests in red.

By the way, if you want to have the flexibility to deal with system endianness, you should consider [Mono.DataConvert](http://www.mono-project.com/archived/mono_dataconvert/) (is not a library, it's a standalone class you can include in your projects). Mono,DataConvert is a very good piece of code really.


NicolasDorier:
sorry, I fixed that after I woke up with a mail from my CI server, just saw you PR :/ next time I'll check out github instead of my mail first :p

PR #85, opened 2015-06-07
SafeSubarray method and minor changes

lontivero:
@NicolasDorier this PR has a new method called SafeSubarray. It is used as follow:

Instead of:
```
var newByteArray = sourceByteArray.Skip(n).Take(count).ToArray();
```

we can use:
```
var newByteArray = sourceByteArray.SafeSubarray(n, count);
```


lontivero:
The latest commit fixes the problems discussed in the code review.

PR #87, opened 2015-06-11
MoneyBag first version

lontivero:
@NicolasDorier, this PR is for code review and discussion.


NicolasDorier:
I am not for the "IsZero", it makes the interface more complex for nothing. (`var zero = result.Sub(result)` gives you zero for all type of money)

Except that it seems good.

I will modify the TransctionBuilder.Send so it understands what a MoneyBag is.

PR #88, opened 2015-06-11
Validations

lontivero:
@NicolasDorier this PR has minor changes as:

* null validations for public methods, 
* static modifier for those classes that only have static methods, 
* static modifier for methods that can be statics, 
* remove unused local variables

But also contains some **comments and warnings** that you should review. There are methods with **parameters that are not used in the method body** and that could be important (I don't know if their values should be used). Please take a look at the //FIXME comments.

I've splitted the PR in a couple of commits because I think in that way it can be easier for reviewing. 

**NOTE:** the PR compiles and all tests pass GREEN. However, individual commits do not!

### DIFF
--- a/NBitcoin.Tests/StealthAddressTests.cs
+++ b/NBitcoin.Tests/StealthAddressTests.cs
@@ -122,7 +122,7 @@ namespace NBitcoin.Tests
 				Transaction transaction = new Transaction();
 				transaction.FromBytes(TestUtils.ParseHex(test.Transaction));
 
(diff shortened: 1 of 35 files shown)


NicolasDorier:
with your changes I have seen 2 important bug on SetAddrman et BitField.GetPayment that I have just fixed. Can you rebase ?

NicolasDorier:
kPar is not used so you can remove it

lontivero:
There is a new PR #89 with these changes.

PR #90, opened 2015-06-12
Change default maximum OP_RETURN size to 80 bytes

lontivero:
@NicolasDorier, the change in this PR is correct.

### DIFF
--- a/NBitcoin.Tests/transaction_tests.cs
+++ b/NBitcoin.Tests/transaction_tests.cs
@@ -1347,12 +1347,12 @@ namespace NBitcoin.Tests
 			t.Outputs[0].ScriptPubKey = new Script() + OpcodeType.OP_1;
 			Assert.True(!StandardScripts.IsStandardTransaction(t));
 
-			// 40-byte TX_NULL_DATA (standard)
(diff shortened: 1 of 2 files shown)


NicolasDorier:
The limit is always 40 bytes in 0.11, even if master says it is not. Look at the tagged version.

lontivero:
It was introduced in the [bitcoin core 0.11 RC - file: /src/script/standard.h](https://github.com/bitcoin/bitcoin/blob/v0.11.0rc1/src/script/standard.h) 

```
static const unsigned int MAX_OP_RETURN_RELAY = 80;      //! bytes
```

As you well said, it is not 80 bytes (yet). Let the PR opened

NicolasDorier:
good to know, also I don't know if it is a good idea to move it to 80 bytes before we get a majority of 0.11 nodes. What do you think ?

lontivero:
NBitcoin won't break backward compatibility with this change and at the same time it will be able to accept transactions broadcasted by clients which take advantage of the new script max size.

I think that, given this is a protocol change, NBitcoin should be updated to understand the new protocol versions, by going hand in hand with the latest official bitcoin core release version. (no release candidate but final releases).

IMO you should do the same that the bitcoin core team did: take this change wait for the official bitcoin realease. Once the 0.11 release is publish, you could release a new NBitcoin version too.

NicolasDorier:
ok so I let this PR open until then ?

lontivero:
Yes, IMO that's ok.

NicolasDorier:
0.11 is released

PR #91, opened 2015-06-13
Base58 tests

lontivero:
@NicolasDorier,

Look, the tests look very good. The new tests are much more clear.

This is just to know if you like this style.


NicolasDorier:
well, if there is not lots of data it makes sense to move the data in the tests.
I am not fan of TheoryAttribute, because I think we can do a dumb foreach ourselves, but don't oppose it.

But if we take this approach we should also take it for "Invalid" test, and not mix both, so can you remove the json file for invalid base58 like you was doing ?

lontivero:
As you can see in the first commit, there is a big chunck of data for "InvalidKey" and that didn't like me really, that's why I changed it in my second commit. 

IMO, test using TheoryAttribute are much clearer that those that use a foreach because I don't have to iterate in my mind to understand them. Look:

		public static IEnumerable<object[]> InvalidKeys
		{
			get
			{
				var dataset = TestCase.read_json("data/base58_keys_invalid.json");
				return dataset.Select(x => x.ToArray());
			}
		}

		[Theory, PropertyData("InvalidKeys")]
		public void base58_keys_invalid(string data)
		{
			// must be invalid as public and as private key
			Assert.Throws<FormatException>(() => Network.Main.CreateBitcoinAddress(data));
			Assert.Throws<FormatException>(() => Network.Main.CreateBitcoinSecret(data));
		}

And now, take a look at the previous one:

		[Fact]
		[Trait("Core", "Core")]
		public void base58_keys_invalid()
		{

			var tests = TestCase.read_json("data/base58_keys_invalid.json"); // Negative testcases

			foreach(var test in tests)
			{
				string strTest = tests.ToString();
				if(test.Count < 1) // Allow for extra stuff (useful for comments)
				{
					Assert.False(true, "Bad test: " + strTest);
					continue;
				}
				string exp_base58string = (string)test[0];

				// must be invalid as public and as private key
				Assert.Throws<FormatException>(() => Network.Main.CreateBitcoinAddress(exp_base58string));
				Assert.Throws<FormatException>(() => Network.Main.CreateBitcoinSecret(exp_base58string));
			}
		}

I know there are more important things in what someone can contribute but anyway, tests are the first place where devs go in order to understand how something work and that's why I think that clearer tests worth the effort.

PR #93, opened 2015-06-15
Using BufferedStream for socket communication

lontivero:
Take a look at these changes. The Buffer size needs to be tunned better, 50kb is the best value for sure. (Only for discussion)

### DIFF
--- a/NBitcoin.Tests/base58_tests.cs
+++ b/NBitcoin.Tests/base58_tests.cs
@@ -1,4 +1,5 @@
-﻿using NBitcoin.DataEncoders;
+﻿using System.Diagnostics;
+using NBitcoin.DataEncoders;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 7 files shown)


NicolasDorier:
I've benchmarked it, I don't see any obvious performance improvements, so I would prefer we don't use it.
But I like the fact that ReadNext do not depends on Socket anymore.
Is it possible to remove the BufferedStream but keep the rest ?

image

lontivero:
I've remove the BufferedStream and include changes in Hex and Base58 encoders. Also, Message.ReadNext() is overloaded with Stream and Socket.

NicolasDorier:
Can you remove the Hex and Base58 part for now ?
I'm bothered by the fact that Hex now require a list, this class is used a lot for intensive indexing applications, and would not require lots of overhead.
Hex encoder is used an order of magnitude more than base58, because it is the most common way to serialize any bitcoin structure.

Maybe we should not change it after all.

lontivero:
Yes, I can. But let me do a benchmark first.

lontivero:
Well, you were right about the hex encoder performance. The problem was not because of the List but the extra comparisons instead.

NicolasDorier:
and also that accessing a list with an index is O(n)

PR #95, opened 2015-06-20
Install missing package

bitcoinbrisbane:
Not buildable, needs xunit.runner.visualstudio


NicolasDorier:
I think you are refering to an old version because the current packages.config already have xunit.runner.visualstudio, can you check ?

lontivero:
Confirmed. This comes from bitcoinbrisbane:master which comes from NicolasDorier:master  f021280458c646f3b8c74b8b83a05d2c1b3a8563 so, yes it is an old. The reference was included in ea1acdbd2a1e90d50ca0a8a1ddea3e9fdba1403a as part of the **"Update Sqlite" commit**

bitcoinbrisbane:
Sorry guys, new VM and didn't sync properly.

PR #100, opened 2015-06-21
minor cleanup

lontivero:
@NicolasDorier this PR is for removing a couple of redundant comparisons and assignments.

### DIFF
--- a/NBitcoin/BIP39/Mnemonic.cs
+++ b/NBitcoin/BIP39/Mnemonic.cs
@@ -176,7 +176,7 @@ namespace NBitcoin
 		public byte[] DeriveSeed(string passphrase = null)
 		{
 			passphrase = passphrase ?? "";
-			var salt = Concat(UTF8Encoding.UTF8.GetBytes("mnemonic"), Normalize(passphrase));
+			var salt = Concat(Encoding.UTF8.GetBytes("mnemonic"), Normalize(passphrase));
 			var bytes = Normalize(_Mnemonic);
 
 #if !USEBC
--- a/NBitcoin/BitcoinCore/BlockStore.cs
+++ b/NBitcoin/BitcoinCore/BlockStore.cs
@@ -62,6 +62,7 @@ namespace NBitcoin.BitcoinCore
 
 		[ThreadStatic]
 		bool headerOnly;
+		// FIXME: this methods doesn't have a path to stop the recursion.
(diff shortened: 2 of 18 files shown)


NicolasDorier:
I fixed the Target code, can you revert just this one ? also for the ThreadStatic, I'm not eager to remove it, since I did it for a reason. (which I don't remember though)

NicolasDorier:
also the not null checks in Wallet.cs are here for a purpose, we should not remove it.

lontivero:
Done. I feel uncomfortable with some pieces of code that I don't understand. For example, ThreadStatic doesn't work for instance fields. What is the reason for having it?

The same sensation I have with the null checks in wallet. I mean, we assign a value and two lineas bellow we check for null (!?) I don't understand. Even if there is a reason for that, it should be explained in a comment at least.

NicolasDorier:
ThreadStatic does not work for instance field ? I'm surprised about that.

For != null checks, I do that because the user can pass null to the method you were modifying. (For exemple if he wants to specify a specific chain, but not the addrman.

The Connect afterward populate with default values if not provided.

NicolasDorier:
Update... you are right on the Connect method indeed. 

In fact this code
```
addrman = addrman ?? new AddressManager();
chain = chain ?? new ConcurrentChain(_Parameters.Network);
tracker = tracker ?? new Tracker();
```
Should not exist. We should remove that, since the other overload of Connect is responsible for providing default values.

lontivero:
> ThreadStatic does not work for instance field ? I'm surprised about that.

Yes, look at it by yourself [here](https://msdn.microsoft.com/en-us/library/system.threadstaticattribute(v=vs.110).aspx). It says clerly: _Indicates that the value of a **static field** is unique for each thread._ This attribute has no effect in your code. Let me know how to proceed, please.

I will remove this code:

```c#
addrman = addrman ?? new AddressManager();
chain = chain ?? new ConcurrentChain(_Parameters.Network);
tracker = tracker ?? new Tracker();
```

lontivero:
Ok, I removed the assignments and didn't remove the ThreadStaticAttribute (please do it if you don't find a reason).

NicolasDorier:
so you were right, we should delete it, I learned something.

lontivero:
Are you going to merge this PR?

NicolasDorier:
yes sorry, it took time because I just arrived in japan.

PR #101, opened 2015-06-25
Protected constructors for abstract classes

lontivero:
@NicolasDorier This PR is for changing the access modifiers in abstract classes' constructors. Also, I implemented the IBlockRepository in the BlockRepository class (at least the names sound very similar).


NicolasDorier:
can you just remove the `using System.IO.MemoryMappedFiles` so the build does not break in portable ?

lontivero:
Done.

PR #104, opened 2015-06-26
ParseJson & ParseHex

lontivero:
When we have to create an object instance from its string representation, the .NET Framework is very consistent, it provides a Parse static method. NBitcoin provides Parse methods in many cases and also contructors with string parameters in other clases.

```UInt32.Parse("12")``` is the .NET Fx convention. ```new UInt32("12")``` is not.  

Now, in some cases we shouldn't call *Parse* to those methods that receive a string as parameter because that could be confusing. In code:

```UInt32.Parse("12345")   // this is ok```
```UInt256.Parse("12345") // fails even when 12345 is a valid uint256 value (base10)```

That's why in some cases the Parse method should be called differently. In this case (uint256) it should be called **ParseHex**.

```UInt256.ParseHex("12345") // now it is clear that 12345 is not a valid hexadecimal value```

Something similar happens with other Parse methods that receive strings but is not clear what the expected string representation is. For example **Block.Parse** which expects a json representation and not a hex representation. Look:

```BlockHeader.Parse() // this method expects a hex representation```
```Block.Parse() // but this method expects a json. it´s not consistent.```
 
In this cases it should be called **ParseJson**.

This PR doesn't break the existing NBitcoin contracts but uses ```ObsoleteAttribute``` in order to let the developer know that the used method could be remove in the future.

### DIFF
--- a/NBitcoin/RPC/RawFormatter.cs
+++ b/NBitcoin/RPC/RawFormatter.cs
@@ -1,4 +1,5 @@
-﻿using Newtonsoft.Json;
+﻿using System;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using System.IO;
 
@@ -15,6 +16,13 @@ namespace NBitcoin.RPC
 			get;
 			set;
 		}
+		public Transaction ParseJson(string str)
+		{
+			JObject obj = JObject.Parse(str);
+			return Parse(obj);
+		}
+
+		[Obsolete("Use RawFormatter.ParseJson method instead")]
 		public Transaction Parse(string str)
 		{
 			JObject obj = JObject.Parse(str);
--- a/NBitcoin.Tests/ColoredCoinsTests.cs
+++ b/NBitcoin.Tests/ColoredCoinsTests.cs
@@ -29,7 +29,7 @@ namespace NBitcoin.Tests
 					var txObj = new Transaction(tx);
 					repository.Put(txObj.GetHash(), txObj);
 				}
-				TestedTxId = uint256.ParseHex(testcase.testedtx);
+				TestedTxId = uint256.Parse(testcase.testedtx);
 				Repository = new NoSqlColoredTransactionRepository(repository, new InMemoryNoSqlRepository());
 			}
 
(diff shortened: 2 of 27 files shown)

### CODE REVIEW

**Comment by NicolasDorier** [comment]
I do not agree with `ParseHex` I think a simple comment in the <summary> of Parse to say it is expecting hex is enough.

However, Block.Parse` method should not parse JSON. I propose we change the implementation to parse hex instead. I bet no person is using `Block.Parse` for json anyway as it is now.

I'd like to keep the string constructor, because it makes the library more discoverable.

**Comment by lontivero** [comment]
* Well, I understand you don't like ```ParseHex```. In fact, I don't like it neither.  **Agree**.
* ```Block.Parser``` should Parse Hex instead of json. **Agree**.
* About *string constructors*. Are you really sure? .NET programmers don't expect those constructors because they have never seeing them before in the .NET world. I mean, everybody knows that DateTime doesn't have a string constructor.

```new DateTime("10/10/2016")``` but a ```DateTime.Parse("10/10/2016")``` and the same is valid for every class in the BCL. Why should NBitcoin be different? I cannot agree wit you here. Please think about it.

**Comment by NicolasDorier** [comment]
Ok, so let's keep the ctor, but with Obsolete.

File: NBitcoin/RPC/RawFormatter.cs [hunk 1]
```diff
@@ -15,6 +16,13 @@ public Network Network
 			get;
 			set;
 		}
+		public Transaction ParseJson(string str)
+		{
+			JObject obj = JObject.Parse(str);
+			return Parse(obj);
+		}
+
+		[Obsolete("Use RawFormatter.ParseJson method instead")]
```
lontivero: Transaction.Parse should expect a hex string too.

lontivero:
Done.

NicolasDorier:
if I merge that, do we have warning about using obsolete methods ? If so I think there is an attribute to disable them in for Parse methods.

NicolasDorier:
And microchannelargument can just be called Parse and not ParseJson, because there is no choice a all.

lontivero:
Well, NBitcoin has no compilation warnings now. Those who use it will be alerted with a warning that let them know they should use the Parse method instead of the constructor. In fact, that's the ObsoleteAttibute attribute goal.

PR #106, opened 2015-06-27
Data Packer / Unpacker (PoC)

lontivero:
This PR is for discussing about the Data Packer / Unpacker feature.

The idea behind this classes is:
+ Improve the code legibility and,
+ Stop dealing with endianess all around the code base.

For example:
```c#
var bytes = hash1.ToBytes(false).Contact(new byte[]{0x14}).Contact(Utils.GetBytes(123u)).Concat(hash2.ToBytes(true)).ToArray();
```
With Data Packer
```c#
var bytes = Packer.Pack("^X_bLX", hash1, 0x14, 123u, hash2);
``` 

In the previous example I used the following commands:
+ **X** for IBitcoinSerializable objects
+ **A** for byte arrays
+ **b** for byte
+ **L** for unsigned long
+ **^** for switch to bigendian
+ **_** for switch to little endian

Let me know if you think this worth the effort.


NicolasDorier:
I don't think it makes things easier to read, and does not fix the problem of endianess. 
Passing endianess in a string instead of as argument does not make the problem disappear.

PR #108, opened 2015-06-29
REST Client 

lontivero:
@NicolasDorier, it is not finished yet even when it works except for the GetUTXOs method.


lontivero:
@NicolasDorier this PR has the REST Client implementation. You need bitcoind/qt 0.11rc1 to test it.

PR #110, opened 2015-07-01
RPC AddNode, GetAddedNodeInfo, BackupWallet and EstimatePriority

lontivero:
@NicolasDorier, I'm sending the PR for review and discussion.

### DIFF
--- a/NBitcoin.Tests/RPCClientTests.cs
+++ b/NBitcoin.Tests/RPCClientTests.cs
@@ -16,11 +16,11 @@ namespace NBitcoin.Tests
 	//Require a rpc server on test network running on default port with -rpcuser=NBitcoin -rpcpassword=NBitcoinPassword
 	//For me : 
 	//"bitcoin-qt.exe" -testnet -server -rpcuser=NBitcoin -rpcpassword=NBitcoinPassword 
+	[Trait("RPCClient", "RPCClient")]
 	public class RPCClientTests
 	{
 		const string TestAccount = "NBitcoin.RPCClientTests";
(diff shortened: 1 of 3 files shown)


NicolasDorier:
This is a good idea, but should have the async versions too.

Also, the IpEndpoint class is not portable. For RestClient, I only used `#if !NOSOCKET` to disable any code coupled with IpEndpoint. But it sucks to delete code, just because IpEndpoint is not available in portable library. I'm not sure about the best way to handle this problem.

I can't test for now, my node is indexing transactions :-S

lontivero:
In the latest commit I moved code for grouping it in code regions. I did it to have a better idea about the  implemented and missing RPC methods. I know this makes the review harder but on the other hand the code is easier to find and work with.

NicolasDorier:
Can you wrap `return GetAddedNodeInfoAsync(detailed).Result;` correctly so we get back a real exception and not a generic aggregate exception ? I did it several time in RPCClient, and RestClient you should find an example.

lontivero:
@NicolasDorier, this PR is done since yesterday. Could you review it?

NicolasDorier:
trait should be [Trait("RPCClient", "RPCClient")], I am using it for my build server.

Some .Result are still not wrapped (EstimatePriority)

lontivero:
Done.

NicolasDorier:
last nit, you use %temp% in the test, can you just use the relative path (which will resolve to the output dir of the test project) ?

lontivero:
Sure, no problem. It is an important point.

PR #111, opened 2015-07-03
Reject transactions with excessive numbers of sigops

lontivero:
See:  bitcoin/bitcoin@4fad8e6


NicolasDorier:
wtf ? 4 sig op max per Transaction ? It basically means you can't have more than 4 input in a transaction ???

NicolasDorier:
ha no sorry

PR #112, opened 2015-07-04
Add consensus library

lontivero:
@NicolasDorier, this PR includes some temporal changes for using the **bitcoin core consensus library** in NBitcoin.

It seems to work and **tests pass green except for two failing tests** (multisign tests):
```
------ Test started: Assembly: NBitcoin.Tests.exe ------

Test 'NBitcoin.Tests.script_tests.script_CHECKMULTISIG12' failed: Assert.True() Failure
	script_tests.cs(421,0): at NBitcoin.Tests.script_tests.script_CHECKMULTISIG12()

Test 'NBitcoin.Tests.script_tests.script_CHECKMULTISIG23' failed: Assert.True() Failure
	script_tests.cs(464,0): at NBitcoin.Tests.script_tests.script_CHECKMULTISIG23()

15 passed, 2 failed, 0 skipped, took 4,71 seconds (xUnit.net 1.9.2 build 1705).
```

I don't fully understand the multisignature logic yet but I will continue trying to find the problem. Could you take a look at this code and let me know if there is some obvious mistake, please? 

**IMPORTANT**: you need to do:

+ copy the _libbitcoinconsensus-0.dll_ to the Test Output Folder and,
+ be sure the target platform is x64 (because the consensus library is x64)


lontivero:
Forget it. I get it.

lontivero:
I have to say something stupid: the concensus library works perfectly well ;)

Currently all script_tests pass green.

```
------ Test started: Assembly: NBitcoin.Tests.exe ------

17 passed, 0 failed, 0 skipped, took 4,41 seconds (xUnit.net 1.9.2 build 1705).
```

Of course we have to do some very few and minor changes in the VerifyScript methods, move some lines of code to a better pace among others but it feel awesome to see the tests green.

lontivero:
With my latest commit the integration is done and tested with and without using the bitcoin consensus library (BCL). By the way, with the BCL tests run almost 300% faster. Look my previous comment (with BCL) and these times without the BCL:
```
------ Test started: Assembly: NBitcoin.Tests.exe ------

17 passed, 0 failed, 0 skipped, took 11,56 seconds (xUnit.net 1.9.2 build 1705).
```

NicolasDorier:
When you define NOCONSENSUSLIBRARY in project file you must separate by using ';' not ','.

I don't understand, why the test were broken ? they pass on master.
Also, I want that code that are actually using the NBitcoin Script interpreter continue to do it as before.

We will only add different methods to use the lib consensus check.

lontivero:
> When you define NOCONSENSUSLIBRARY in project file you must separate by using ';' not ','.

No problem. I will fix it.

> I don't understand, why the test were broken ? they pass on master.

They are not broken. Once I tested with the BCL they failed because the scriptSig were not in the transaction input and the BCL uses takes the scriptSig from txTo.Inputs[txIn].ScriptSig. I mean, if you look ate the VerifyScript parameters, it requires the scriptPublicKey, toTx and nIN but doesn't requires the scriptSig because it takes it from the transaction input.

>Also, I want that code that are actually using the NBitcoin Script interpreter continue to do it as before.

Ok, just to undertand. Why? I mean, what you did is fantastic but don't you think the BCL can provide a much easier way to be on sync with the protocol rules to achieve consensus? Next time something change in the script verification, a NBitcoin user could just to download the latest libbitcoinconsensus-x.dll and replace the previous one with the latest one; otherwise he has to wait for an update in NBitcoin. 

> We will only add different methods to use the lib consensus check.

Ok, I can do that. But I have a question:  which Script.VerifyScript method should be used by the Transaction.VerifyScript method?

NicolasDorier:
> Ok, just to undertand. Why? I mean, what you did is fantastic but don't you think the BCL can provide a much easier way to be on sync with the protocol rules to achieve consensus?

Because I don't want to impose to people deploying the dll. Deploying a native dll with an app is a major headache because the dll to use depends on the bitness (32 or 64 or other) you are using.
Also the lib consensus can't be portable, while my implementation is.
Also my implementation send detailed result about why a script don't pass, while lib consensus is more general.

I think for heavy intensive application (or full node implementation), it makes sense for them to choose lib consensus though. For others, it is more an hassle.

> Ok, I can do that. But I have a question: which Script.VerifyScript method should be used by the Transaction.VerifyScript method?

The one that it used before. (my implementation)
We can add a method VerifyScriptConsenus()... for the consensus lib version.

>  Once I tested with the BCL they failed because the scriptSig were not in the transaction input and the BCL takes the scriptSig from txTo.Inputs[txIn].ScriptSig.

Ok I understand, indeed it was a problem.

NicolasDorier:
I forget to say : the data driven tests script_valid and script_invalid should also try the consensus implementation. (I can do it in another commit thhough)

lontivero:
I added the ```Script.VerifyScriptConsensus``` method. Status: ```script_valid```, ```script_CHECKMULTISIG12``` and ```script_CHECKMULTISIG23``` pass green with both algorithm (yours and BCL). ```script_invalid``` is **RED** because the test \#463 seems to be valid for the consensus library. That's very wierd because in my previous commit all tests were green.

image

You can see I updated the script_invalid.json file. I did that just to be sure the test cases are correct. The point is the ```script_invalid``` test fails with the same error: **P2PK with unnecessary input** with both, the previous script_invalid.json file and the new one. 

NOTE: XUnit Test Explorer cannot discover/run tests with platform other than AnyCPU and x86. So, if you want to run the tests with xUnit Test Explorer, you have to use the 32bits version of libbitcoinconsensus-0.dll .

Could you please review this PR? Thanks

NicolasDorier:
Seems good to me.
I will download libconsensus should be downloaded by the tests automatically, and SHA checksum checked.

PR #115, opened 2015-07-10
ContextStack as hybrid stack

lontivero:
This PR is for improving the ```ScriptEvaluationContext```'s stack operations, because the ```System.Collections.Generic.Stack``` class didn't fit very well. Basically, it is necessary to perform some operations that are not provided by the Stack class:

* Insert, 
* Swap,
* Erase (renamed as Remove) and,
* Top

This PR includes a new ```ContextStack``` stack class that provides these missing methods and operates in the internal array. 
 
For example:

```c#
    swap(ref _Stack, -3, -2);   ---> _stack.Swap(-3, -2);
``` 

As a side product, there is a very small gain in performance.

Please @NicolasDorier, take a look at these changes and give me some feedback.

Ah! there was a broken test in the RestClientTests because someone spent an UTXO that I used in the test (now, the test uses an UTXO that is mine and that I won't spend)

### DIFF
--- a/NBitcoin.Tests/AssertEx.cs
+++ b/NBitcoin.Tests/AssertEx.cs
@@ -35,7 +35,7 @@ namespace NBitcoin.Tests
 		}
 
 		[DebuggerHidden]
-		internal static void StackEquals(Stack<byte[]> stack1, Stack<byte[]> stack2)
+		internal static void StackEquals(ContextStack<byte[]> stack1, ContextStack<byte[]> stack2)
 		{
 			var hash1 = stack1.Select(o => Hashes.Hash256(o)).ToArray();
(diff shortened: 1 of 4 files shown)


NicolasDorier:
Not sure DebuggerTypeProxy will not crash on other plateform.

You also wrote "tipically" instead of "typically".

The PR messed up indentation which make it hard to review, but it seems right.

lontivero:
> Not sure DebuggerTypeProxy will not crash on other plateform.

I have no the answer but I think it shouldn't. I removed it from the PR.

> You also wrote "tipically" instead of "typically".

Fixed

> The PR messed up indentation which make it hard to review, but it seems right.

I know. I made so many changes and then rolled them back that the code changed a lot and messed it up. I'm sorry for that. But the changes are really a few.

PR #119, opened 2015-07-18
some jonasschnelli based ip fixes

lontivero:
There is an interesting issue in https://github.com/bitcoin/bitcoin/pull/6446#issuecomment-121883995 because some ISPs detect a netscan in 18333 port.

This PR includes a @jonasschnelli's "quick and dirty solution" (as he called it) https://github.com/jonasschnelli/bitcoin-seeder/commit/4abf94b12baa62b1da60c0129c754f3f20171268 and @sipa is ok with this (requests a PR).

I didn't test it. Just let me know if you are okay with this solution and I will test it then.

### DIFF
--- a/NBitcoin/IpExtensions.cs
+++ b/NBitcoin/IpExtensions.cs
@@ -214,7 +214,7 @@ namespace NBitcoin
 
 		public static bool IsRoutable(this IPAddress address, bool allowLocal)
 		{
-			return address.IsValid() && !(
+			return address.IsValid() && !(  address.IsReserved() ||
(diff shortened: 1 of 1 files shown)


NicolasDorier:
since we ensureIPv6 just before, I think the result of that is always false.

jonasschnelli:
I would recommend this patch for DNS seeder only,... because it's a really ugly hack.

NicolasDorier:
Ok, so I close this one, since we are not making DNS seeder. Normal wallets should not be bothered by it.

PR #122, opened 2015-09-11
Added an overloaded version of Verify method

emrekenci:
The overloaded method consistently returns false or true. The original method returns false in some cases and throws an exception in others. Added an out parameter for explanation of the result.

### DIFF
--- a/NBitcoin/TransactionBuilder.cs
+++ b/NBitcoin/TransactionBuilder.cs
@@ -144,6 +144,20 @@ namespace NBitcoin
 		#endregion
 	}
 
+    /// <summary>
+    /// Exception thrown when not enough funds are present for verifying or building a transaction
+    /// </summary>
+    public class DuplicateInputsException : Exception
+    {
+        public DuplicateInputsException(IndexedTxIn input)
+            : base(String.Format("There are duplicate inputs. Duplicated input: : {0}", input)) { }
+    }
+
+    public class ScriptVerificationFailedException : Exception
+    {
+        public ScriptVerificationFailedException(IndexedTxIn input)
+            : base(String.Format("Could not verify script for this input. Please check your keys. Input: {0}", input)) { }
+    }
 
 	/// <summary>
 	/// Exception thrown when not enough funds are present for verifying or building a transaction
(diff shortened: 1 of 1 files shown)


NicolasDorier:
I am good for it, however the code need to be refactored, there is duplication of the same code in two places with this commit.

The original should call the new one and throw an exception if funds are not enough to not break existing clients.

lontivero:
I know @NicolasDorier doesn't agree with me in this point but returning a boolean is not ok, it should throw an exception. 

Think a moment about this: how should a dev use this code??
Look, the following code looks hilarious, doesn't it?

```
string explanation;
var isValid = Verify(tx, out explanation);
if(!isValid){
   if(explanation.StartsWith("Coin not found!")
   {
       // Ok, and now what!? should I return another error code with another explanation?
   }
   else if(explanation.StartsWith("There are duplicate inputs")
   {
   }
  else ......
  .......
}
```

The *explanation* parameter only has a very little sense for something like this:
```
string explanation;
var isValid = Verify(tx, out explanation);
if(!isValid){
    Log(explanation);
    //ok, again, and now what? return false?
}
```

IMO this is wrong.

NicolasDorier:
Yes, I agree that it should be an enum instead also.

emrekenci:
@lontivero How is that different from handling different exceptions? You either have:

```c#
string explanation;
var isValid = Verify(tx, out explanation);
if(!isValid)
{
    Log(explanation);
    return false; // Or do whatever you need to do when you have a TX that is not valid
}
```
Or you have:

```c#
try
{
    Verify(tx);
}
catch (Exception e)
{
    Log(e.Message);
    return false; // Or do whatever you need to do when you have a TX that is not valid
}
```

You are unlikely to use the explanation to remedy the error. Your code will likely be built for scenarios where the result will be true. i.e. the transaction will be valid. If it is not, you will need to debug the code and all you need is the message.

Manually throwing an exception is bad practice in my opinion as it has performance effects.

But if we have to preserve the old version and if we cannot duplicate code, then we have to throw the exception and return an enum according to the exception type. But actually to do that, you will need to write the hillarious type of code that you wrote. Or switch the exception.

So the second method that will return an enum will have to call the first one and return a result according to the exception the first one returns. But in this approach, you cannot get good information about the error. You can't do this, for instance:

```c#
"Could not verify script for input with transaction hash: " + input.PrevOut.Hash + " index: " + input.Index
```
Because you will be confined to the limits of an enum. So an out parameter would be better IMO.

Anyway. For consistency, I removed the extra method that had the out parameter. And replaced the "return false" lines with appropriate throw exceptions.

This way, the dev. can have an idea of why the TX is not valid.

But we have a method that only returns true or throws an exception. Never returns false :)

Still, I think it's better than the older version.

Ultimately, I think this method should be depracated and replaced with a bool (tx, out explanation) method.

The code can remain duplicated until the clients see the depreciation message and change their code.

Let me know what you guys think.

lontivero:
@emrekenci 

I don't think this change will be accepted because we should never throw the generic *Exception* exception. Doing that forces the developer to use a *catch all* that is not good. Also, as you says, a method returning always true makes no sense.

Now, about **exceptions** and **error codes** (true/false and out messages) there is a big deffernce:

+ An exception *bubbles* and can be handled at the *right level*.
+ An exception can contain *a message, inner exceptions, error codes, extra data, stack trace* and whatever is needed to handle it.
+ An exception *can not be ignored* (if we don't handle it, the application finishes)

On the other hand:

+ An returned error code  has to be *handled by the caller* (most of the time it is the incorrect place because of the lack of enough context). 
+ An returned error code  (and a *reason* out message) is usually not enough. 
+ An returned error code *can by ignored* (a developer can forget to check the returned true/false value) what creates silent, but very important and hard to catch, defects.

I think this discussion will very productive because we need to agree on something as this: exceptions vs error codes.

emrekenci:
For consistency, we can have a custom exception for this method. As other cases return custom exceptions. 

But the cases that return false should throw an exception. Or the exceptions must be replaced with return false. This is again for consistency.

I'm not arguing against exceptions vs error codes. I'm arguing against exception or error message vs enum. I don't think we should have enum here because we can't have enough information in an enum.

You are right that you can have more information in an exception. An in cases of an error, the code should throw an exception.

But for a validation method, I think it's better to have error messages. Because you are actually expecting the method to give you a result that you will evaluate. This is -by definition- not an exceptional situation. You don't need the stack trace to know if you are trying to spend more than you inputs for instance.

To me, the proper use of a validation method looks like this:

An returned error code has to be handled by the caller (most of the time it is the incorrect place because of the lack of enough context).

* So does an exception.

An returned error code (and a reason out message) is usually not enough.

* Not for a validation method. Like I said, you don't need the stack trace for the results of this method.

An returned error code can by ignored (a developer can forget to check the returned true/false value) what creates silent, but very important and hard to catch, defects.

* Again not for the validation method, if the dev. is willingly trying to validate a method, then he must be expecting a result and will not ignore the result. 

These would be true for a method like TransactionBuilder.BuildTransaction for instance.

emrekenci:
If your code looks like this:

```#
try
{
     builder.Verify(tx)
}
catch (Exception e)
{
     switch e
     {
          case NotEnoughFundsException
          {
               // Do something
          }
          case AnotherException
          {
               // Do something else
          }
     }
}
```

This is not a good way to use a validator IMO.

You should not be programming in a way that EXPECTS to receive an Exception. An exception should remain for exceptional situations.

To me, the this is what it should look like:

```#
try
{
     builder.Verify(tx)
}
catch (Exception e)
{
     Log(e)
     // Stop this transaction leave it to be debugged later.
}
```

OR

```#
string explanation;
if (!builder.Verify(tx, out explanation))
{
     Log(explanation)
     // Stop this transaction leave it to be debugged later.
}
```
The second one has very large performance advantages against the first one as an exception will not be thrown and you won't need to get the entire stack trace into memory.

lontivero:
I think you have a point when you say *this is a validation method* and I agree on that. So, provably the best should be something like the following:

```
var verificationResult = builder.Verify(tx, out explanation);
if( verificationResult != Verification.Ok){
}
```

Where *Verification* is an enum (Ok, CoinNotFound, NotFundsEnough, etc...). I think @NicolasDorier could agree with something like this and this is better than the out result string message.

However IMHO, given the caller usually cannot handle the problem, it will wrap it in a "TransactionVerificationException" (or something like that) and trows it to be handled at the right level.

Anyway, the enum is IMO better than what we have now and there are two possible implementations, I think:

```
public bool Verify(Transaction tx, out VerificationError error)
```
And

```
public VerificationResult Verify(Transaction tx)
```

What do you think, @emrekenci

emrekenci:
@lontivero The enum is cleaner. But the out message can give more feedback to the dev. For instance, for duplicate inputs, you can specify which input is duplicated via an out message. With the enum, you can only give generic feedback. So it's a trade off.

In either case, I think the current state of this pull request is better than the main branch.

I had to download, replacte the NBitcoin .dlls and manually debug the method to figure out which "return false" statement was returning in my case.

At least now the dev can consistently have a clear idea about why the verification fails via the exceptions.

If you both agree, I suggest we merge this one and then open another issue for refactoring the method to a better format. We can make it enum or error message. That can be discussed in another issue.

Agreed? @NicolasDorier @lontivero ?

lontivero:
Well, I don't have to agree, @NicolasDorier is our *benevolent dictator*. And believe me that I really understand you, I know how bad it feels to be forced to debug the NBitcoin code just to know why it does what it does.

I think the current PR creates several exceptions when we need just one *NBitcoinException* or *TransactionVerificationException*. Also, it makes no sense to return always *true* and finally, I think is to ok to force developers to parse the exception message to get useful infomation.

Provably we should move to an enum *first* and discuss the exceptions issue later.

emrekenci:
I was just asking to be polite :) Nobody has to agree in the end. I'm already using this version.

It makes more sense to return true or throw an exception in comparison to returning false for some cases and throwing exceptions for others.

NicolasDorier:
no need to close that.

I agree this is a problem to fix, several people told me about TransactionBuilder being a pain in the ass about the feedback. Also, I don't like the double path. (exception and boolean)

But yes, this PR as now is not well refactored and there is code duplication. I think the best is an object which wrap the problem in either an exception or a class which have contextual information helping to debug. I'll keep that in mind to change soon though.

emrekenci:
@NicolasDorier There is no duplicate code anymore. I removed the method with the out parameter and replaced the "return false" exit points with exceptions. I think you should merge this until you write a wrapper.

NicolasDorier:
sadly, it currently break currently break expectation of Verify, and I prefer not breaking user's code.
I plan more to make another Verify and make the old one obsolete, but still having the same code path for both.

Let's keep that open for now.

NicolasDorier:
Addressed with test and without breaking changes in : fe0ba6ececead0a1dd3b2063d4f653882e237048 with the TransactionBuilder.Evaluate method, pushing a new nuget package.

NicolasDorier:
Included in 2.1.4.9

emrekenci:
Thanks Nico.

NicolasDorier:
@emrekenci  I released 2.1.5.0, I'm breaking again stuff.
Now it is not Evaluate method but the Verify. Now verify checks way more stuff than before so it might have broke your code.

PR #123, opened 2015-09-17
Refactoring - Split Base58Encoder and Base58CheckEncoder

lontivero:
This PR splits the Base58Encoder in two classes 

- **Base58Encoder** and,
- **Base58CheckEncoder**

It makes the code easier to read and avoid the need for having the **Check property flag**, the **if(Check) statement** and the **private methods**. The change is quite isolated with no side effects. The **Core Tests** pass green all.


NicolasDorier:
just nit, can you not rename pszBase58 to PszBase58 ?

lontivero:
Done!

Issue #127, opened 2015-09-29
Why do we have a DeterministicECDSA class?


lontivero:
NBitcoin includes BouncyCastle as part of its code base and BC already allows us to get a deterministic K value as part of the [ECDsaSigner](https://github.com/MetacoSA/NBitcoin/blob/4f660fa5abf3a0a4fe72ffa958661df813da07e7/NBitcoin.BouncyCastle/crypto/signers/ECDsaSigner.cs).

Basically, the **ECDsaSigner class** has ctor that accept an **IDsaKCalculator** as parameter which has two implementations, one of those implementations is the **HMacDsaKCalculator class** that prodives _deterministic K values using HMac_. 

It seems we don't need the DeterministicECDSA class because the funtionality is already provided by BC.

@NicolasDorier: Is this correct? or, is there some reason to have our own implementation?

lontivero:
This can be closed because the PR #128 was merged.

PR #128, opened 2015-09-30
Simplify DeterministicECDSA class

lontivero:
This PR is just to explain what I say in the issue #127. 
I replaced the current DeterministicECDSA class implementations with a thin wrapper around the BC ECDsaSigner and it works ok.

If you take a look at the new implementation (simplification), you will see that we hardly need this class because all what it does is already provided by BC.


NicolasDorier:
This is awesome, but why have have you removed some Assert.Equals in the tests in IETFDetailedExample ?

lontivero:
I removed those asserts because this PR is just to show that BC already provides the mechanism for using hmac to have deterministic K values for signing. And given the DeterministicECDSA class is an _internal_ class that is used in only one place (ECKey class) my point is that we don't need this class and we could just delete it.

In other words: my idea is not simplify the class as I show in this PR, my idea is remove the class and that's why I didn't care about commenting those asserts.

Anyway, if you think this class worths, I can improve the code and keep the asserts. Just let me know.

lontivero:
I've gave it a deeper look at the BC's ECDsaSigner class and it doesn't expose the LatsK and LastR values , sadly. I understand that's a problem for testeability reasons. However, given that the ECDsaSigner comes from a well-tested 3rd party library we could assume it works (as we do with the rest of the BC classes).

I also was thinking about delete the DeterministicECDSA class and now I think a better approach would be just use this simplified version (step nr 1) and then we could consider delete it if you agree with that (step nr 2).

NicolasDorier:
We don't need the LastK and LastR but I'd like to keep the test.
```
Assert.Equal(test.S, ECDSASignature.FromDER(result).S);
```
I don't understand why I'm not also testing R, can you add.

```
Assert.Equal(test.R, ECDSASignature.FromDER(result).R);
```

Let's keep DeterministicECDSA with the implementation of BC as you did for now, and on a second pr delete it if we can. But I'd like to keep the test, because it is an important method.

PR #129, opened 2015-10-04
Remove unused references

lontivero:
There are some unsed references that can be removed. Core and Crypto UT Traits pass green (I didn't run the rest, honestly).


NicolasDorier:
Can you check that portable libs always compile ? if I remember I added these references for portable tests.

NicolasDorier:
what do I do with this PR ? I think portable libs do not compile anymore with it.

lontivero:
you're right. It doesn't work for PLs.

PR #131, opened 2015-10-05
Disallow creation of new Networks

lontivero:
Add a private constructor for Network class to prevent someone screw it up.


NicolasDorier:
well, actually I wanted anyone to create its own Network for simple alt coin support, but I guess for now it does not hurt.

PR #136, opened 2015-12-25
RPC related improvements (importprivkey, importaddress, listunspent)

veqtrus:


NicolasDorier:
good idea, however 

```
await SendCommandAsync("importaddress", address.ToString());
```

Will result in deadlock if someone call .Wait.
So can you replace with either

```
return SendCommandAsync("importaddress", address.ToString());
```

Or 

```
await SendCommandAsync("importaddress", address.ToString()).ConfigureAwait(false);
```

veqtrus:
Done

NicolasDorier:
Thanks included in 2.1.5.12 I just pushed on nuget

PR #137, opened 2015-12-28
Increase size of BytesSent & BytesReceived in RPC client to avoid overflow exceptions

graffen:


NicolasDorier:
you already got problem with that in practice ? if so I will push a new nuget package just for that.

graffen:
Yup,  ran into it pretty quickly when retrieving the list of connected clients on my node. The first one on the list had quite large send/recv counters.

NicolasDorier:
ok, I just pushed a new version on Nuget with this PR merged (2.1.5.13). Thanks!

PR #143, opened 2016-03-08
Rename getDataPayload.cs to GetDataPayload.cs to fix build in Linux

knocte:
Linux filesystem is case sensitive :)


NicolasDorier:
what you can compile on linux ???

knocte:
> what you can compile on linux ???

Of course, don't you know the mono project? :)

NicolasDorier:
I know xamarin, but did not know that they ported msbuild on it.

knocte:
Well Xamarin is a company (well, *was* a company! now it has been acquired by Microsoft!).

That company was in charge of maintaining 4 products:
- The mono project (open source, for Linux/MacOSX/Windows).
- Xamarin.iOS (proprietary, for iOS devices).
- Xamarin.Mac (proprietary, for MacOSX).
- Xamarin.Android (propietary, for Android devices).

The three latter were all based on Mono, which works beautifully in Linux since many years. The equivalent of MSBuild in the mono world is called "xbuild", and was created many years ago (although now it's going to be replaced by MSBuild itself because it has been recently opensourced by Microsoft, and the Xamarin people are fixing it to be cross-platform).

That being said, today I tried to build NBitcoin using MonoDevelop, not xbuild (but if it builds in MonoDevelop, it's very likely that it also does in xbuild).

NicolasDorier:
and it just worked minus the nits ?

knocte:
Well, I had to comment a couple of things that are not supported in the version of Mono I'm using (3.10.0), but I've checked the sources of mono master (https://github.com/mono/mono) and both things are implemented in the latest version (4.3) so it's only a matter of upgrading.

NicolasDorier:
I think you are using the wrong project, for mono use

> "NBitcoin.Portable\NBitcoin(Mono).csproj"

It should compile right away.

NicolasDorier:
Also there is the NBitcoin.Mono package that should work with nuget on Xamarin studio.

knocte:
I know that you're offering two different packages for MS.NET and Mono in nuget :) But I wanted to check for myself why.

And so, IMNSHO, NBitcoin users should use, moving forward, just one Nuget package, regardless if they're using Mono or not, if they're using Mono 4.3 or higher.

Unfortunately, Mono 4.3 is still unstable (stable version will be 4.4, which is not released yet; and even when it's released, it will take a while until distributions end up packaging it), so let's live with NBitcoin.Mono for now :)

PR #145, opened 2016-03-09
Some nitpicks related to MonoDevelop

knocte:
See each commit for explanation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.suo
 *.user
 *.sln.docstates
+*.userprefs
 
 # Build results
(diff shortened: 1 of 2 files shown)


NicolasDorier:
awesome thanks for checking that. Also, I hope to adapt NBitcoin to .NET core, but still not had the occasion to play with it, do you know if it is worth it ?

knocte:
I think .NET core is cool. It's the future of .NET. But it's maybe a bit early to adopt it.

I myself will not look at it, at least until F# is compatible with it :-)

PR #146, opened 2016-04-07
xunit upgraded to v2.1.0

andbene72:
-PropertyData -> MemberData
-Assert.DoesNotThrow not needed because an exception is a failed test [https://github.com/xunit/xunit/issues/188](https://github.com/xunit/xunit/issues/188)
-Unused references removed (ProtocolBuffer...)


NicolasDorier:
thanks can you remove the binaries from package completely instead of updating them ? I'm surprised I commited them.

NicolasDorier:
ah sorry that's already the case good

NicolasDorier:
For information I was not sure about updating to 2.1.0, because I have problems with my continuous integration server. Very often the tests don't want to run for mysterious reason, it is a bug that happens only above 2.0, and nothing has been done since several years about it.

NicolasDorier:
image

NicolasDorier:
For info, this is the bug I have on several build time to time above 2.0:

image

andbene72:
I think can be this https://github.com/xunit/xunit/issues/261
solution is to upgrade nuget on the integration server.

Anyway I see that I have also to update the portable solution so if you prefer you can revert this commit, but I think it's not a long term solution, especially if it's a matter of "package supports frameworks that your version of NuGet doesn't know about. " as stated in the xunit issue.

NicolasDorier:
I updated nuget now:

image

NicolasDorier:
I will reverse this commit for now.

andbene72:
OK.
Googling I found that 
> xunit.dll is for v1, and xunit.execution.desktop.dll is for v2.
I can speculate that somewhere there is a v1 reference.

NicolasDorier:
I reversed it as well as pushed some of my outgoing commit, be carefull you'll probably need to rebase.

NicolasDorier:
Here is the line for running the console tests:

> C:\Xunit\xunit.console.clr4.x86.exe NBitcoin.Tests/bin/Release/NBitcoin.Tests.exe /trait Core=Core /trait UnitTest=UnitTest /trait RPCClient=RPCClient /trait TestNet=TestNet /trait MainNet=MainNet /trait RestClient=RestClient

If you can make a PR which make it works will be happy to merge

NicolasDorier:
oh wait, maybe the problem comes from me using the xunit v1 console app

NicolasDorier:
you deleted your repo, I can't try your commit anymore, I think the problem came I was using wrong version of the console xunit runner, can you get the commit again ?

andbene72:
Just a moment that I'm fixing portable test :)

NicolasDorier:
I'm going to bed (2am here) will take a look tomorrow, don't forget to rebase as I pushed some commit.

andbene72:
Ok, good night.

andbene72:
Follow here https://github.com/MetacoSA/NBitcoin/pull/147

PR #147, opened 2016-04-07
xunit upgraded to v2.1.0 in both solutions

andbene72:
-PropertyData -> MemberData
-Assert.DoesNotThrow not needed because an exception is a failed test [xunit/xunit#188](https://github.com/xunit/xunit/issues/188)
-Unused references removed (protobuf-net...)


NicolasDorier:
I ran the tests locally, it worked. My CI server crashed and is offline for now though.

NicolasDorier:
sorry, I will reverse it again...

image

Also, the tests does not appear in my visual studio anymore, I googled a lot and followed similar problems but could not fix it. On top of that I still have problem of "xunit.runner.desktop" not found on other projects using 2.1.0.

I will likely just downgrade to 1.9.2 all my projects.

PR #153, opened 2016-04-18
Updated Bouncy Castle to 1.8.1

andbene72:
Platform specific calls mostly wrapped in utils/platform class
Added PORTABLE compiling directive


NicolasDorier:
wow, thanks, I would like to review it but with 543 files it is kind of hard.

Can you describe me how I can verify your updates match the version of BouncyCastle with some set of magic git commands ?

andbene72:
Hi Nicolas,

I have tried and tried again using git following this idea
http://stackoverflow.com/questions/1968512/getting-the-difference-between-two-repositories

but the problem here is that there is a different directory structure and git diff not support a way to specify directories to compare. Also I haven't found a GUI way to use it.
Exist anyway a quite fast non git workaround using winmerge.
Consider that the real difference between bc and this pull nbitcoin is in MacUtilities.cs where I have adjusted for hungarian bug, the other of 542 files difference is "namespace" and "using" or reference in comment to namespace.

1. Download the current latest commit
https://github.com/bcgit/bc-csharp/archive/22467cd66eea50f418f6398910def1f3b84db404.zip
(the zip of the commit I used, just to avoid newer commits)
Unzip it.

2. Open winmerge, open on the left to bc-csharp\crypto\src, the right to NBitcoin\NBitcoin.BouncyCastle. 

3 On tools menu, customize column, check "Differences". On view menu, check "show difference" and "right unique". Uncheck "identical" and "left unique".On edit -> options general tab, check "auto scroll to 1st diff", "close win with esc", compare tab "ignore blank lines", "ignore CR diff"

4. Now you can order the differences column decreasing. This is the result
image

Andrea

NicolasDorier:
Normally I already fixed the hungarian bug in MacUtilities. (it is a NBitcoin user which found it originally)
Thanks, I'll try your way to compare.
I can replace all namespaces so it does not show me too much change.

NicolasDorier:
Can you remove all files that are not needed for NBitcoin that you added ? like all the curves.
Also namespace should not be changed.
I'd like to minimize the diff so review is easier both for me and other.

andbene72:
Well, the curves was added by the new BC version. I have now removed these and updated the CustomNamedCurves.cs that reference them. 
But if you minimize now the difference between the source of BC embedded in NBitcoin and this pull coming from 1.8.1, you get the tradeoff of having in the future a BC in NBitcoin that have more diff with a future version of Bouncycastle. 
Regarding namespace, the added one is using NBitcoin.BouncyCastle.Utilities , it mean that in that class now there is a call to a Platform specific function in the Platform class of that namespace.

As having updated cryptofunction is better for security and for performance my advice is too keep that files the same as original Bouncycastle (even the initial changing of Org to NBitcoin was probably a not good choice). In my project I have changed the approach back to using original BC library due to platform Hell becoming quick unmanageable.

NicolasDorier:
@andbene72 I am surprised you came back to Org for fixing plateform Hell. What do you mean ? I never had problems with other plateform, I isolated BouncyCastle in NBitcoin because the BouncyCastle package had a bug in Hungarian and was way slower for checksig and signature creation than their master. So I just used their code instead.

My ultimate goal is to get rid of all the code except what NBitcoin use on other plateforms. (NBitcoin does not depends on BC for windows)

I'll need to review all those file one by one so it will take me time. (most changes seems obvious though)
I will try to remove files with lots of diff but not used by NBitcoin.

My goal is to minimize the review surface so I can compare between BC and the submitted code. I'll give it a try again with the removed curves.

NicolasDorier:
Why is their Normalize calls in pubkey and key ?

andbene72:
> Why is their Normalize calls in pubkey and key ?

It's the same call as before, see ECPoint.cs line 100, I have only fixed the warning instead of suppressing it.

        /**
         * Normalizes this point, and then returns the affine x-coordinate.
         * 
         * Note: normalization can be expensive, this method is deprecated in favour
         * of caller-controlled normalization.
         */
        [Obsolete("Use AffineXCoord, or Normalize() and XCoord, instead")]
        public virtual ECFieldElement X
        {
            get { return Normalize().XCoord; }
        }

NicolasDorier:
for information I am working on it. I'm attempting to strip everything NBitcoin does not use and include BouncyCastle directly in the NBitcoin assembly.

andbene72:
> I am surprised you came back to Org for fixing plateform Hell.

I'm trying to add a NBItcoin.UWP project and still trying to figure out the best way.
I have currently reached the point having a Bouncycastle project that compile under UWC.
Now I like to move to NBitcoin part. There is a short explanation of all conditional compiling statement used ?

Fortunately this is part of a personal project, i can wait your clean up.

NicolasDorier:
I'll finish tonight or tomorrow the cleanup.

Normally NBitcoin.Portable have 2 portable profile which should work with nuget or even compile directly with UWP. The conditionals are in the csproj of those portable profiles.

NicolasDorier:
merged it 9b9eb91148be8c4310fa9b9b648f75d228380583 and 5960ad46b91e5ae4cdc2443a2f46f0a445d0ecb3

PR #156, opened 2016-04-28
Add method to retrieve chain code bytes for an ExtKey object.

Avalaxy:
I need to get (and use) the chain code for generated HD nodes. I tried retrieving the chain code by calling GetBytes() so I can use the right 256 bits for the chain code, but somehow GetBytes returns 592 bits instead of the expect 512 bits (left 256 bits master private key, right 256 bits master chain code). Since the chain code bytes are right there in the class as a private byte array I think it would be very useful to expose them through a method.

Let me know what you think about this approach (I need the chain code no matter what though).

### DIFF
--- a/NBitcoin/BIP32/ExtKey.cs
+++ b/NBitcoin/BIP32/ExtKey.cs
@@ -38,6 +38,16 @@ namespace NBitcoin
 				return nChild;
 			}
 		}
+		public byte[] ChainCode
+		{
+			get
+			{
+				byte[] chainCodeCopy = new byte[vchChainCode.Length];
+				Buffer.BlockCopy(vchChainCode, 0, chainCodeCopy, 0, vchChainCode.Length);
+
+				return chainCodeCopy;
+			}
+		}
 
 		public ExtKey(BitcoinExtPubKey extPubKey, BitcoinSecret key)
 			: this(extPubKey.ExtPubKey, key.PrivateKey)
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Ok, can you also add it to ExtPubKey ? I am curious, why do you need to retrieve the chain code ?

NicolasDorier:
Ha use a property also, not a method.

lontivero:
Other fields like nChild and nDepth are available as read-only properties and I think that vchChainCode  should be available as a property too, not using a GetX method. I also think we shouldn't allow modifications of this array elements and so, you could return a copy of it.

Avalaxy:
See my new commits, I changed the method to a read-only property and also added it for the ExtPubKey.

For me personally I need the chain code because the API I'm using requires this (they want the private key and the chain code. Not sure why it's implemented this way, but current clients already work like this and I have to work with it).

NicolasDorier:
thanks, can you fix identation and squash your commits ? (git rebase -i HEAD~5)

The chainCode is not a copy, but well, not a big deal.

Avalaxy:
Changed the spaces to tabs, now returning a copy of the byte array and I squashed the commits.

Avalaxy:
@NicolasDorier thanks, would it be possible to publish a minor release on NuGet?

NicolasDorier:
just did 3.0.0.28

PR #160, opened 2016-05-09
Removed unused references

andbene72:
I have removed some unused assembly reference from the main project.
I see that also Microsoft.CSharp is unused, but I have not removed it, as removing it will help if someone wish to use a previous framework version but on the other side keeping it can be useful if in future someone need dynamic keyword. Let me know your preference.

Very good job your cleanup of BC. Just to understand there is a reason for the unreferenced interfaces (like for example Asn1SetParser) you keep there ?

### DIFF
--- a/NBitcoin/NBitcoin.csproj
+++ b/NBitcoin/NBitcoin.csproj
@@ -42,14 +42,8 @@
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Net.Http" />
-    <Reference Include="System.Net.Http.WebRequest" />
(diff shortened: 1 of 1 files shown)


NicolasDorier:
I think Microsoft.CSharp should go if not using it.

For unused reference, there is no reason. I tried to remove as much as I could given limited time, if you see code still unused in BouncyCastle, it must also go. Doing so bring down the size of the NBitcoin package. (the cleanup decreased size of the nuget package from 10MB to 5MB)

NicolasDorier:
my goal is to have such a BC library that I can get rid of it and just include it into NBitcoin dll.

PR #161, opened 2016-05-10
More cleanup

andbene72:
- Microsoft.CSharp removed
- More cleanuo of BC
- (pbkdf2) Resolved an ambiguity (when BC will be embedded in NBitcoin)


andbene72:
If you plan to remove USEBC conditional compiling directive a few more classes could be removed.
Not really necessary but if in a future you plan to create a NBitcoin that compile directly under UWP you should consider that things like this 

#if !USEBC
using System.Security.Cryptography;
#endif

will grow in things like this

#if !USEBC
#if !WINDOWS_UWP
using System.Security.Cryptography; 
#else
using Windows.Security.Cryptography; 
#endif
#endif

andbene72:
Well, better not remove USEBC, I have verified that for example RIPEMD160 is not present in that framework and BC version can be reused.

NicolasDorier:
awesome

PR #164, opened 2016-05-22
ExtKey constructor to create object from a private key and chain code

Avalaxy:
As discussed on Twitter (https://twitter.com/NicolasDorier/status/733997857120739328) I have a private key and a chain code and I need to be able to create an ExtKey object from that. I added a constructor that does this.


Avalaxy:
Woops, hold on. Forgot to change my spaces to tabs again. Going to squash two commits together here.

PR #166, opened 2016-05-31
A name fix and some really small adds toward UWP native compilability

andbene72:
One question, I see that you use TraceSource class that for what I have understand is going to be phased out. Are you interested if I migrate that part to the new EventSource class ?


knocte:
There seem to be many unrelated-whitespace changes in the 2nd commit.

andbene72:
Just installed VS on a new pc, seem that by default change tab in space, going to fix the commit. Thank you.

NBitcoin/Utils.cs
```diff
@@ -699,7 +698,21 @@ public static IPEndPoint ParseIpEndpoint(string endpoint, int defaultPort)
 			}
 			catch(FormatException)
 			{
+#if !WINDOWS_UWP
 				address = Dns.GetHostEntry(ip).AddressList[0];
+#else
+				foreach(HostName localHostName in NetworkInformation.GetHostNames())
+				{
```
NicolasDorier: I don't understand that, it seems to change completely the logic of ParseIpEndpoint by returning localhost everytimes an IP is not well formed.

NicolasDorier:
Can you squash your indent fixes ?
Also for UWP, I think I need to create a separate CSProj with the correct pre processor constants. One of which being USEBC, so the #if USEBC | UWP is not necessary.

Also see my comment about the parse ip

NicolasDorier:
EventSource seems to be coupled to ETW which is windows specific.
TraceSource is a native .NET logging system where you can route logs to "listeners" by modifying the config file.

NicolasDorier:
Also, now that I have a windows 10 environment I'll create very soon a package specific for UWP distributed on nugget.

andbene72:
I'm not going to create a UWP package alone, I have tried only to understand the implications of trying to do this, I'm also new to UWP and as blockchain professional the best way to learn his to do it on a blockchain project.
What I have done outside my git fork is to do something like in the following picture, having separate project for each target and a common shared one for common code.
image
Now all projects compiles except for UWP as his framework (and logics) are quite different.
WINDOWS_UWP is a directive that any UWP project intrinsicaly declare for the common project.
The logic of a shared project is to keep together the shared sources moving different implementation of platform specific classes in their project.
I see that many project follow a similar pattern so I give the use of WINDOWS_UWP as acquired even because I doubt that there is a perfect matching of use between USEBC and WINDOWS_UWP.

I understand about Tracesource that the reason of my question. Eventsource is tied but available in any framework. I think that it's possible to create an implementation of tracesource in the specific UWP project so keeping the old way.

About ParseIpEndpoint, thats was a my error in copy and paste from an external project where I have tested the dns related functionality of UWP. Really sorry, this should never happen. I will in any case commit the right code, it can be useful for you when you work on UWP specific package.

Sorry for my bad english bad what do you mean with "Can you squash your indent fixes ?"

Let me know if i can be useful to this project in any way.

Andrea

NicolasDorier:
NBitcoin already has an android project.
Squash mean fixing your PR so there is no useless/fixup commit: https://github.com/bitcoin/bitcoin/blob/90963e53707cf244e6a2a804341e9280628b0c34/CONTRIBUTING.md#squashing-commits

NBitcoin/Crypto/Cryptsharp/Pbkdf2.cs
```diff
@@ -94,19 +91,19 @@ public Pbkdf2(KeyedHashAlgorithm hmacAlgorithm, byte[] salt, int iterations)
...
-			_iterations = iterations;
-			_hmacAlgorithm = hmacAlgorithm;
-			_digest = new byte[hmacLength];
-			_digestT1 = new byte[hmacLength];
+			NBitcoin.Crypto.Internal.Check.Null("hmacAlgorithm", hmacAlgorithm);
+			NBitcoin.Crypto.Internal.Check.Null("salt", salt);
+			NBitcoin.Crypto.Internal.Check.Length("salt", salt, 0, int.MaxValue - 4);
+			NBitcoin.Crypto.Internal.Check.Range("iterations", iterations, 1, int.MaxValue);
+			if(hmacAlgorithm.HashSize == 0 || hmacAlgorithm.HashSize % 8 != 0)
+			{ throw Exceptions.Argument("hmacAlgorithm", "Unsupported hash size."); }
```
NicolasDorier: can you add new lines here ?

NicolasDorier:
cool, check my only nit. Once squashed I'll merge. UWP has the socket class by the way ?
I remember I never ported to WinRT runtime before because they removed the Socket class.

andbene72:
Squashed :)

Socket was [added back](https://blogs.msdn.microsoft.com/dotnet/2015/07/28/net-networking-apis-for-uwp-apps/) in net core 5 but is quite incomplete/different..
For example Disconnect is achieved doing a .Dispose of the instance. And I'm quite sure that is not the only syntactic/behavioral change. 
I see that there is a package on Nuget called System.Net.Sockets 4.0.0 that should provide some class as NetworkStream.

Other things that I have noticed trying to do a UWP project from the current source code are:
- TraceSource not available 
-  [Serializable] not available
- ICloneable not available
- System.Security.Cryptography now Windows.Security.Cryptography but incomplete (BC cover missing part but probably better to use BC at all instead, in the my test I have compacted that separate project in the main project as your plan)
- ToLower behave like PORTABLE
- Tasks instead of threads
- Some Stream methods unavailable 
104 errors many can be fixed in a simple way, but some need to think in the new way and code accordingly.

NicolasDorier:
For TraceSource, you should do like I did I think for android: Basically I created TraceSource without implementation, so the code still build without too much define. I'll review the PR again later. Thanks for the help :)

andbene72:
I see that there is source code of TraceSource [here](https://github.com/dotnet/corefx/tree/eede273a4dfabcea608621f5e1bbf8ad00584cfb/src/System.Diagnostics.TraceSource). I will explore it to understand if we can have a version to embed usable for UWP and Android.

NicolasDorier:
@andbene72 is it safe to assume that microsoft wants to drop portable profile in favor of shared project ? When you add a nuget reference in a side project, which assembly is resolved for development auto completion ?

andbene72:
Question 1 - Is it safe to assume that microsoft wants to drop portable profile in favor of shared project ? 

Well I don't think that shared project is in place of portable project.
In my example there isn't but only because it's a my choice doing some tests (as i do for mono).
image
The advantage is that projects are clearly separated, use different directories so no problem with packages.config or project.json sharing the same directory and so project
Also you can create a different shared project for source code shared only in a subset of target platforms as System.Numerics for example.

NBitcoin/Utils.cs
```diff
@@ -699,10 +698,39 @@ public static IPEndPoint ParseIpEndpoint(string endpoint, int defaultPort)
...
+					address = IPAddress.Parse(adr);
+#endif
 			}
 			return new IPEndPoint(address, port);
 		}
+
+#if WINDOWS_UWP
+		private static async Task<string> DnsLookup(string remoteHostName)
+		{
+			IReadOnlyList<EndpointPair> data = await DatagramSocket.GetEndpointPairsAsync(new HostName(remoteHostName), "0");
```
NicolasDorier: here you have to add .ConfigureAwait(false); or you'll endup with deadlock in context like ASP.NET or WPF, and likely UWP if there is a user interface.

NicolasDorier:
Can you fix the small nit to prevent a deadlock ? I'll ten merge it. Thanks a lot.

andbene72:
Question 2 - When you add a nuget reference in a side project, which assembly is resolved for development auto completion ?

You mean which is resolved in the shared project ?
Well the same library in theory should be coherent between different platforms but in any case 
image

image

For nuget assembly is the same, the one in the selected project.

NicolasDorier:
wow this is powerful awesome. I definitively need to move to shared projects for making more easy to compile NBitcoin anywhere. The way I'm doing it now is through some msbuild black magic incantation that I don't even remember.

NBitcoin/Utils.cs
andbene72: Windows.Foundation.IAsyncOperation<IReadOnlyList<EndpointPair>> seem to have not .ConfigureAwait method. My sample app work without it. Do you think that I have to put the code in this way ?

private static async Task<string> DnsLookup(string remoteHostName)
{
SemaphoreSlim mutex = new SemaphoreSlim(1);
IReadOnlyList<EndpointPair> data;

  await mutex.WaitAsync().ConfigureAwait(false);
  try
  {
    data = await DatagramSocket.GetEndpointPairsAsync(new HostName(remoteHostName), "0");
  }
  finally
  {
    mutex.Release();
  }

[....]
}`

NicolasDorier: I doubt you don't have it or NBitcoin would explode everywhere during compilation on UWP.

```
GetEndpointPairsAsync(new HostName(remoteHostName), "0").ConfigureAwait(false);
```

andbene72: Solved 
IReadOnlyList<EndpointPair> data = await DatagramSocket.GetEndpointPairsAsync(new HostName(remoteHostName), "0").AsTask().ConfigureAwait(false);

NicolasDorier:
interesting that it require you to call .AsTask()

Issue #183, opened 2016-08-19
ReadEx bug and other improvements


GSPP:
There's a bug in ReadEx (which I'm willing to fix). Also, there is some confusion there I think. So before addressing that I wanted to check with the project owners.

The bug is that Stream.Read never returns -1. So the [special case](https://github.com/MetacoSA/NBitcoin/blob/b591359ea3ff7ed3b9d439d1dadb7ddfd8eb8abe/NBitcoin/Utils.cs#L151) never hits. It should trigger based on zero which signals the end of the stream.

This signal is universal. It is not necessary to special case certain stream types. Therefore, all type checks in [this line](https://github.com/MetacoSA/NBitcoin/blob/b591359ea3ff7ed3b9d439d1dadb7ddfd8eb8abe/NBitcoin/Utils.cs#L153) should be removed.

The [cancellation handling](https://github.com/MetacoSA/NBitcoin/blob/b591359ea3ff7ed3b9d439d1dadb7ddfd8eb8abe/NBitcoin/Utils.cs#L137) is dubious: In case of cancellation the IO will continue to run. It will continue to consume bytes from the stream and write into the buffer. Therefore, the only safe thing to do in case of cancellation is to discard the stream (the caller must do that). IOs in .NET are not cancellable.

Is it possible to use the `public virtual Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)` overload? That would be the best fix. In newer .NET Framework versions all common streams should have an efficient implementation for this.

Alternatively, I'd suggest creating a Task for the read using `FromAsync`. Tasks are composable more easily and faster. Implementing cancellation with them should be faster than the double-WaitHandle approach.

[This](https://github.com/MetacoSA/NBitcoin/blob/b591359ea3ff7ed3b9d439d1dadb7ddfd8eb8abe/NBitcoin/Utils.cs#L139) is a bug because it fails to call `EndRead`. This violates the contract of `Stream`. The bug probably is of no practical consequence since the stream must be discarded after cancellation anyway.

What are your thoughts?

NicolasDorier:
Frankly, I hate this piece of code, it grew organically by fixing strange bugs and performance problems I hit in the past.

Stream.Read returns -1 in the case of NetworkStream for example. (if I remember, it is 
The special case of Network stream type was because of a big performance issue when you read a lot. (You can try that yourself by downloading all the blocks from one of your nodes and timing the speed)

[...]

PR #184, opened 2016-08-19
Improve ReadEx

GSPP:
This PR does the following:

1. Only take the slow path if token is cancelable.
2. Always call EndRead.
3. Fix stream depletion condition (0 instead of -1).
4. Fix tabs.
5. Generals cleanups. Hope they are OK.
6. Fix callers expecting -1.
8. Fix callers reading 0 bytes. They should not do that. This might have resulted in a performance improvement.
9. Argument validation to harden this brittle code path.

I ran the tests for both ReadEx variants on Desktop 4.6. Is this enough? I don't have anything else set up.

There are test failures but they seem configuration related. Some did not terminate which was annoying. Maybe a test subset could be defined which is easy and safe to run on contributors machines.

I did not test performance since none of this should have caused any regressions.

In the `ReadMagic` method I found `if(read != 1)` which is always false. Maybe this should be reviewed.

Questions:

1. Is it safe to run tests on a computer with a production wallet?
2. Would you consider switching code style over to the default VS style? I can see that the "if(" styling is non-standard. It's easiest to contribute if you can just press Ctrl-K-D. I do that routinely. It removes the need to manually format many trivial things. I found myself constantly fighting the built-in formatting and it can be quite persistent :)

Addresses https://github.com/MetacoSA/NBitcoin/issues/183.

### DIFF
--- a/NBitcoin/Utils.cs
+++ b/NBitcoin/Utils.cs
@@ -125,62 +125,72 @@ namespace NBitcoin
 #if !(PORTABLE || NETCORE)
 		public static int ReadEx(this Stream stream, byte[] buffer, int offset, int count, CancellationToken cancellation = default(CancellationToken))
 		{
-			int readen = 0;
-			while(readen < count)
+			if(stream == null) throw new ArgumentNullException("stream");
+			if(buffer == null) throw new ArgumentNullException("buffer");
+			if(offset < 0 || offset > buffer.Length) throw new ArgumentOutOfRangeException("offset");
+			if(count <= 0 || count > buffer.Length) throw new ArgumentOutOfRangeException("count"); //Disallow 0 as a debugging aid.
+			if(offset > buffer.Length - count) throw new ArgumentOutOfRangeException("count");
+
+			int totalReadCount = 0;
+
+			while(totalReadCount < count)
 			{
-				int thisRead = 0;
-				if(stream is NetworkStream) //Big performance problem with begin read for other stream than NetworkStream
+				cancellation.ThrowIfCancellationRequested();
+
+				int currentReadCount;
+
+				//Big performance problem with BeginRead for other stream types than NetworkStream.
+				//Only take the slow path if cancellation is possible.
+				if(stream is NetworkStream && cancellation.CanBeCanceled)
 				{
-					var ar = stream.BeginRead(buffer, offset + readen, count - readen, null, null);
+					var ar = stream.BeginRead(buffer, offset + totalReadCount, count - totalReadCount, null, null);
 					if(!ar.CompletedSynchronously)
 					{
 						WaitHandle.WaitAny(new WaitHandle[] { ar.AsyncWaitHandle, cancellation.WaitHandle }, -1);
 					}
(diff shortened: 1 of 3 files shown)


knocte:
> From c5cabd615715330ff0c894d474927f59d893944f Mon Sep 17 00:00:00 2001
> From: "user.name" <user.email>

You seem to have forgotten to set up git user and email.

NicolasDorier:
"Some did not terminate which was annoying" it should not happen, there are some failing tests normally (2 or 3 of them)
But non terminating one might be result of this commit.

NicolasDorier:
> Would you consider switching code style over to the default VS style? I can see that the "if(" styling is non-standard. It's easiest to contribute if you can just press Ctrl-K-D. I do that routinely. It removes the need to manually format many trivial things. I found myself constantly fighting the built-in formatting and it can be quite persistent :)

I think you can use per project settings in vs now, but no idea how.

NBitcoin/Utils.cs
```diff
@@ -125,62 +125,72 @@ public static IEnumerable<List<T>> Partition<T>(this IEnumerable<T> source, Func
...
 				{
-					var ar = stream.BeginRead(buffer, offset + readen, count - readen, null, null);
+					var ar = stream.BeginRead(buffer, offset + totalReadCount, count - totalReadCount, null, null);
 					if(!ar.CompletedSynchronously)
 					{
 						WaitHandle.WaitAny(new WaitHandle[] { ar.AsyncWaitHandle, cancellation.WaitHandle }, -1);
 					}
-					cancellation.ThrowIfCancellationRequested();
-					thisRead = stream.EndRead(ar);
-					if(thisRead == 0 && (stream is Message.CustomNetworkStream) && ((Message.CustomNetworkStream)stream).Connected)
```
NicolasDorier: probably why some tests are hanging.

NicolasDorier:
I'm not very confident about this PR. In the case of NetworkStream, it might returns 0 because it is waiting data to be sent by the Socket, not because there is nothing to read anymore.
This is tricky... I think I should just make sure that no NetworkStream is passed to ReadEx somehow, but it requires some refactoring. :(

GSPP:
`NetworkStream` only returns 0 on stream depletion. *All* streams are like that. If not, they are broken and BCL streams are not broken.

Socket waits ... wait. They don't return until done.

Regarding the EndRead thing, yes this is a bug in this PR. I will fix that. I will fix it by not fixing the underlying bug which is not calling EndRead...

GSPP:
Regarding user and email: I'm not very familiar with Git. In any case I will not post my email address to the commit history. I do not require any attribution either. Any advice?

I could set the author to GSPP. That seems best.

Am I supposed to amend the last commit or create a new one? There is no point in keeping the broken commit around I think.

I'm otherwise ready to post a fix.

NicolasDorier:
just amend your commit and force push to update the pr.

NicolasDorier:
I tested against .NETCore and .NET, it worked fine, so I merged 888c9cad2c12ead788adef3742eb3799259a0042.

PR #194, opened 2016-10-05
Checking the block size (instead of length)

dangershony:
This is probably more correct

### DIFF
--- a/NBitcoin/ValidationState.cs
+++ b/NBitcoin/ValidationState.cs
@@ -191,7 +191,7 @@ namespace NBitcoin
 
 			var root = block.GetMerkleRoot();
 
-			if(block.Transactions.Count == 0 || block.Transactions.Count > MAX_BLOCK_SIZE || block.Length > MAX_BLOCK_SIZE)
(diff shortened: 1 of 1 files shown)


NicolasDorier:
hey I will probably soon remove this code altogether as the consensus validation except the script evaluation part is quite out of date and not very tested.

I am implementing a full node here: https://github.com/stratisproject/StratisBitcoinFullNode/ which validate the consensus fully (segwit as well), it is a work in progress.

NicolasDorier:
oh I'm stupid you are from Stratis ahah :D

NicolasDorier:
removed the code on 3717631b, such code better belong to the full node project.

dangershony:
It's outdated I see ok.
Yeah I am aware of the full node and you are currently working on the consensus but it will reference NBitcoin so I thought this was still valid.
In any case if its obsolete then ignore this pull

NicolasDorier:
the only parts of the code concerning the consensus I'm happy on NBitcoin is the script evaluator and POW check. I'm using still NBitcoin for those two, the other rules, I redevelopped them from scratch for stratis full node.

PR #196, opened 2016-10-22
Magic byte - first byte duplicated bug fix (and test)

dangershony:


dangershony:
This was discovered when reading the stratis blockchain blk0001.dat

NicolasDorier:
thanks a lot, can you fix indents to tab ?

NicolasDorier:
mmh, I don't think we should do that: ReadMagic is expected to read only the Magic and nothing else.

I think this should be fixed at the caller level. (BlockStore)

NicolasDorier:
Oh you are right, nice finding, if you can just fix the indent I'll merge that now, thanks a lot!

dangershony:
Done. 
VS acting funny with identity my code.

dangershony:
By the way this probably explains why when parsing only block headers I had no missing blocks. 
The BlockStore class uses the block size field to skip the content, it probably jumped directly to the next magic byte.

knocte:
Thanks for indenting the code.

Do you mind squashing your 3 commits into one? You can do that by typing `EDITOR=nano git rebase -i HEAD~3` then changing from "pick" to "s" the first word of the 2 lines from the last to commits, then merge all the commit messages into one.

Thanks

NicolasDorier:
Alternatively you can change "pick" to "f" which remove the commit message of the squashed commits.

dangershony:
Done (I think) this looks correct?

knocte:
I see 5 commits now: https://github.com/MetacoSA/NBitcoin/pull/196/commits

What happened?

dangershony:
Ha not sure, I squashed the last two commits (the indent commits) and then was asked to merge (not sure why as no commits where done since my branch was taken) any ideas?

knocte:
@NicolasDorier I think github has now an option that lets you merge a PR by squashing all commits into one from the web UI, can you check?

NicolasDorier:
oh yeah brilliant!

dangershony:
Thanks @NicolasDorier and @knocte 
Any idea why my squash got messed up?

NicolasDorier:
no idea, I've occasionally messed up squash with git (but it is rather rare normally).

I don't know how it can mess up as I don't see how it could create a conflict which need manual merge.

PR #199, opened 2016-10-28
Minor cosmetics (inconsistent naming and method order)

nopara73:
Just minor cosmetics refactoring. I felt like it's a good way to get familiar with these classes.

-Fix inconsistent method parameter naming inside the BuilderExtension child classes (sometimes scriptPubkey is used instead of scriptPubKey)

-Reorder methods of BuilderExtension child classes to be present in the same order as in  the BuilderExtension base class

### DIFF
--- a/NBitcoin/BuilderExtensions/P2MultiSigBuilderExtension.cs
+++ b/NBitcoin/BuilderExtensions/P2MultiSigBuilderExtension.cs
@@ -6,98 +6,98 @@ using System.Threading.Tasks;
 
 namespace NBitcoin.BuilderExtensions
 {
-    public class P2MultiSigBuilderExtension : BuilderExtension
-    {
-        public override bool CanCombineScriptSig(Script scriptPubKey, Script a, Script b)
-        {
-            return PayToMultiSigTemplate.Instance.ExtractScriptPubKeyParameters(scriptPubKey) != null;
-        }
+	public class P2MultiSigBuilderExtension : BuilderExtension
+	{
+		public override bool CanCombineScriptSig(Script scriptPubKey, Script a, Script b)
+		{
+			return PayToMultiSigTemplate.Instance.ExtractScriptPubKeyParameters(scriptPubKey) != null;
+		}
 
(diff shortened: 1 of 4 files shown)


knocte:
> It is not because of the tab/space difference.

Well, the original classes had tabs on them, you're using spaces.

> It is because I reordered child classes functions to be in the same order as in the baseclass.

That's cool, but it should go in a different commit than the fix for inconsistent parameter method naming.

If you fix those 2 things, it becomes readable :)

nopara73:
Yes, that make sense. I could've done that. To be honest I am not very keen to spend 20 more minutes on some minor beautifying for the shake of administration.  

>Well, the original classes had tabs on them, you're using spaces.  
I see it now. It is interesting, I'm just using Visual Studio Community 2015 Update 3 with its default settings on this. My settings: http://imgur.com/a/BRfjv
Wonder if Nicolas has something different and if so why. Or it was just created with an other IDE.

NicolasDorier:
I'd like you fix using TAB to get consistent. Space is for linux guys, we are worth better. :p

andbene72:
@nopara73  select "Keep tabs"

NBitcoin/BuilderExtensions/P2MultiSigBuilderExtension.cs
```diff
@@ -75,29 +33,71 @@ public override Script CombineScriptSig(Script scriptPubKey, Script a, Script b)
...
+			if (aSigs == null)
 				return b;
 			var bSigs = PayToMultiSigTemplate.Instance.ExtractScriptSigParameters(b);
-			if(bSigs == null)
+			if (bSigs == null)
 				return a;
 			int pubkeyCount = 0;
 			TransactionSignature[] sigs = new TransactionSignature[para.PubKeys.Length];
-			for(int i = 0; i < para.PubKeys.Length; i++)
+			for (int i = 0; i < para.PubKeys.Length; i++)
```
NicolasDorier: can you fix those spaces after for and if ?

PR #204, opened 2016-11-04
Implement PutAsync

lontivero:
I didn't test this change because I am editing it online. Is it any reason this shouldn't be implemented?

### DIFF
--- a/NBitcoin/RPCTransactionRepository.cs
+++ b/NBitcoin/RPCTransactionRepository.cs
@@ -25,7 +25,7 @@ namespace NBitcoin
 
 		public Task PutAsync(uint256 txId, Transaction tx)
 		{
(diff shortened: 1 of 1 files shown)


NicolasDorier:
mmh actually, I'm not sure. I don't remember the original reason behind this interface.
I'm checking in my code when I get access back to my computer.

NicolasDorier:
Ok, I think I need to redesign this interface. I am confused myself by it.

NicolasDorier:
@lontivero PutAsync has the semantic of pure CRUD system, and no broadcast. I Included your method as BroadcastAsync though, and reverted your change to BlockrRepository. I kept what you did by adding a BlockrRepository.BroadcastAsync method

PR #205, opened 2016-11-04
implement transaction repository putasync method

lontivero:
This PR implements the PutAsync method for the BlockrTransactionRepository repository. I didn't test it with real transactions instead, i use the `tx/decode` service which shares exactly the same interface that the `tx/push` service.


NicolasDorier:
hey by the way, join us on https://gitter.im/MetacoSA/NBitcoin :)

PR #207, opened 2016-11-08
make FeeRate comparable

lontivero:
I have to play with fees coming fron different fee providers and the code looks better when FeeRate class implements IComparable.

The following is just pseudocode for illustrating the point:
```c#
var fee = await _feeProvider.GetFeeAsync();
fee = FeeRate.Max(fee, Configuration.MinFee);

if(fee > Configuration.MaxFee)
    throw new PanicException($"Too high fees for {tx.GetSerialized()} bytes = {fee}");
```


NicolasDorier:
awesome thanks

PR #208, opened 2016-11-14
Minor cleanup

andbene72:
Removed some unused using and fixed comment in correct VS format.

Nicolas,
this one is currently in Block.cs line 168
public const uint MAX_BLOCK_SIZE = 1000 * 1000;
Even there is no plan to increase it, I think that more correctly should be put in Network.cs in consensus or network class.

Also what is your opinion in splitting big files like Transaction.cs in multiple files matching class name ? IMHO improve readability. I can do it if you like the idea. I will clone files multiple times then commit and then I will delete parts so your review will be simpler.

With this pull I'm also testing the new git commit "verified" functionality.

Andrea


lontivero:
> public const uint MAX_BLOCK_SIZE = 1000 * 1000;
> Even there is no plan to increase it, I think that more correctly should be put in Network.cs in consensus or network class.

I agree with putting this in Network, however I think it could be a function instead of a constant:
```c#
int GetMaxBlockSize(long blockHeigh)
{
    return 1000*1000;
}
```

Sooner or later I think this will be:

```c#
int GetMaxBlockSize(long blockHeigh)
{
     if (blockHeigh < BIPXXX_WHEN_BLOCK_SIZE_CHANGED)
     {
        return 1000*1000;
     }
    else
    {
        return MAX_BLOCK_SIZE_SPECIFIED_BY_BIPXXX;
    }
}
```

I also agree with separating public classes in their own .cs files.

NicolasDorier:
Awesome, I wanted to keep Spendable because I am using it in NBitcoin.Indexer... but well, I will duplicate in NBitcoin.Indexer.

Concerning MAX_BLOCKSIZE, it should be removed. NBitcoin is not meant to be a full node library. The full node I am developping with Stratis has all the consensus rules properly implemented.

PR #210, opened 2016-11-15
Refactoring transaction.cs

andbene72:


NicolasDorier:
will surely merge that after reviewing. Class like Witness though are private to Transaction so I think are better as nested class.

dangershony:
This should probably go in to a new folder (maybe Primitives?)

lontivero:
This PR is from November 2016. IMO it should be closed.

PR #211, opened 2016-11-17
Add doc comments to BitcoinExtKey and ExtKey

sgryphon:
* Mostly just adding doc comments (which appear as help in VS Intellisense) to methods.
* Re-ordered a couple of methods to keep overloads together.
* Fixed a small typo/bug in the argument validation of one constructor of ExtKey (it was comparing a length against itself).

### DIFF
--- a/NBitcoin/BIP32/BitcoinExtKey.cs
+++ b/NBitcoin/BIP32/BitcoinExtKey.cs
@@ -24,21 +24,29 @@
 	}
 
 	/// <summary>
-	/// Base58 representation of an ExtKey
+	/// Base58 representation of an ExtKey, within a particular network.
 	/// </summary>
 	public class BitcoinExtKey : BitcoinExtKeyBase, ISecret
 	{
+		/// <summary>
+		/// Constructor. Creates an extended key from the Base58 representation, checking the expected network.
+		/// </summary>
 		public BitcoinExtKey(string base58, Network expectedNetwork = null)
 			: base(base58, expectedNetwork)
 		{
-
 		}
+
+		/// <summary>
+		/// Constructor. Creates a representation of an extended key, within the specified network.
+		/// </summary>
 		public BitcoinExtKey(ExtKey key, Network network)
 			: base(key, network)
 		{
-
 		}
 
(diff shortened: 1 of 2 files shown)


knocte:
2 nitpicks:
* You're using spaces, while the original code is using tabs.
* You're doing 3 changes, but mixing them in one commit, please separate each unit of work in one commit (one for adding docs, one for moving code, one for fixing typo/bug).

NicolasDorier:
That's ok for the 1 commit, but can you fix the spaces?

knocte:
> That's ok for the 1 commit

Haven't looked at the particular typo/bug fix, but in general, mixing changes is not a good approach, moving forward. The reason is this: if in the future you find that a mistake was done in the commit, you may want to revert it, but if it contains changes for different purposes, then you need to do a partial-revert, which is more tricky, and makes the history not so clean. (You could even automate this from a CI point of view: you create a test that you know it's a regression and you tell CI to bisect the commit that caused the regression and revert it. In that case you don't want your CI to revert changes that are not related to the regression.)

NicolasDorier:
yeah I know but that arrive rarely enough to bother. Particulary for a commit like this one that is trivial.
Also those two changes are about the same topic.

sgryphon:
Hadn't got around to checking the comments; will change to use tabs and update.

sgryphon:
knocte - I agree that the bug fix is a separate issue, so can put that in a separate branch (it was a trivial thing I noticed while adding the comments).

Are you okay if I leave changing the order of methods (just moving code, no actual changes) and adding doc comments, as they are both related to readability and neither actually makes any changes to the code? (If not, separating is fine too)

knocte:
(First I have to say that I'm not a maintainer of NBitcoin so I don't dictate what goes in or not, I'm just giving advice of best practices.) Separating in 2 commits instead of 3 seems good enough.

sgryphon:
Okay, I fixed the tabs and also removed the bug fix (to a separate pull request).  I amended the original commit, but also did a rebase on the latest master anyway (these shouldn't affect the PR at the branch level though).

PR #212, opened 2016-11-17
New rpcclient calls

sgryphon:
Additions to the RPCClient:

New code:
* Update listaccounts with optional parameters, to see watch only addresses
* Add getaddressesbyaccount
* Add receivedbyaddress`
* Updates listunspent return values to include spendable (indicates watch only) and redeemScript

Also, the RPCClient.cs file was very long, so I moved the wallet methods into a separate file (using partial class) -- I think separate, short, files are generally easier to understand (then using regions in one big file).

If you prefer the region style, then let me know and I will move it back and amend the PR.

Note that the move to partial class files was done as several separate steps, similar to the changes by andbene72 -- I cloned the entire file/class in one commit, then deleted the opposite parts from each file; this makes it easy to check going through the history that the clone was done correctly. Several other small changes, e.g. rearranging the methods into the API order, were done as separate commits.

For the listunspent change, I also added some tests to validate backwards compatibility, i.e. if the return value does not include a spendable field (pre 0.10.0 server) then it defaults to true. Not sure if the test format is acceptable (existing tests I saw load JSON from files).

Also, let me know if you would prefer separate small pull requests for each function as it is added, i.e. this is potentially 5 separate changes: the partial class file, the two method changes, and the two new methods. I am happy to do it separately if you prefer. (In general they are separate commits, just all done on one branch.)


knocte:
You're still using spaces instead of tabs :-( (At least what I saw in the first commit)

NicolasDorier:
very nice, can you fix the spaces ?

NicolasDorier:
This is awesome work thanks. By the way, for next tests I think you should take a look at how I test RPC. I have a very handy class which create a bitcoin full node on regtest and allow to mine and call RPC/REST/P2P easily. 

I think it is better doing testing with real RPCClient, as we will see RPC parsing breaking when the version of bitcoin core upgrade. Not mentioning the fact that we can also catch regression on Bitcoin Core itself not covered by their tests.

sgryphon:
Yes, I ran all the tests for the RPCClient, and it is good to test against the full node. 

I did remove the node from one test, because the test never used it -- it just loaded some JSON from a file and checked it parsed correctly. I guess when the node stuff was added it just got added to every single test (even though some didn't need it).

In the case of the listunspent, I was adding a property from 0.10.0 that wasn't in earlier versions, so wanted to check it wouldn't break with backwards compatibility. I think all the different levels of test could be important:
 - test against a real node is important;
 - test specific scenarios or combinations (if complicated to set up in a node, or just simply lots)
 - test specific backwards compatibility scenarios, i.e. against what previous versions of the node sent.

Thanks for merging. It is an excellent project, esp. with the support of .NET Core.

NicolasDorier:
Good point, indeed this test was testing backward compatibility. I will release version today.
Btw we have https://gitter.im/MetacoSA/NBitcoin for contributors, or people interested talking about NBitcoin.

Issue #209, opened 2016-11-14
RPC methods EstimateFee/Async change proposal/discussion


lontivero:
The RPCClient class contains sync and async methods which are very consistent in their signatures, what is valid for most of the methods with the exception of `EstimateFee` and `EstimateFeeAsync` methods because one of them returns a `FeeRate` instance and the other one returns a `Money` instance. _IMO both should retrun a FeeRate instance_. We can see the signatures below:

```c#
public FeeRate EstimateFee(int nblock)
public async Task<Money> EstimateFeeAsync(int nblock)
```

On the other hand, the synchronous version return `new FeeRate(Money.Zero)` when it receives a _-1_ as response. This forces the caller to check the returned value:

```c#
var fee = rpcClient.EstimateFee(6);
if(fee.FeePerK == Money.Zero){
   // do something
}
```

If programmer forget to check the value something wrong could happen. Probably the less dangerous thing could be to broadcast a transaction with no enough fee to be mined. This could, or not, be a problem.

On the other hand, the synchronous version return `Money.Parse(response.Result.ToString())` without validating the _special case -1_ and again this is, IMO, not okay because the **-1 value is an error code** and not a valid value. Again, IMO, this could end in an unexpected behaviour if the programmer doesn't check teh returned value. Think about scenarios like the follow one:

```c#
var fee = await rpcClient.EstimateFeeAsync(6); // returns -1
var valueOut = valueIn - fee; // unexpected behavior that can pass unnoticed for a long time
TxOut.Value = valueOut;
```

There is a reason why .NET developers don't check the returned values and it is because there is an unwritten rule in all methods' contracts: **if they cannot return what they have to, they have to throw an exception**. (bold is just to highligh what i think is important)

IMO this is specialy important when we handle money. 

I didn't send a PR before having a discussion about this and because these are breaking changes. I would like to know your opinion about this.

NicolasDorier:
Good catch on the signature which is obviously wrong, and I agree that both method should behave the same.

I don't agree with throwing an exception though. Not having fee estimation is an expected result, on which the user might decide to try again with a different nblock. Exception handling should not be used for processing this sort of logic.

I think that if it returns -1, then we should returns null.

lontivero:
I don't agree. Not having a fee estimation is an exceptional result, not an expected one. The same could be applied to other exceptional situations like not finding a file. 

[...]

PR #214, opened 2016-11-20
GetEstimatedFee & TryGetEstimatedFee methods

lontivero:
This PR is for issue https://github.com/MetacoSA/NBitcoin/issues/209 and marks `EstimateFee` and `EstimateFeeAsync` as Obsolete methods. It also creates the three new methods:  `GetEstimatedFee`, `TryGetEstimatedFee`  and `GetEstimateFeeAsync`.

### DIFF
--- a/NBitcoin.Tests/RPCClientTests.cs
+++ b/NBitcoin.Tests/RPCClientTests.cs
@@ -123,7 +123,7 @@ namespace NBitcoin.Tests
 		}
 
 		[Fact]
-		public void GetEstimatedFee()
+		public void EstimateFeeRate()
 		{
 			using (var builder = NodeBuilder.Create())
 			{
@@ -131,12 +131,12 @@ namespace NBitcoin.Tests
 				node.Start();
 				node.Generate(101);
 				var rpc = node.CreateRPCClient();
-				Assert.Throws<InvalidOperationException>(()=> rpc.GetEstimatedFee(1));
+				Assert.Throws<NoEstimationException>(()=> rpc.EstimateFeeRate(1));
 			}
 		}
 
 		[Fact]
-		public void TryGetEstimatedFee()
+		public void TryEstimateFeeRate()
 		{
 			using (var builder = NodeBuilder.Create())
 			{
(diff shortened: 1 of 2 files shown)


NicolasDorier:
just a nit, what about EstimateFeeRate and TryEstimateFeeRate, instead of get?

NicolasDorier:
ha and InvalidOperationException should not be used.

`The following sections show how some common cases in which in InvalidOperationException exception is thrown in an app. How you handle the issue depends on the specific situation. Most commonly, however, the exception results from developer error, and the InvalidOperationException exception can be anticipated and avoided.`

But here it can't be anticipated by the dev. I would say, create a new excpetion in RPC namespace like EstimatedFeeNotFoundException()

knocte:
> But here it can't be anticipated by the dev. I would say, create a new excpetion

Agreed.

> in RPC namespace like EstimatedFeeNotFoundException()

Is there any way to know the specific reason why it could not be estimated? If there are many possible reasons (instead of just one), ideally we would have different exception types for each.

NicolasDorier:
@knocte  only reason is that Bitcoin Core does not have enough data for it. Maybe we should say it in error message.

knocte:
Ok, but to not encourage ex.Message parsing, I would also add it on the exception name. Something like NotEnoughDataForEstimationException.

NicolasDorier:
@knocte there is only one reason the fee can t be estimated

NicolasDorier:
If it was not the case we would have details as prop of the exception,

NicolasDorier:
or subclass of EstimatedFeeNotFoundException so we don t break existing code.

knocte:
> or subclass of EstimatedFeeNotFoundException so we don t break existing code.

Ok but "estimated fee not found" doesn't seem correct wording. Either you can estimate or you cannot, but you don't "find" an estimation.

knocte:
How about `EstimationNotPossibleException`

NicolasDorier:
Strange, what about `NoEstimationException()` with error message saying "Try a lower estimation target or wait for Bitcoin Core to get more data."

knocte:
Ok, sounds better than EstimatedFeeNotFound

PR #223, opened 2017-02-03
added unity builds

mandelmonkey:
added two dll files need to use nbitcoin in unity


NicolasDorier:
closing, documentation to compile unity is in the README.

PR #225, opened 2017-02-28
Set max incoming connections

dangershony:


NicolasDorier:
I think that backlog != maxconnection, can you write a simple unit test to check it ? But I may be wrong.

dangershony:
@NicolasDorier how about this?
I wasn't sure how FindOrConnect should handle the max limit though.

NicolasDorier:
@dangershony I am not sure FindOrConnect make sense at all.
Find without connection would.

I think you can close the connection even before processing any message, in the socket Accept loop.

dangershony:
Yeah its odd as that would create an outbound connection and the list should only have inbound, 
I will change it no problem.

I was thinking on closing the connection earlier that means I'll have to call the method Utils.SafeCloseSocket(Socket) rather then the standard Node functionality

dangershony:
it seems FindOrConnect is used for testing so I will ignore it for now, unless it has a production purpose.

dangershony:
Can you also bump the nuget build?

NicolasDorier:
do when going back home tomorrow

dangershony:
Thanks I will update the fullnode

PR #230, opened 2017-03-31
Base32 address format for native witness outputs

lontivero:
## Intro
In the [this](https://github.com/sipa/bech32) Sipa's github repo, there is a BIP proposal describing a new base32 based address encoding format for native witness outputs. Also, in the same repo, there are reference implementations for Javascript, C and Python. 

## What's this PR for?
This PR is for adapting [my .NET implementation](https://github.com/lontivero/Bech32) to NBitcoin.

## Scope
* This PR only contains the Bech32 encoder/decoder code, its unit tests and minimal adaptions. 
* This doesn't make changes for using bech32 as default encoding format for `BitcoinWitPubKeyAddress`

## Notes
The Bech32 Encoder was included in the `Encoders` class. However it doesn't inherits from `DataEncoder` abstract class because the methods signatures don't match (Bech32 EncodeData and DecodeData accept two parameters instead of only one)

### DIFF
--- a/NBitcoin.Tests/Bech32Test.cs
+++ b/NBitcoin.Tests/Bech32Test.cs
@@ -1,0 +1,101 @@
+﻿using System;
+using System.Linq;
+using System.Text;
+using NBitcoin.DataEncoders;
+using Xunit;
+
+namespace NBitcoin.Tests
+{
+	[Trait("UnitTest", "UnitTest")]
+	public class Bech32Test
+	{
+		private static string[] VALID_CHECKSUM =
+		{
+			"A12UEL5L",
+			"an83characterlonghumanreadablepartthatcontainsthenumber1andtheexcludedcharactersbio1tt5tgs",
+			"abcdef1qpzry9x8gf2tvdw0s3jn54khce6mua7lmqqqxw",
+			"11qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqc8247j",
+			"split1checkupstagehandshakeupstreamerranterredcaperred2y9e3w"
+		};
+
+		private static string[][] VALID_ADDRESS = {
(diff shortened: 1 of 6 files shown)


NicolasDorier:
Later I might change this code so that `hrp` becomes a field of Bench32 encoder.

So you would do `Encoders.Bench32("yourHRP").Encode(data);` instead of `Encoders.Encode("yourHRP", data)`. That way it does not break `DataEncoder`. It also makes sense because hrp is a fixed parameter anyway.

NicolasDorier:
I made the above change, along with some optimization. I will soon create a `BitcoinDestination.Create` method whose goal is to replace `BitcoinAddress.Create` without breaking too much stuff.

lontivero:
ack

NicolasDorier:
I am still really mixed into changing `BitcoinAddress` into an interface or class unrelated to Base58. This would probably break a bunch of code, which is unfortunate.

Maybe even getting ride of Base58 base class entirely. I guess I will do that in future big NBitcoin release so I can break everything at once.

lontivero:
@NicolasDorier IMO a `BitcoinAddress` is not a Base58Data specialization.

I've faced this same design decision in my repo [BitcoinLite](https://github.com/lontivero/BitcoinLite) where I decided to break with the Bitcoin Core original design and make [`Address`](https://github.com/lontivero/BitcoinLite/blob/master/BitcoinLite/Address.cs#L21) an abstract class. All the `Base58Data` functionality can be available throught _composition_ instead of _inheritance_.

NicolasDorier:
I agree. I am just worried about breaking existing code.... but yeah I think I have no choice, not breaking the code would result in frankeistein design.

knocte:
Cannot you make new types/interfaces with the correct design, without removing the old classes/interfaces? Then you mark the old classes/interfaces as [Obsolete()], leave methods with both overloads for the old and new types (marking as Obsolete also the overloads for the old types), and done?

NicolasDorier:
@knocte I don't like it, so much code depends on `BitcoinAddress.Create`, also it has been documented everywhere. And it would be a mess to explain why BitcoinAddress should not be used anymore.
I think it is not a big deal to change BitcoinAddress to remove reliance on Base58Data, it should more likely only need a recompile.

PR #231, opened 2017-04-04

ptirouz:
BlockR retourne "fail" en status et non "error" sur BroadcastAsync
exemple : {"status":"fail","data":"Could not push your transaction!","code":500,"message":"Did you sign your transaction?"}

Désolé, c'est la premiere fois qur GitHub et j'ai oublié une parenthèse a la fin de ma modification
if(status != null && (status.ToString() == "error" || status.ToString() == "fail"))

### DIFF
--- a/NBitcoin/BlockrTransactionRepository.cs
+++ b/NBitcoin/BlockrTransactionRepository.cs
@@ -125,7 +125,7 @@ namespace NBitcoin
 				var json = JObject.Parse(result);
 				var status = json["status"];
 				var code = json["code"];
-				if(status != null && status.ToString() == "error")
+				if(status != null && (status.ToString() == "error" || status.ToString() == "fail")
 				{
 					throw new BlockrException(json);
 				}


NBitcoin/BlockrTransactionRepository.cs
```diff
@@ -125,7 +125,7 @@ public Task PutAsync(uint256 txId, Transaction tx)
 				var json = JObject.Parse(result);
 				var status = json["status"];
 				var code = json["code"];
-				if(status != null && status.ToString() == "error")
+				if(status != null && (status.ToString() == "error" || status.ToString() == "fail")
```
lontivero: Right. I've verified this:
```bash
~# curl -d '{"hex":"BAD-TX-HEX"}' http://btc.blockr.io/api/v1/tx/push
{"status":"fail","data":"Could not decode your transaction!","code":500,"message":""}
```

Then, you should remove the `status.ToString() == "error"` part because it is wrong.

NicolasDorier:
updated NBitcoin with this fix to 3.0.2.12

PR #234, opened 2017-04-06
WIP - Breaking Base58Data / BitcoinAddress hierarchy 

lontivero:
`BitcoinAddress` class no inherit from `Base58Data` class anymore. This is achieve thanks to a new (temporary?) interface that I called `IWalletData` by the moment. This interface contains the `Network` and `Type` properties and the methods `ToBytes()` and `ToWif()`.

Test pass green. I've messed this up with the VS tabulation config (sorry for that).

> Note: this is a work-in-progress PR that I'm sending for revision and discussion, this is not the final work.

### DIFF
--- a/NBitcoin.Tests/ColoredCoinsTests.cs
+++ b/NBitcoin.Tests/ColoredCoinsTests.cs
@@ -112,20 +112,20 @@ namespace NBitcoin.Tests
 			Assert.Equal("akB4NBW9UuCmHuepksob6yfZs6naHtRCPNy", colored.ToWif());
 			Assert.Equal(address.ScriptPubKey, colored.ScriptPubKey);
 
-			var testAddress = address.ToNetwork(Network.TestNet);
+			var testAddress = address.ToNetworkAddress(Network.TestNet);
 			var testColored = testAddress.ToColoredAddress();
 
 			Assert.Equal(Network.TestNet, testAddress.Network);
 			Assert.Equal(address.Hash, testAddress.Hash);
 
-			Assert.Equal(colored.ToNetwork(Network.TestNet), testColored);
+			Assert.Equal(colored.ToNetworkAddress(Network.TestNet), testColored);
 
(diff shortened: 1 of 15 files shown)


NicolasDorier:
So I think the interface should be `IBitcoinAddress` and have only `Network and ScriptPubKey`, it should also implement IDestination.

NicolasDorier:
Maybe not interface actually, just using BitcoinAddress as the base class.

lontivero:
I don't think so, if you remove the interface then all hell breaks loose and you have to make lots of changes in the `Network` class. I wil do it. I will remove the interface an see what happen.

NicolasDorier:
@lontivero I don't think it will be so terrible.

NicolasDorier:
do not hesitate to use Extension methods to make stuff that existed before still working.

lontivero:
I've removed the interface and test pass green. Anyway, it requires to be review carefully. I wait for feedback.

NicolasDorier:
1. BitcoinAddress should be limited to P2SH, P2PKH and later BECH. ExtKeys are not address
2. Can you not change the name of Methods ? (like not renaming ToNetwork to ToNetworkAddress)

lontivero:
I've undo changes in the bip32 files (ExtKeys).

I didn't find a way to avoid renaming ToNetwork extension method without an common interface implemented by Base58Data and BitcoinAddress. I'm sure that's the only way we can have a ToNetwork method for both classes, otherwise the C# language doesn't allow it.

NicolasDorier:
why did you changed some tests, removed the case for ColoredAddress in ToNetwork, and do not cover same stuff in CreateBase58Data ?

The goal is to get replace Base58Data by object. I think you did that because you do not have access to the Base58Type property? If that's the case we can use a IBase58Data interface instead which expose Base58Type so we do not break code.

I am concerned by the tests changing, the tests should not break normally.

NicolasDorier:
Closing this one, now BitcoinAddress is correctly not using Base58Data and support Bech32 addresses.

PR #237, opened 2017-04-11
Implement IEquatable<MerkleBlock> for MerkleBlock

nopara73:


NicolasDorier:
There is no tests, also I don't think it is a good idea. IEquatable is not implemented in Block or Transaction for example.

PR #241, opened 2017-04-13
Add new IReadOnlyNodesCollection for NodesCollection

Derrick-:
Create new read-only interface for NodesCollection

Re: https://github.com/stratisproject/StratisBitcoinFullNode/issues/115


NicolasDorier:
make sense

PR #244, opened 2017-04-26
Refator properties taking latest C# features into account

ArturKarbone:
Just want to improve readability of NBitcoin code, by leveraging latest language features. This is just a draft/sample where basically only properties have been refactored. My goal is to go further and contribute to the project by applying best practices I know and make the code readable as a book.

### DIFF
--- a/NBitcoin/BitcoinCore/Coins.cs
+++ b/NBitcoin/BitcoinCore/Coins.cs
@@ -1,8 +1,5 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace NBitcoin.BitcoinCore
 {
@@ -63,10 +60,10 @@ namespace NBitcoin.BitcoinCore
 
 		private void UpdateValue()
 		{
-            Value = Outputs
-                        .Where(o => !IsNull(o))
-                        .Sum(o=> o.Value);
-        }
+			Value = Outputs
+				.Where(o => !IsNull(o))
+				.Sum(o=> o.Value);
+		}
 
 		private bool IsNull(TxOut o) => o.Value.Satoshi == -1;
 		public bool IsEmpty => Outputs.Count == 0;
@@ -84,9 +81,9 @@ namespace NBitcoin.BitcoinCore
 			}
 		}
 
-        public int UnspentCount => Outputs.Count(c => !IsNull(c));
+		public int UnspentCount => Outputs.Count(c => !IsNull(c));
 
-        public bool Spend(int position, out TxInUndo undo)
+		public bool Spend(int position, out TxInUndo undo)
 		{
 			undo = null;
 			if(position >= Outputs.Count)
--- a/NBitcoin/BitcoinCore/Coins.cs
+++ b/NBitcoin/BitcoinCore/Coins.cs
@@ -63,9 +63,10 @@ namespace NBitcoin.BitcoinCore
 
 		private void UpdateValue()
 		{
-			Value = Outputs.Where(o => !IsNull(o))
-							.Select(o => o.Value).Sum();
-		}
+            Value = Outputs
+                        .Where(o => !IsNull(o))
+                        .Sum(o=> o.Value);
+        }
 
 		private bool IsNull(TxOut o) => o.Value.Satoshi == -1;
 		public bool IsEmpty => Outputs.Count == 0;
@@ -83,10 +84,9 @@ namespace NBitcoin.BitcoinCore
 			}
 		}
 
-		public int UnspentCount => Outputs.Where(c => !IsNull(c)).Count();
+        public int UnspentCount => Outputs.Count(c => !IsNull(c));
 
-
-		public bool Spend(int position, out TxInUndo undo)
+        public bool Spend(int position, out TxInUndo undo)
 		{
 			undo = null;
 			if(position >= Outputs.Count)
--- a/NBitcoin/BitcoinCore/Coins.cs
+++ b/NBitcoin/BitcoinCore/Coins.cs
@@ -96,7 +96,7 @@ namespace NBitcoin.BitcoinCore
 			undo = new TxInUndo(Outputs[position].Clone());
 			Outputs[position] = NullTxOut;
 			Cleanup();
-			if(Outputs.Count == 0)
+			if(IsEmpty)
 			{
 				undo.Height = nHeight;
 				undo.CoinBase = CoinBase;
@@ -119,7 +119,7 @@ namespace NBitcoin.BitcoinCore
 			{
 				uint nMaskSize = 0, nMaskCode = 0;
 				CalcMaskSize(ref nMaskSize, ref nMaskCode);
-				bool fFirst = Outputs.Count > 0 && !IsNull(Outputs[0]);
+				bool fFirst = !IsEmpty && !IsNull(Outputs[0]);
 				bool fSecond = Outputs.Count > 1 && !IsNull(Outputs[1]);
 				uint nCode = unchecked((uint)(8 * (nMaskCode - (fFirst || fSecond ? 0 : 1)) + (CoinBase ? 1 : 0) + (fFirst ? 2 : 0) + (fSecond ? 4 : 0)));
 				// version
@@ -242,7 +242,7 @@ namespace NBitcoin.BitcoinCore
 
 		// check whether the entire CCoins is spent
 		// note that only !IsPruned() CCoins can be serialized
-		public bool IsPruned => Outputs.Count == 0 || Outputs.All(v => IsNull(v));
+		public bool IsPruned => IsEmpty || Outputs.All(v => IsNull(v));
 
 
 		#endregion
--- a/NBitcoin/BitcoinCore/Coins.cs
+++ b/NBitcoin/BitcoinCore/Coins.cs
@@ -9,29 +9,10 @@ namespace NBitcoin.BitcoinCore
 	public class Coins : IBitcoinSerializable
 	{
 		// whether transaction is a coinbase
-		bool fCoinBase;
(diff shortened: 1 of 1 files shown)


knocte:
You changed tabs to spaces, please change back to tabs.

ArturKarbone:
@knocte , done

knocte:
There are still many whitespaces changes that you have introduced, which you should revert, before a proper review can take place.

Derrick-:
The space after if () and for () I like, but it's not consistent with the rest of the project.

lontivero:
@ArturKarbone I've reviewed your changes and they are mostly ok to me (there are just some details to review changes in the access modifiers for setters from public/internal to private). However, it is important follow the stablished coding style. Mixing (unintentionally) with spaces and tabs is something that many contributors have done (me too so, you are not the first one). 

On the other hand, it is also important to have in mind that reviewers need to see your changes as clear as possible, and in your last commit it is no possible to do that because the comparer shows every line red on the left panel and every line green on the right panel as two completely different content.

NBitcoin is a very friendly project to join and maintainer @NicolasDorier is also very open to many changes, but not to all ofcourse. Please fix the commit and wellcome!

ArturKarbone:
@Derrick- spaces after if()/for() have been removed
@lontivero - Thanks for the explanation. Completely agreeing with You now, that looking at commits are too hard since I changed from tabs/spaces back and forth. I suggest looking directly at file changes https://github.com/MetacoSA/NBitcoin/pull/244/files. 

Another suggestion to all the contributors is to share Visual Studio profile for the project ( with a link somewhere in Read.ME), to avoid such situations for newcomers. I think I will create an issue for that.

Thanks

knocte:
> spaces after if()/for() have been removed

Nitpick: you're still missing a space change in a `while` call.

> I suggest looking directly at file changes 

That's a good workaround to be able to review easily. But ideally your change should only be in one commit before it is merged. You can squash all your 3 commits in one with `git rebase -i HEAD~3` command (which will prompt an editor and you will have to mark the last commits as "f", for fixup).

In regards to the changes of the pull request, IMHO it's useful to have automatic properties (less boilerplate code), but I don't see much value in changing get-only properties from getter syntax to lambda syntax.

ArturKarbone:
I think its is ready for another round of review

NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -9,29 +9,10 @@ namespace NBitcoin.BitcoinCore
...
-			get
-			{
-				return fCoinBase;
-			}
-			set
-			{
-				fCoinBase = value;
-			}
-		}
+		public bool CoinBase { get; private set; }
```
lontivero: Don't change the setter visibility. It was ``public`` and shouldn't change to ``private``.

NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -88,58 +63,43 @@ public Coins(Transaction tx, int height)
...
 		}
 
-		public int UnspentCount
-		{
-			get
-			{
-				return vout.Where(c => !IsNull(c)).Count();
-			}
-		}
+		public int UnspentCount => Outputs.Where(c => !IsNull(c)).Count();
```
lontivero: I think that could be more easy to read if you change ``Outputs.Where(predicate).Sum()`` by ``Outputs.Sum(predicate)``

NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -88,58 +63,43 @@ public Coins(Transaction tx, int height)
...
-		}
-
-		public bool IsEmpty
-		{
-			get
-			{
-				return vout.Count == 0;
-			}
-		}
+		private bool IsNull(TxOut o) => o.Value.Satoshi == -1;
```
lontivero: **IsNull** is a private method and does not use ``this`` so, i think it could be declared as ``static``

NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -88,58 +63,43 @@ public Coins(Transaction tx, int height)
...
-			if(IsNull(vout[position]))
+			if(IsNull(Outputs[position]))
 				return false;
-			undo = new TxInUndo(vout[position].Clone());
-			vout[position] = NullTxOut;
+			undo = new TxInUndo(Outputs[position].Clone());
+			Outputs[position] = NullTxOut;
 			Cleanup();
-			if(vout.Count == 0)
+			if(Outputs.Count == 0)
```
lontivero: Here you could use the ``IsEmpty`` property:
Change this line by this other:
```c#
if(IsEmpty)
```

NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -159,9 +119,9 @@ public void ReadWrite(BitcoinStream stream)
 			{
 				uint nMaskSize = 0, nMaskCode = 0;
 				CalcMaskSize(ref nMaskSize, ref nMaskCode);
-				bool fFirst = vout.Count > 0 && !IsNull(vout[0]);
-				bool fSecond = vout.Count > 1 && !IsNull(vout[1]);
-				uint nCode = unchecked((uint)(8 * (nMaskCode - (fFirst || fSecond ? 0 : 1)) + (fCoinBase ? 1 : 0) + (fFirst ? 2 : 0) + (fSecond ? 4 : 0)));
+				bool fFirst = Outputs.Count > 0 && !IsNull(Outputs[0]);
```
lontivero: Here you also can change ``Output.Count > 0`` by ``!Empty``.

NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -272,52 +230,47 @@ private void CalcMaskSize(ref uint nBytes, ref uint nNonzeroBytes)
...
 		// check whether the entire CCoins is spent
 		// note that only !IsPruned() CCoins can be serialized
-		public bool IsPruned
-		{
-			get
-			{
-				return vout.Count == 0 || vout.All(v => IsNull(v));
-			}
-		}
+		public bool IsPruned => Outputs.Count == 0 || Outputs.All(v => IsNull(v));
```
lontivero: Same here, you can use ``Empty`` property

NBitcoin/BitcoinCore/Coins.cs
ArturKarbone: In that particular case CoinBase was assigned within the class, so I decided to not expose it.

NBitcoin/BitcoinCore/Coins.cs
ArturKarbone: Good catch. Will update

NBitcoin/BitcoinCore/Coins.cs
lontivero: How do you know that no one is using the property setter? If someone is using the setter an you make it private then, you are breaking the other guy project. NBitcoin is a library and you never know how the developers are using it. Anyway, it is a small change and it is not uncommon to break something from time to time.

ArturKarbone:
@lontivero just improved readability of Linq expressions and leveraged IsEmpty usage

NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -88,58 +63,43 @@ public Coins(Transaction tx, int height)
 
 		private void UpdateValue()
 		{
-			_Value = vout.Where(o => !IsNull(o))
-							.Select(o => o.Value).Sum();
-		}
+            Value = Outputs
```
NicolasDorier: space

NBitcoin/BitcoinCore/Coins.cs
NicolasDorier: can be done in separate PR. Actually maybe we can remove it completely.

NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -88,58 +63,43 @@ public Coins(Transaction tx, int height)
...
-		{
-			get
-			{
-				return vout.Where(c => !IsNull(c)).Count();
-			}
-		}
+        public int UnspentCount => Outputs.Count(c => !IsNull(c));
 
-		public bool Spend(int position, out TxInUndo undo)
+        public bool Spend(int position, out TxInUndo undo)
```
NicolasDorier: space

NicolasDorier:
cool, just some more spaces. Left, then I merge

NBitcoin/BitcoinCore/Coins.cs
ArturKarbone: Oh G, commited from another VS with different settings. just a sec

NBitcoin/BitcoinCore/Coins.cs
ArturKarbone: Just semantically it seems like private to me. If this is an issue lets cover it with unit tests later on.

NBitcoin/BitcoinCore/Coins.cs
ArturKarbone: Guys, i see two issues here. 


1. I think  that it is TxOut responsibility to tell wheter it is Null. 
2. **IsNull** seems to technical term here.  What business name would be more suitable here? **HasSatoshies** ? Other?

So instead of 
```
private void UpdateValue()
		{
			Value = Outputs
				.Where(o => !IsNull(o))
				.Sum(o=> o.Value);
		}
```

I suggest 

```
private void UpdateValue()
		{
			Value = Outputs
				.Where(o => !o.IsNull())
				.Sum(o=> o.Value);
		}
```

Or even better without negation
```
private void UpdateValue()
		{
			Value = Outputs
				.Where(o => o.HasSatoshies())
				.Sum(o=> o.Value);
		}
```

ArturKarbone:
@NicolasDorier done. Should I rebase intermediate commits?

NicolasDorier:
that is ok I can squash commit

PR #248, opened 2017-05-09
Add the BIP44 coin type to consensus

dangershony:


dangershony:
Hi Nicolas, if you dont want to add the BIP44 coin type its fine I will just merge this directly to NStratis.

PR #259, opened 2017-07-07
NBitcoin,NBitcoin.Tests: retarget from 4.6.2 to 4.6.1 (fix build w/vs2017)

### LINKED ISSUE #258: Can't compile with VS2017
I just cloned NBitcoin repo and I can't compile it with VS2017, I'm getting this error twice:

```
Error	MSB3644	The reference assemblies for framework ".NETFramework,Version=v4.6.2" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend.	NBitcoin.Tests	C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets	1111	
```

knocte:
Fixes #258

### DIFF
--- a/NBitcoin.Tests/NBitcoin.Tests.csproj
+++ b/NBitcoin.Tests/NBitcoin.Tests.csproj
@@ -6,9 +6,9 @@
 		<Description>The C# Bitcoin Library</Description>
 	</PropertyGroup>
   <PropertyGroup>
-	  <TargetFrameworks>net462;netstandard1.5</TargetFrameworks>
+	  <TargetFrameworks>net461;netstandard1.5</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


knocte:
@NicolasDorier this fixes the build for me. Any special reason it was targetting 4.6.2 instead of 4.6.1? Was NBitcoin using some feature of 4.6.2 not present in 4.6.1? If not, then this PR should be valid :)

PR #260, opened 2017-07-08
Transaction: make signing idiot-proof

### LINKED ISSUE #257: Transaction passes Check() but QbitNinja refuses to broadcast, with error: mandatory-script-verify-flag-failed (Operation not valid with the current stack size) (INVALID)
I'm using the Transaction class, adding some inputs this way:

```
        let transactionDraft = Transaction()
        for inputTx,index,value in inputs do
            transactionDraft.AddInput(inputTx, index) |> ignore
```

And adding some outputs this way:

```
        let dummyAddressForDraftTx = "1KsFhYKLs8qb1GHqrPxHoywNQpet2CtP9t"
        let destAddress = BitcoinAddress.Create(dummyAddressForDraftTx, Network.Main)
        let sumOfInputValues = inputs.Sum(fun (_,_,value) -> value)
        let txOutDraft = TxOut(Money(sumOfInputValues), destAddress)
        transactionDraft.Outputs.Add(txOutDraft)
        Console.WriteLine("num of bytes with draft: " + (transactionDraft.ToBytes().Length.ToString()))
        // then I show fee to user interface, user clicks ACCEPT or CANCEL, then:
        transaction.Outputs.Remove(transaction.Outputs.[0]) |> ignore
        if (transaction.Outputs.Count > 0) then
            failwith "Draft transaction should only contain one output"
        let destAddress = BitcoinAddress.Create(destination, Network.Main)
        let txOutAfterRemovingFee = TxOut(Money(Convert.ToInt64(amount * decimal 100000000) - minerFee.ValueInSatoshis), destAddress)
        transaction.Outputs.Add(txOutAfterRemovingFee)
        // then ask password and recipient to the user
        let encryptedPrivateKey = File.ReadAllText(account.AccountFile.FullName)
        let encryptedSecret = BitcoinEncryptedSecretNoEC(encryptedPrivateKey, Network.Main)
        let privateKey =
            try
                encryptedSecret.GetKey(password)
            with
            | :? SecurityException ->
                raise (InvalidPassword)
        transaction.Sign(privateKey)
```

Then I try to broadcast it this way:

```
        let transactionCheckResult = transaction.Check()
        if (transactionCheckResult <> TransactionCheckResult.Success) then
            failwith (sprintf "Transaction check failed with %s. Please report this bug" (transactionCheckResult.ToString()))
        let client = new QBitNinjaClient(Network.Main)
        let broadcastTask = client.Broadcast(transaction)
        broadcastTask.Wait()
        if not (broadcastTask.Result.Success) then
            failwith (sprintf "Error broadcasting transaction: %s (%s)"
                          broadcastTask.Result.Error.Reason
                          (broadcastTask.Result.Error.ErrorCode.ToString()))
        transaction.GetHash().ToString()
```

And then I get: Error broadcasting transaction: mandatory-script-verify-flag-failed (Operation not valid with the current stack size) (INVALID)

Am I doing something wrong?

knocte:
If transaction.Sign(key) was used, one could have
assumed that he was using an overload in which the
second parameter `bool assumeP2SH` would be optional.
However, there was no overload with a default value
for the second bool parameter; but there is an overload
that takes a `params` keyword for the second parameter,
which means that it can accept an empty array. But being
empty meant that it was not signing at all, even if
the method Sign() was called.

Make it idiot-proof so that if an empty array is passed,
it throws an ArgumentException. On top of that, make
Check() function also check if the transaction is signed.

Fixes #257

### DIFF
--- a/NBitcoin.Tests/transaction_tests.cs
+++ b/NBitcoin.Tests/transaction_tests.cs
@@ -2771,6 +2771,17 @@ namespace NBitcoin.Tests
 
 		[Fact]
 		[Trait("Core", "Core")]
+		public void tx_invalid_unsigned()
+		{
+			var someSignedAndExistingTrans = Transaction.Parse("010000000189632848f99722915727c5c75da8db2dbf194342a0429828f66ff88fab2af7d6000000008b483045022100abbc8a73fe2054480bda3f3281da2d0c51e2841391abd4c09f4f908a2034c18d02205bc9e4d68eafb918f3e9662338647a4419c0de1a650ab8983f1d216e2a31d8e30141046f55d7adeff6011c7eac294fe540c57830be80e9355c83869c9260a4b8bf4767a66bacbd70b804dc63d5beeb14180292ad7f3b083372b1d02d7a37dd97ff5c9effffffff0140420f000000000017a914f815b036d9bbbce5e9f2a00abd1bf3dc91e955108700000000");
+			Assert.Equal(TransactionCheckResult.Success, someSignedAndExistingTrans.Check());
+
+			var unsignedTrans = Transaction.Parse("0100000002b49ac2c0e715c06b323c301e8a200270b52e7fc9cf084a9bde6b911de543a91c0000000000ffffffff4d5ca0837c477711241b4ef540de5dbc7c1fd623013c8d3ec0c99f30601e57530000000000ffffffff0154351700000000001976a9142b1bf86dc6878152fd3994405c3f656c19e18e4988ac00000000");
+			Assert.Equal(TransactionCheckResult.Unsigned, unsignedTrans.Check());
+		}
+
+		[Fact]
+		[Trait("Core", "Core")]
 		public void test_Get()
 		{
 			byte[] dummyPubKey = TransactionSignature.Empty.ToBytes();
--- a/NBitcoin/NBitcoin.csproj
+++ b/NBitcoin/NBitcoin.csproj
@@ -6,7 +6,7 @@
 		<Description>The C# Bitcoin Library</Description>
 	</PropertyGroup>
 	<PropertyGroup>
-        <Version>4.0.0.8</Version>
+        <Version>4.0.0.9</Version>
     </PropertyGroup>
   <PropertyGroup>
     <TargetFrameworks>net461;netstandard1.3;netstandard1.1</TargetFrameworks>
--- a/NBitcoin/Transaction.cs
+++ b/NBitcoin/Transaction.cs
@@ -1448,6 +1448,11 @@ namespace NBitcoin
 		/// <param name="coins">Coins to sign</param>
 		public void Sign(Key[] keys, params ICoin[] coins)
 		{
+			if (!coins.Any())
+			{
+				throw new ArgumentException("coins array to sign must not be empty", nameof(coins));
+			}
+
 			TransactionBuilder builder = new TransactionBuilder();
 			builder.AddKeys(keys);
 			builder.AddCoins(coins);
@@ -1835,6 +1840,12 @@ namespace NBitcoin
 						return TransactionCheckResult.NullInputPrevOut;
 			}
 
+			foreach (var txin in Inputs)
+			{
+				if (Script.IsNullOrEmpty(txin.ScriptSig))
+					return TransactionCheckResult.Unsigned;
+			}
+
 			return TransactionCheckResult.Success;
 		}
 	}
@@ -1851,5 +1862,6 @@ namespace NBitcoin
 		DuplicateInputs,
 		NullInputPrevOut,
 		CoinbaseScriptTooLarge,
+		Unsigned,
 	}
 }


NicolasDorier:
If I merge that then all segwit stuff will break. In segwit, the scriptSig is empty. Even without segwit, it is possible to have an empty scriptSig which is still valid.

Moreover, `Sign` might be done on a transaction which involve multiple parties, one party might sign his part, leaving the rest blank.

If no coin is passed, then no signature should be done, this is how it should be. I have code which will just crash if you check empty array while I was not expecting to crash before.

knocte:
> If I merge that then all segwit stuff will break. In segwit, the scriptSig is empty. Even without segwit, it is possible to have an empty scriptSig which is still valid.

Mmm, ok, so what if we make a TransactionType enum, which is passed to Check() as a parameter? If TransactionType.SegWit is passed, then the Unsigned check doesn't happen.

> If no coin is passed, then no signature should be done, this is how it should be. I have code which will just crash if you check empty array while I was not expecting to crash before.

With `code` you mean external code to NBitcoin or internal? If it's internal, we could make that code use another overload that does the same but doesn't throw on empty array?

You know, I think making the API idiot-proof is important. I even watched your youtube video here last weekend (https://www.youtube.com/watch?v=X4ZwRWIF49w) and I spotted that you even did same initial mistake about using only one argument for Sign() and resulting in an unsigned transaction!

NicolasDorier:
@knocte I mean that if partial signature are expected to work, then I do not expect that if no signature happen there is something thrown. 

>  I spotted that you even did same initial mistake about using only one argument for Sign() and resulting in an unsigned transaction!

I think I should remove the Sign altogether, its behavior (without previous coins passed) is confusing as hell, for me as well. I think I took it from BitcoinJ originally.

`TransactionBuilder.Check()` tells you if the transaction is correctly signed if you pass all the coins.

knocte:
> I think I should remove the Sign altogether, its behavior (without previous coins passed) is confusing as hell, for me as well. I think I took it from BitcoinJ originally.

Sounds good, we mark it as [Obsolete] then? But then we need to supply a message, what to say in it? To use TransactionBuilder?

NicolasDorier:
Yes, or to use `Sign(keys, coins)` instead.

knocte:
> Yes, or to use Sign(keys, coins) instead.

Errrm, what? You just agreed to mark Sign(key) as [Obsolete], but you seem to be forgetting that Sign(key) usage resolves to Sign(key, params coins) where coins is an empty array. You cannot mark something as Obsolete suggesting to use the same thing :)

NicolasDorier:
sorry. What I meant is `Sign(key, bool)` should be obsolete. `Sign(key, params coins)` is how things should be signed.

knocte:
> What I meant is Sign(key, bool) should be obsolete. Sign(key, params coins) is how things should be signed.

That would neither have solved the problem I had the first time I tried to sign a transaction, nor would have solved the confusion you got in your youtube video. Precisely, if both you and me had used `Sign(key,bool)` we would not have had this problem.

NicolasDorier:
`Sign(key, coins)` should not throw. It is designed to try to sign as much as it can given the private keys.

`Sign(key, bool)` assume that you are put in your input's scriptSig the scriptPubKey or redeemScript of the previous output. Such information are actually in the `coins` already. 

By using `Sign(key, coins)` you would support P2SH, P2WSH, P2WPKH, P2PKH, and any combinaison of multi sig.

By using `Sign(key, bool)` you only dumb p2sh and p2pkh, and it ask you to exactract data you already have in your coins, and put the data in the scriptSig before getting called.

In other words, `Sign(key, bool)` requires you more work for supporting less cases.

knocte:
But `Sign(key, coins)` allows beginners shoot themselves in the foot when they don't pass any 2nd argument, thinking they would use default settings to sign.

knocte:
(And not only beginners, as demonstrated by the video ;) )

NicolasDorier:
ha got it. sorry. Maybe I should remove "parms" ?

NicolasDorier:
Yeah indeed you are right, calling `Sign(keys)` make no sense at all. the keyword `params` should be removed.

knocte:
> calling Sign(keys) make no sense at all

Right, that's why I wanted to throw the exception for empty array, as it would throw in this case.

> the keyword params should be removed.

Unfortunately, as far as I understand this would be an API break.

NicolasDorier:
I mean it is possible for one reason or the other that someone pass an empty coins array, because of some logic in their code at runtime, and this is fine, they don't expect it to crash.
However calling `Sign(keys)` is clearly something that never make sense at all. It is a no-op. So it should be prevented at compile time.

knocte:
@NicolasDorier not sure you saw my last message, since you sent one almost exactly at the same time as my last one.

To make sure we're on the same page: removing `params` is desirable, but it would be an API break. Are you willing to do API breaks? I'm guessing you're ok with it so long as you do a major-version change (e.g. from 4.x to 5.x).

As a sidenote, if you're willing to break API, I was thinking this: why not all the way and deprecate non-SegregatedWitness-ways of signing, in this break? This way, we would kill 2 birds with 1 stone.

knocte:
typo: `not all the way` -> `not go all the way`

NicolasDorier:
Non-segwitwitness signature are still widely in use in Bitcoin ecosystem, I guess we are stuck several years before we can deprecate it. For the params change, I am willing to break that. Fixing it is pretty obvious and not much people must depends on it.

knocte:
Closing since this has been address differently in PR#269.

knocte:
typo: address -> addressed

NicolasDorier:
new version pushed on nuget btw

knocte:
Great, we can target many more idiots now :P :)

NicolasDorier:
That's precisely the goal of a well designed library

PR #261, opened 2017-07-11
Transaction: make AddInput(Transaction,int) be more clever

knocte:
I assume this method was added to make it easier to add
inputs to a transaction without the need to create the
Input object itself. This is great, but to make it even
more great, we can even fill the ScriptSig property of
the new input for the API user so that they don't need
to mutate the inputs of the transaction after returning
it in order to be able to sign it later.

(If the ScriptSig of the input returned by this method
was not modified before trying to sign the transaction,
signing it would throw the exception below:)

```
System.InvalidOperationException: ScriptSigs should be filled with either previo
us scriptPubKeys or redeem script (for P2SH)
```

### DIFF
--- a/NBitcoin/NBitcoin.csproj
+++ b/NBitcoin/NBitcoin.csproj
@@ -6,7 +6,7 @@
 		<Description>The C# Bitcoin Library</Description>
 	</PropertyGroup>
 	<PropertyGroup>
-        <Version>4.0.0.8</Version>
+        <Version>4.0.0.9</Version>
     </PropertyGroup>
   <PropertyGroup>
     <TargetFrameworks>net461;netstandard1.3;netstandard1.1</TargetFrameworks>
--- a/NBitcoin/Transaction.cs
+++ b/NBitcoin/Transaction.cs
@@ -1426,6 +1426,7 @@ namespace NBitcoin
 			var @in = new TxIn();
 			@in.PrevOut.Hash = prevTx.GetHash();
 			@in.PrevOut.N = (uint)outIndex;
+			@in.ScriptSig = prevTx.Outputs.ElementAt(outIndex).ScriptPubKey;
 			AddInput(@in);
 			return @in;
 		}


NicolasDorier:
This is doing what is expected. The problem is the `Sign` method that I should have deprecated long ago in favor of signing with the `previousCoins`.
I should never need to do the hack of putting the previous `scriptPubKey` in the `scriptSig`. This does not work nor for segwit, nor for p2sh.

knocte:
> This does not work nor for segwit

Certainly, haven't tried segwit yet.

> nor for p2sh

Really? So this PR only works for me because I'm using `Sign(pk, false)` instead of `Sign(pk, true)`?

NicolasDorier:
Yes, please do not use `Sign(pk, bool)`. I know I taught to use it, but this should be deprecated.
You really need to use the `Coin` abstraction. The reason is that this Coin abstraction will allow you to sign segwit, p2sh, or p2pkh or multi sig with exactly the same code.

knocte:
I've been looking into using `Sign(pk, ICoin[])` instead of `Sign(pk, false)` in my code.

Before, I had this:

```
var trans = new Transaction();
var input = trans.AddInput(previousTx, indexInPreviousTx)
input.ScriptSig = previousTx.Outputs.ElementAt(indexInPreviousTx).ScriptPubKey;
trans.Sign(pk, false);
```

Now I would have this as far as I understand:

```
var trans = new Transaction();
var input = Coin(previousTx, indexInPreviousTx)
trans.Sign(pk, input);
```

Is this correct? I don't need to add it as an input if I use it in the Sign() method, right?

If this assumption is correct: then I have another problem, I cannot use EstimateFee() before signing because the estimation would happen before the transaction is complete.
If this assumption was incorrect, then it feels a bit redundant to have to specify the inputs when signing if I had to add the inputs previously. Why cannot Sign() find the proper already-added inputs that correspond to this pk?

NicolasDorier:
```
 I don't need to add it as an input if I use it in the Sign() method, right?
```
I never use this Sign, so I am not sure. I think you need to, because if under the hood it uses the transaction builder with Builder.SignTransaction, then the correct process is.

1. fetch the coins you can spend
2. select the one you want to spend, and put them as input of your transaction
3. add the keys
4. sign 

I advise you use the `TransactionBuilder` with `BuildTransaction(true)`, it can `EstimateFee`  properly estimate the size of the resulting transaction after everything is signed, and does not require you to fill out the input of your transaction.

NicolasDorier:
I close this one, as said, adding the scriptPubKey of the previous output in the scriptSig was a hack needed by `Transaction.Sign`. This has been deprecated as it does not support all cases. (won't work in p2sh and segwit)

PR #265, opened 2017-07-27
Feat/walletpassphrase

DanGould:
This implements the `walletpassphrase` rpc call as part of the RPCClient.Wallet. I'm not sure if my test is relevant/shows that the method works because I'm green to unit testing but I figured I'd try. I used this implementation in BreezeHub/BreezeServer and it worked unlocked an encrypted wallet.


NicolasDorier:
Nice

NicolasDorier:
Actually it does not compile, missing ';'.

PR #269, opened 2017-08-09
 Transaction: make Sign(key,coin) overloads stop using `params` keyword

knocte:
Having the params keyword meant that it was easy to shoot
yourself in the foot, as a newbie, by not providing any
element of the array at all, which would end up not signing
the transaction at all without throwing any exception (silent
failing). See discussion in:
https://github.com/MetacoSA/NBitcoin/pull/260

Also provide two new overloads for non-array variables, though
(this is harmless, as opposed to params keyword, because
we make sure at least there's one element provided).

As this is an API breaking change, version is bumped to 5.0.

### DIFF
--- a/NBitcoin/NBitcoin.csproj
+++ b/NBitcoin/NBitcoin.csproj
@@ -12,7 +12,7 @@
 		<RepositoryType>git</RepositoryType>
 	</PropertyGroup>
 	<PropertyGroup>
-        <Version>4.0.0.16</Version>
+        <Version>4.0.0.17</Version>
     </PropertyGroup>
   <PropertyGroup>
     <TargetFrameworks>net461;net452;netstandard1.3;netstandard1.1</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


NicolasDorier:
No need to bump to 5.0.
Changing this code will break no existing code.

There is no reason that anyone was using `Sign(keys, c1,c 2)`, except for testing stuff.

knocte:
Even if you believe a specific API doesn't have any consumers, bumping version is always the right thing to do when breaking it.

How about 4.1? Although according to http://semver.org/ , an API break should always pose a major version change actually.

NicolasDorier:
I think "every breaking change requires major version update" is overkill, I keep major version when there is really a big change that is likely to impact people (last version bump was BitconAddress not depending on Base58Data)

knocte:
Ok then I change to 4.1?

NicolasDorier:
Just to 4.0.0.17

knocte:
Fair enough. Updated.

PR #298, opened 2017-10-30
Filter by the entire endpoint

dangershony:
I suspect this is a bug, if not please close this PR.


NicolasDorier:
A bug, thanks a lot!

PR #301, opened 2017-11-01
Enable public function for GetBase58Type

quartsystem:
Useful when importing a string value from external support (ex qr code or NFC)

### DIFF
--- a/NBitcoin/Network.cs
+++ b/NBitcoin/Network.cs
@@ -981,7 +981,7 @@ namespace NBitcoin
 			return new BitcoinScriptAddress(base58, this);
 		}
 
-		private Base58Type? GetBase58Type(string base58)
+		public Base58Type? GetBase58Type(string base58)
 		{
 			var bytes = Encoders.Base58Check.DecodeData(base58);
 			for(int i = 0; i < base58Prefixes.Length; i++)


NicolasDorier:
So there is a good reason why it is private: QRCode and NFC does not have necessarily a Base58 type. It might be Bech32.

For this reason I created `Network.Parse<T>`. Typically you can use `Network.Parse<BitcoinAddress>`.

quartsystem:
`Network.Parse<T>` is a simple way to perform a conversion, but there is no easy way to map a string to the objectType.

I would propose two very useful public functions when importing large quantities of objects from a storage between the *address and *key types.
These functions can be very performance.

```
public Base58Type? GetBase58Type (string base58)
public Bech32Type? GetBech32Type (string bech32)
```

I'm developing an excel add-in to give the user the ability to manipulate seeds, addresses and principal entities of library, as well as other tracking capabilities (balance, conversions, etc.).
These two functions can facilitate my development.

I take the opportunity to thank you for the great work you are doing. Without this library and its documentation I would never have entered the bitcoin/blockchain world.

NicolasDorier:
Glad you like NBitcoin :)

I don't understand why `Network.Parse<IBase58Data>(st).Type` or `Network.Parse<IBech32Data>(st).Type` would not do the trick?

quartsystem:
You're right ... the night I do not have a sharp eye.
Thank you :)

mattheys:
@NicolasDorier A TryParse would be greatly appreciated.

NicolasDorier:
Can I close this one ?

quartsystem:
yes

2017-11-27 10:47 GMT+01:00 Nicolas Dorier <notifications@github.com>:

Issue #304, opened 2017-11-13
new BitcoinWitPubKeyAddress fails without expected network


nopara73:
Sorry for not fixing it myself, should be something trivial, I am just in the middle of something. Will get back to it if I don't forget.  

```cs
new BitcoinWitPubKeyAddress("tb1qr5d68t6qm8t2n7ch4nph3ha4prztteuw98ewda", Network.TestNet); // SUCCESS
new BitcoinWitPubKeyAddress("tb1qr5d68t6qm8t2n7ch4nph3ha4prztteuw98ewda"); // System.NullReferenceException
```

NicolasDorier:
good catch, might take a look today

NicolasDorier:
There was the same bug a few line below https://github.com/MetacoSA/NBitcoin/commit/5ca39bf8d031c4e288f0a2c0f33305bbd476ff98

PR #306, opened 2017-11-21
fix for issue #304

lontivero:


NicolasDorier:
Nice catch

NicolasDorier:
Release 4.0.0.43 including this fix.

NBitcoin/BitcoinSegwitAddress.cs
```diff
@@ -34,6 +34,7 @@ private static string Validate(string bech32, ref Network expectedNetwork)
 					var data = encoder.Decode(bech32, out witVersion);
 					if(data.Length == 20 && witVersion == 0)
 					{
+						expectedNetwork = network;
```
NicolasDorier: Not needed, expectedNetwork is not used at this point.

NicolasDorier: Oh actually it is ref... it is needed.

PR #310, opened 2017-11-25
Rpc/importmulti

dimsumcode:
Implement RPC method "importmulti" (https://bitcoin.org/en/developer-reference#importmulti). Write several unit tests.  I still have more unit tests to write and will commit those in the near future.


dimsumcode:
I think it's failing because importmulti was introduced in version Bitcoin Core 0.14.0
But the NodeBuilder is hard-coded to a 0.13 node instance.  I don't think there's much I can do to make the automated tests work. Any feedback would be appreciated.

dimsumcode:
OK. I understand better how the NodeBuilder works now. I now specify version 0.15.1 and my automated build is working, so I think it should be fine for merging.

NicolasDorier:
awesome, thanks! Yes we might bump the default to 0.15.1 one of these days.

dimsumcode:
Thank you for creating this project!  I plan to work on ValidateAddress next.

NicolasDorier:
@dimsumcode I refactored quite a bit of your code for making the response fully typed and using exception for errors, and released a new version of the nuget package. Take a look when you can.

dimsumcode:
@NicolasDorier Yes I saw that and I've already merged that into my local repo.  Thanks for reviewing it and making the changes.  I'm fine with those improvements.

PR #317, opened 2017-12-07
Update xunit.runner.visualstudio

nopara73:
Tests were not running. I updated the runner and now they are running properly.  

What I also tried:  

1. I installed .NET Framework development for VS, it still didn't work.

2. However I was able to make it work with the old runner version if I edited the following line in the test project's csproj:

```xml
<TargetFrameworks>net461;netcoreapp1.1</TargetFrameworks>
```

I added: `netcoreapp2.0` - did not work.  
I replaced `netcoreapp1.1` with `netcoreapp2.0` - did not work.  
I removed both `net461;netcoreapp1.1` and replaced with `netcoreapp2.0` - did work.

### DIFF
--- a/NBitcoin.Tests/NBitcoin.Tests.csproj
+++ b/NBitcoin.Tests/NBitcoin.Tests.csproj
@@ -23,7 +23,7 @@
 		<PackageReference Include="System.Net.Requests" Version="4.3.0" />
 		<PackageReference Include="System.Net.Http" Version="4.3.2" />
 		<PackageReference Include="xunit" Version="2.2.0" />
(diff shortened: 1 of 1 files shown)


NicolasDorier:
so with `2.3.1` it works with both?

NicolasDorier:
it makes sense to use netcoreapp2.0 now.

nopara73:
With 2.3.1, everything works. Do you want me to add 2.0 or replace `1.1 with 2.0?

NicolasDorier:
replace 1.1 with 2.0 and verify that 2.3.1 still work.

nopara73:
@NicolasDorier  Replaced. Verified. Works.

PR #320, opened 2017-12-08
Fix tests

nopara73:
Fixed the feerate doesnt return null at error

### DIFF
--- a/NBitcoin/RPC/RPCClient.cs
+++ b/NBitcoin/RPC/RPCClient.cs
@@ -1220,7 +1220,7 @@ namespace NBitcoin.RPC
 			var feeRateDecimal = resultJToken.Value<decimal>("feerate"); // estimate fee-per-kilobyte (in BTC)
 			var blocks = resultJToken.Value<int>("blocks"); // block number where estimate was found
 			var money = Money.Coins(feeRateDecimal);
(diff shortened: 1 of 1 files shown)


NicolasDorier:
@nopara73 I am wondering if gettxout does not break because I change version to 0.15.1. Core may have done breaking changes.

nopara73:
It's `CanGetUTXOsMempool`, not `CanGetTxOutFromRPCAsync`. I confused the two in my first comment.  

The problem is here: `Assert.Equal(1, (int)utxos.Outputs[0].Version);`.  
It expects version 1, but gets version 0.  
I never met with `UTxOut` version property before, so I'm not sure what that does.  
Test runs properly by uncommenting the line.

NicolasDorier:
No idea what is this property.

nopara73:
Not sure if it's the same version being talked about here: https://bitcoin.stackexchange.com/questions/53271/parse-utxo-of-a-transaction-from-chainstate

> Where 01 is the version, 9070 is the nCode which tells if its a coinbase tx

nopara73:
By this logic, indeed 1 should be expected, since it's a coinbase utxo we are checking there. But I'm really reaching the edge of my knowledge here.

PR #331, opened 2018-01-11
prefer use ChainBase

micky53:


NicolasDorier:
I agree it would make sense in theory. However, I plan to make `ChainBase` obsolete and remove it in a future version, so I prefer people do not use it directly. I have not found a single usecase where it was useful to have anything else than `ConcurrentChain`.

micky53:
It's sad to hear this. I would like to implement my class derived from `ChainBase`, which will store the headers in the database and load them as needed. Thus, when you start the program, it will not be necessary to spend time reading from the disk more than 100 MB.

NicolasDorier:
@micky53 that was my intention in the beginning. 

However clients of this class expect `GetBlock` to not fail or have any IO (need to be fast), so having it back by your own database would break lot's of assumptions.

A database backed version should have async functions, and probably another interface and another class. I don't know how a good interface would look like though, so if you have ideas, I am open to it.

You don't want 1 DB request per header though, it would kill perf. (`GetBlockAsync`)

micky53:
I have no ideas. Cancel polmostechest.

PR #337, opened 2018-02-11
Golomb-compressed Sets

lontivero:
Adds Golomb-compressed Sets which is needed for implementing the
BIP158 - Compact Block Filters for Light Clients. 

Classes for creating and querying Golomb-compressed Sets (GCS), a statistical compressed data-structure. The idea behind this implementation is using it in a new kind of Bitcoin light client that uses GCS instead of bloom filters.

This projects is based on the [original BIP (Bitcoin Improvement Proposal)](https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki)
and [the Olaoluwa Osuntokun's reference implementation](https://github.com/Roasbeef/btcutil/tree/gcs/gcs)

An standalone project which doesn't depend on NBitcoin is [available here](https://github.com/lontivero/gcs/)

## Motivation of the Pull Request
@nopara73 and I think this code should be part of NBitcoin in order to allow users to work with GCS. This is the first of a set of PRs that we will do as soon we have new funtionalities done.

## How to use it
The example below shows it can be used in a similar way that bloom filters.

```c#
var cities = new[] { "New York", "Amsterdam", "Paris", "Buenos Aires", "La Habana" }
var citiesAsByteArrar = from city in cities select Encoding.ASCII.GetBytes(city);

// A random key
var key = new byte[] { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 };

// The false possitive rate (FPR) is calculated as: FPR = 1/(2**P)
var P = 16
var filter = GCSFilter.Build(key, P, citiesAsByteArrar);

// The filter should match all ther values that were added
foreach(var city cities)
{
	Assert.IsTrue(filter.Match(city, key));
}

// The filter should NOT match any extra value
Assert.IsFalse(filter.Match(Encoding.ASCII.GetBytes("Porto Alegre"), key));
Assert.IsFalse(filter.Match(Encoding.ASCII.GetBytes("Madrid"), key));
```

### DIFF
--- a/NBitcoin.Tests/Bip158_tests.cs
+++ b/NBitcoin.Tests/Bip158_tests.cs
@@ -10,6 +10,7 @@ namespace NBitcoin.Tests
 	public class GolombRiceFilterTest
 	{
 		[Fact]
+		[Trait("UnitTest", "UnitTest")]
 		public void BuildFilterAndMatchValues()
 		{
 			var names = from name in new[] { "New York", "Amsterdam", "Paris", "Buenos Aires", "La Habana" }
@@ -52,6 +53,7 @@ namespace NBitcoin.Tests
 		}
 
 		[Fact]
+		[Trait("UnitTest", "UnitTest")]
 		public void FalsePositivesTest()
 		{
 			// Given this library can be used for building and query filters for each block of 
@@ -139,6 +141,7 @@ namespace NBitcoin.Tests
 	public class FastBitArrayTest
 	{
 		[Fact]
+		[Trait("UnitTest", "UnitTest")]
 		public void GetBitsTest()
 		{
 			// 1 1 1 0 1 0 1 1 - 1 0 1 0 1 0 1 1 - 1 0 1 0 1 1 1 0 - 1 0 1 0 1 1 1 0 
@@ -169,6 +172,7 @@ namespace NBitcoin.Tests
 		}
 
 		[Fact]
+		[Trait("UnitTest", "UnitTest")]
 		public void SetRandomBitsTest()
 		{
 			var barr = new FastBitArray(new byte[0]);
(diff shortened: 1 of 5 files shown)


nopara73:
Some things to consider: 

1. The `Filter` should be named `GolombRiceFilter`, since NBitcoin already has a `BloomFilter` class in it.  
2. `BloomFilter` class is in the root directory, so the `GolombRiceFilter` class should be placed to the main directory, too.
3. It may be better to put this into the SPV folder, though I'm not sure about this one.

@NicolasDorier What do you think?

NicolasDorier:
This is awesome.

Not the SPV folder. The SPV folder is for the SPV wallet related code. (Which I think to deprecate or put into a separate lib later)

I agree for name change though. Will review the API a bit later.

nopara73:
I have another suggestion. The BIP says:

> The parameter P MUST be set to 20. This value was chosen as simulations show that it minimizes the bandwidth utilized, considering both the expected number of blocks downloaded due to false positives and the size of the filters themselves.

Where you give the user full power over the P, which is fine, but maybe you want to add a default value for the P, as 20, since it's globally agreed upon in Bitcoin.

NicolasDorier:
Can you add trait "UnitTest" to your tests so it can be picked up by the CI ? (check other tests for example)

NicolasDorier:
one test is broken

NicolasDorier:
The tests were great! :)

NBitcoin.Tests/Bip158_tests.cs
```diff
@@ -0,0 +1,264 @@
...
+				}
+			}
+
+			// Get bits in the same int.
+			Assert.Equal((ulong)0b111, barr.GetBits(0, 3));
+			Assert.Equal((ulong)0b10111, barr.GetBits(0, 5));
+			Assert.Equal((ulong)0b01010111010, barr.GetBits(3, 11));
+
+			// Get bits in cross int
+			Assert.Equal((ulong)0b101110101110101, barr.GetBits(24, 16));
```
NicolasDorier: oh I did not know you could do `0b` before!

NicolasDorier:
Can you remove the `Console.Write` then I think we are good!

PR #351, opened 2018-02-28
Update Tests' default Core version to 0.16.0

nopara73:


NicolasDorier:
test fail

nopara73:
I don't know what's wrong, test logs in appveyor is useless.

NicolasDorier:
You can see which tests fails, seems `getinfo` which got deprecated on 0.16.0

lontivero:
Do you want to keep that test? If yes, what about replace the call to `getinfo` by one to `getblockchaininfo` or `getmininginfo`?

```c#
[Fact]
public void CanSendCommand()
{
	using(var builder = NodeBuilder.Create())
	{
		var rpc = builder.CreateNode().CreateRPCClient();
		builder.StartAll();
		var response = rpc.SendCommand(RPCOperations.getmininginfo);
		Assert.NotNull(response.Result);
		var copy = RPCCredentialString.Parse(rpc.CredentialString.ToString());
		copy.Server = rpc.Address.AbsoluteUri;
		rpc = new RPCClient(copy, null as string, Network.RegTest);
		response = rpc.SendCommand(RPCOperations.getmininginfo);
		Assert.NotNull(response.Result);
	}
}
```

NicolasDorier:
nop we don't want to keep that test, replacing is fine.

PR #363, opened 2018-03-08
Add RPC waitfornewblock call

lontivero:
This PR is for adding a new RPC procedure call `waitfornewblock'.


lontivero:
Oh! I didn't know it. It make sense don't merge this.

Issue #371, opened 2018-03-14
Improving Altcoin support


NicolasDorier:
So @Kukks came with an idea which seem a viable path for increasing altcoin support in NBitcoin.

# Current situation 

Current altcoin support is possible only if the transaction, block, blockheader, and way of calculating the block headerhash are similar to bitcoin. (like Litecoin, Viacoin, or BCash)
This is done by building a `Network` through the `NetworkBuilder` class.
One can specify a way to calculate the `PoWHash` algorithm as well so it is possible to check PoW on altcoins not using SHA256 as their PoW algorithm.

# Proposed change

In order to not break any existing code, @Kukks proposed a solution which might work and involve minimal work.

For block header for example, the Network class would have a `BlockHeaderFactory` instance, responsible for parsing a Block Header from bytes.
Non-Bitcoin block header would be subclass of block header, overriding `GetHash`, `GetPowHash` and  `ReadWrite`.
A block header instance would then know exactly, without any outside information: 

* How to calculate its hash, 
* How to calculate its PoW Hash, 
* How to serialize itself,

# Consequence

It means that in order to correctly `deserialize` a blockheader we would need to have a Network instance at hand. And luckily, this should be already the case. (Node class is an example)

Place deserializing a block header and not having Network/BlockHeaderFactory at hand should assume Bitcoin format and marked as issues in github for later refactoring.

# Breaking change

The `GetPoWHash` of `BlockHeader` currently takes a `Network` parameter. This is not needed anymore as the BlockHeader would get the logic for `GetPowHash` calculation inside itself.
We can just keep the current `GetPowHash(Network)`  and mark as obsolete.
Then having the real `GetPowHash()` without Network parameter

# Note

We would use the following plan for `BlockHeader/BlockHeaderFactory`, `Block/BlockFactory`, `Transaction/TransactionFactory`.

I hope we can get one separate PR for each of them, starting by `BlockHeader/BlockHeaderFactory`.

@Kukks @knocte @dangershony @bokobza @Aprogiena @nopara73

nopara73:
Concept ACK. I think it's a great solution. Altcoiners may need to go for the extra mile to understand it, but as far as I understand it only minimally [messes with the Bitcoin user space](https://lkml.org/lkml/2012/12/23/75) (IMO `GetPowHash` and `BitcoinStream` are not often used), which is the most important.

Aprogiena:
[...]

PR #379, opened 2018-03-27
Feature/altcoin support

Kukks:
This is my initial work for #371. Please provide feedback on where to improve and change. :)


NBitcoin/Block.cs
```diff
@@ -9,9 +9,39 @@
...
+			return new BlockHeader(hex);
+		}
+	}
+
+	public interface IBlockHeaderFactory
+	{
+		IBlockHeader<ICoinStream> CreateNewBlockHeader();
+	}
+
+	public interface IBlockHeader<in TCoinStream> : ICoinSerializable<TCoinStream> where TCoinStream: ICoinStream
```
dangershony: I am not sure I understand whats `ICoinStream` for.

The notion of `IBlockHeader` is good I would be happier if it was an `abstract` called `BlockHeaderBase` but I guess it means many changes.

dangershony:
This is a very good start 👍

NBitcoin/Block.cs
Kukks: I went the interface route instead of a base class to leave Bitcoin as the first class integrated coin in the lib. It also minimizes code changes like you said. ;)

While abstracting `IBitcoinSerializable` to allow different  `ReadWrite`  methods on a blockheader level I realized that it relied completely on `BitcoinStream` which may or may not be too coupled to btc to make all blockheaders utilize `BitcoinStream` in `ReadWrite`. I'm probably wrong though,  not very sure what that part does.

NicolasDorier:
Why introducing an interface ` IBlockHeader<ICoinStream>`? I would prefer not doing that because replacing `new BlockHeader()` with this will break tons of code. IMHO it should returns `BlockHeader`, and other headers should just subclass it.

NicolasDorier:
This and AppVeyor has a compile issue on windows.

NBitcoin/Block.cs
NicolasDorier: I would agree if we could break existing code, but we want to prevent this, I think we should not try to abstract this and just subclass `BlockHeader` for the time being.

Kukks: I'll change this now. 

Maybe it's because I haven't used NBitcoin as extensively as you but I can't see why abstracting NBitcoin's BlockHeader would be breaking: Everyone can still use BlockHeader as before and it will still be the default, it's only when they try to use the new functionality that usage of the interface would be needed and visible.

Kukks:
@NicolasDorier breaking build was just an unused import from what I could see.

Kukks:
@NicolasDorier  I've removed the interfaces in question

NicolasDorier:
Sounds good! Have you start looking to where we can replace existing reference to `new BlockHeader()` ?

Kukks:
Checking that out now. Do we want some sort of protection to disallow users from instantiating a new BlockHeader directly?  I'll be adding a way to do that in the next commit, just let me know what you think then.

Travis is failing but I'm not sure why, any insight on that @NicolasDorier?

Kukks:
@NicolasDorier I might have went a bit too deep into the rabbit hole. :P As soon as I marked the contsructors for `BlockHeader` protected, there were 3 compilation errors, in `Block.cs, RestClient.cs and RPCClient.cs`.  This makes all sort of methods need an instance of the block header factory just to specify the block header for it. I'll commit as soon as I iron out the compile errors and then we can review how to scale it down

Kukks:
Well, that was very messy. Any advice?

NicolasDorier:
So `RestClient` and `RPCClient` knows the `Network`, so you can get the factory from here.

For `Block`, it makes sense that no Factory is used. Each class derived from `Block` knows the concrete instance of their header anyway.
I think `Block` should have a `CreateNewBlock()` method in the factory as well.

NicolasDorier:
So `Block` should not have to know the Network. `Block` will knows how to serialize itself because it will be created by the factory anyway.

Kukks:
@NicolasDorier  

> So RestClient and RPCClient knows the Network, so you can get the factory from here.
I can't seem to find any references or access to the Network from `RestClient `

> I think Block should have a CreateNewBlock() method in the factory as well.
I thought this was out of scope for this issue? 

The remaining warnings are: 
image

Kukks:
`RestClient` does not have any references to the Network. Am I missing something @NicolasDorier ?

NicolasDorier:
@dangershony do you think the approach @Kukks took here is sufficient for complete support of say Stratis on NBXplorer ?

NicolasDorier:
@Kukks can you rebase? I fixed the failing tests.

NicolasDorier:
@Kukks do you have an altcoin on which we can experiment this new API on NBXPlorer with this PR?

NBitcoin/RPC/RPCClient.cs
```diff
@@ -987,18 +987,18 @@ public Block GetBlock(int height)
...
 
 		public async Task<BlockHeader> GetBlockHeaderAsync(uint256 blockHash)
 		{
 			var resp = await SendCommandAsync("getblockheader", blockHash.ToString()).ConfigureAwait(false);
-			return ParseBlockHeader(resp);
+			return ParseBlockHeader(resp, Network.Consensus.BlockHeaderFactory.CreateNewBlockHeader());
 		}
 
-		private static BlockHeader ParseBlockHeader(RPCResponse resp)
+		private static BlockHeader ParseBlockHeader(RPCResponse resp, BlockHeader blockheader)
```
NicolasDorier: I think it would be easier to just remove `static` and directly use the factory inside.

NBitcoin/Block.cs
```diff
@@ -418,27 +463,28 @@ public void UpdateMerkleRoot()
...
-			return Check(null);
+			return CheckMerkleRoot() && Header.CheckProofOfWork();
 		}
 
 		/// <summary>
 		/// Check proof of work and merkle root
 		/// </summary>
 		/// <param name="consensus"></param>
 		/// <returns></returns>
+		[Obsolete]
```
NicolasDorier: Use `Check()` instead...

NicolasDorier:
I can work on it once I merged this, there is some place where we don't call `new BlockHeader()` but `new T()` via generics that I need to fix.

Kukks:
@NicolasDorier I have a branch on NBXplorer integrating a dash fork called Straks. The Network class for it is ready @ https://github.com/Kukks/NBXplorer/tree/feature/straks

NicolasDorier:
So should be ready to merge after the tests pass? There is some places I should update probably though (like the P2P messages parsing)

Kukks:
It would seem so! Hopefully we can keep at it and do Block and Transaction after this trial run. :)

NicolasDorier:
let's give it a try.

NicolasDorier:
Ok so, I played with it a bit and I decided I was wrong, the easiest way to make this possible is to pass the `Network` (or at least the factories) directly to `BitcoinStream`.

This is not as bothersome that I thought, because setting this property is only needed we deserialize, not when we serialize.

NicolasDorier:
I found a way to refactor by being even more abstract. Will commit later today.

NicolasDorier:
@Kukks let me know when you can try the new design. Shitcoins are in NBitcoin.Altcoins in this repo, there is some guide there to add you own crypto and do quick tests.

We can still break stuff if needed based on feedback if we really need.

Issue #383, opened 2018-04-05
RestClient should have a reference to Network


NicolasDorier:
https://github.com/MetacoSA/NBitcoin/pull/379#issuecomment-377838834

PR #384, opened 2018-04-05
RestClient keeps a network reference

lontivero:
@NicolasDorier, is this what you need for **RestClient should have a reference to Network** #383?


NicolasDorier:
Yep! thanks :)

PR #385, opened 2018-04-07
Wordlist Portuguese Brazil BIP39

alissonsolitto:
Implementation of the list of words in the Portuguese-Brazilian language.
The word list was obtained at this link:

Issue: https://github.com/bitpay/bitcore-mnemonic/issues/52
WordList: https://github.com/bitpay/bitcore-mnemonic/files/559361/brazilian_english.js.zip

Credits for the word list: https://github.com/eddieoz


NicolasDorier:
Nice!

johnnyasantoss:
The list created by @eddieoz isn't the one set by the BIP39 (see https://github.com/bitcoin/bips/pull/998). Actually, at the time that this PR was merged, there wasn't any pt-BR lists that were approved by the BIP39 proposal.

I just hope no one used this word list.

alissonsolitto:
@johnnyasantoss 

Very well pointed out, I stayed out of the crypt world until then.

johnnyasantoss:
Well, let's get this fixed. I'm just not sure how it would be the transition process from one word list to another. Maybe a program that exchanges the words based on their index, but that might not be the best ux.

NicolasDorier:
wow ok I will use the new version

NicolasDorier:
@johnnyasantoss I replace the old list with the new one. What I can do if somebody need it is to keep the old list and call it "legacy". Then end wallet can just check the current wordlist, and if checksum fail, try to fallback on the legacy one.

johnnyasantoss:
@NicolasDorier great! I'll be looking out for other bitcoin libs to check if they also have the non-standard wordlist.

PR #387, opened 2018-04-07
Feathercoin Support

ChekaZ:
Initial Support for Feathercoin.
- GetPoWHash() needs to be updated later on and switched to Neoscrypt.


ChekaZ:
Trying right now, but some problems occured, will notify when they are cleared.

NicolasDorier:
```
[DllImport("NBitcoin\\lib\\NeoScrypt.dll", EntryPoint = "neoscrypt_export", CallingConvention = CallingConvention.Cdecl)]
public static extern unsafe int neoscrypt(byte* input, byte* output, uint inputLength, uint profile);
```

Sorry, but don't use this. I don't want to depend on some native library. Either leave out the PoW implementation or do it is CSharp.

ChekaZ:
Sure will reverse it, and also stall this request as Feathercoin hasnt got a working RegTest right now.

So its closed at this state as I cant perform all tests.

ChekaZ:
Feathercoin hasnt got a working RegTest right now. So I cant perform all tests, will perform a new PR when its solved.

PR #392, opened 2018-04-11
Fixed Dogecoin AuxPow header

bdangh:
Fixed Dogecoin chain synchronization issue because of hard-fork at 371337rd block.

https://github.com/dogecoin/dogecoin/blob/bb4b082c086689434d67490a15224c42fefdfd13/src/primitives/block.h#L38
https://github.com/dogecoin/dogecoin/blob/bb4b082c086689434d67490a15224c42fefdfd13/src/auxpow.h

### DIFF
--- a/NBitcoin.Altcoins/Dogecoin.cs
+++ b/NBitcoin.Altcoins/Dogecoin.cs
@@ -27,13 +27,153 @@ namespace NBitcoin.Altcoins
 		}
 
 #pragma warning disable CS0618 // Type or member is obsolete
+		public class AuxPow : IBitcoinSerializable
+		{
+			Transaction tx = new Transaction();
+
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Cool!

NicolasDorier:
Push 1.0.0.9. Let me know if it works, this is first time we try an altcoin with different header (@Kukks you might be interested)

NicolasDorier:
Ping @dangershony you might like this

Kukks:
Awesome! Seems like the issue is handled fine with this so far

bdangh:
@NicolasDorier I tested it with mainnet node, all 2,175,500+ headers synchronized.

NicolasDorier:
Super cool!

PR #398, opened 2018-04-18
Add Network.NetworkType

bazooka70:
https://github.com/dgarage/NBXplorer/issues/45

Something like this?
All other alt coins should be updated as well...


bazooka70:
Do you approve with the enum naming or should it be the same as ChainType? i.e.
Main, Test, Regtest?

```
NetworkType = enum
{
  Regtest,
  Main,
  Test
}
```

the above seems kind of sloppy I must say.

NicolasDorier:
Same as ChainType is better, less chance we break stuff.

bazooka70:
I would probably change both to
```
{
  Main,
  Testnet
  Regtest,    
}
```

It's up to you

bazooka70:
I think its fine as is now

NicolasDorier:
Will remove the EnsureRegistered from here. It is important to make it explicit as it register some stuff globally.

bazooka70:
So now the `BCash.EnsureRegistered();` is no longer needed?
And used like ?
`var xnetwork = new NBXplorerNetworkProvider(NetworkType.Testnet).GetBCH();`

And in NBitcoin like this?
`Network _network1 = NBitcoin.Altcoins.BCash.Instance.Testnet;`
OR?
`Network _network2 = AltNetworkSets.BCash.Testnet;`

NicolasDorier:
yes

PR #399, opened 2018-04-20
Viacoin support 

romanornr:
Viacoin does have Merged scrypt mining. 
Unsure if I get it working but I will try. 

This might not be an issue for btcpay right? 
Since the explorer should trust the node.

### DIFF
--- a/NBitcoin.TestFramework/WellknownNodeDownloadData.cs
+++ b/NBitcoin.TestFramework/WellknownNodeDownloadData.cs
@@ -115,6 +115,35 @@ namespace NBitcoin.Tests
 			};
 		}
 
+		public class ViacoinNodeDownloadData
+		{
+			public NodeDownloadData v0_15_1 = new NodeDownloadData()
+			{
+				Version = "0.15.1",
+				Windows = new NodeOSDownloadData()
+				{
+					DownloadLink = "https://github.com/viacoin/viacoin/releases/download/v{0}.0/viacoin-{0}-win-setup-unsigned.exe",
+					Archive = "viacoin-{0}-win64.zip",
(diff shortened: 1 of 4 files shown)


NicolasDorier:
You can test if you did things correctly:

Add Viacoin here https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin.Tests/NodeBuilderEx.cs

Then you can run the tests with

```
dotnet test NBitcoin.Tests.csproj --filter "Altcoins=Altcoins" -p:ParallelizeTestCollections=false --framework netcoreapp2.0
```

inside NBitcoin.Tests folder.

NicolasDorier:
If PoW does not work it is not a big deal, but at least you will know if you did any other stupid error.

NBitcoin.TestFramework/WellknownNodeDownloadData.cs
```diff
@@ -115,6 +115,35 @@ public class LitecoinNodeDownloadData
 			};
 		}
 
+		public class ViacoinNodeDownloadData
```
NicolasDorier: you forgot to expose it as a property as well. Scroll down this file to see how.

NicolasDorier:
great! did you try to run the Altcoins tests so it catches low hanging mistakes you may have done?

NicolasDorier:
does not pass. One is that the regtest genesis block is wrong, other is NBitcoin which fails to connect (probably wrong network magic).

image

romanornr:
Alright I will try to fix this. 
Going to check out today how I can run these C# tests on my arch Linux machine.

NicolasDorier:
@romanornr no different from how to run any test in .NET Core. Just install .NET Core on your machine and run the command line I documented in NBitcoin.Tests.

PR #400, opened 2018-04-23
VerifyMessage method overloads

lontivero:
Provides method overloads for verifying raw (`byte[]`) messages with raw signatures in both, `BitcoinPubkeyAddress` and `BitcoinWitPubkeyAddress` as follow:

Currently (before this PR) we have one method:
```c#
public bool VerifyMessage(string message, string signature)
```

Proposed (after this PR) we have an overload:
```c#
public bool VerifyMessage(byte[] message, byte[] signature)
```

The PR also introduces a new interface to abstract both type of addresses given that they both have `VerifyMessage` method. @NicolasDorier please take a look at this and let me know your feedback.

Finally, the already-existing `CanVerifySignature` unit test was extended in order to include also a test for verifying signatures from bech32 addresses.

### DIFF
--- a/NBitcoin.Tests/key_tests.cs
+++ b/NBitcoin.Tests/key_tests.cs
@@ -69,21 +69,28 @@ namespace NBitcoin.Tests
 					Message = "this is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long messagethis is a very long message",
 					Signature = "HFKBHewleUsotk6fWG0OvWS/E2pP4o5hixdD6ui60in/x4376FBI4DvtJYrljXLNJTG1pBOZG+qRT/7S9WiIBfQ="
 				},
+				new
+				{
+					Address = "bc1q463gmsagg5u8wvqqcqj92yytt0pmevvg39h9jp",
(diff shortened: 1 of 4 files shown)


NicolasDorier:
no, there is @kallewoof who was working on something to make VerifyMessage more universal, but unsure how far he is. Will ask him.

nopara73:
To be clear, we don't need that feature, I'm just curious.

lontivero:
Just a comment about why it is not possible: Alice signs a message `m` with her private key `ka` such that `signature = ecdsa_sign(ka, m)`. ECDSA signature scheme allows us to recover the Alice's public key `Ka` from the signature and the message such that `Ka = ecdsa_recover_pubkey(signature, m)`. Now that we have the Alice's public key, we can check if the corresponding bitcoin address match with the Alice's address: `hash160(hash256(Ka)) == alice_address`.

The important point here is that we can recover the public key so, we can generate an address from it. We **cannot recover the script (neither its hash)** necessary for generating a p2sh address.

NicolasDorier:
It is possible if you change the definition where "Verifying the signature" means "Providing a scriptSig/Witness solving the scriptPubKey"

nopara73:
Quoting @SamouraiDev's explanation here: https://github.com/nopara73/ZeroLink/issues/45

> This can be done with p2sh wrapped segwit since the script format is known beforehand (as opposed to other p2sh). Pubkey is recovered from signature and used to derive p2sh-p2wpkh address.

lontivero:
Ohhh I see. Then, something like this should work:

```c#
public bool VerifyMessage(string message, string signature)
{
	var key = PubKey.RecoverFromMessage(message, signature);
	return key.WitHash.ScriptPubKey.Hash == Hash;
}
```

Let try.

nopara73:
Neat 😄

PR #401, opened 2018-04-23
Light client p2p protocol

lontivero:
This is for keeping some progress and discussions about the coming light client (bip 157 and bip 158). Note that Bitcoin Core has a first on-review PR for Golomb-Rice filters (https://github.com/bitcoin/bitcoin/pull/12254).

The idea is starting with the low-hanging fruit such as the more relevant p2p protocol message. Once it is compatible with btcd nodes, we could use it in WasabiWallet (far future)

### DIFF
--- a/NBitcoin/Protocol/Payloads/GetCompactFiltersPayload.cs
+++ b/NBitcoin/Protocol/Payloads/GetCompactFiltersPayload.cs
@@ -7,4 +7,55 @@ namespace NBitcoin.Protocol
 		Basic = (0x00),
 	}
 
+	/// <summary>
+	/// Represents the p2p message payload used for requesting a range of compact filter/headers.
+	/// </summary>
+	public abstract class CompactFiltersQueryPayload : Payload
+	{
+		private byte _FilterType = (byte)FilterType.Basic;
+		private uint _StartHeight = 0;
+		private uint256 _StopHash = new uint256();
+
+		/// <summary>
+		/// Gets the Filter type for which headers are requested.
+		/// </summary>
+		public FilterType FilterType
+		{
+			get => (FilterType)_FilterType;
+			internal set => _FilterType = (byte)value;
+		}
+
+		/// <summary>
+		/// Gets the height of the first block in the requested range.
+		/// </summary>
+		public uint StartHeight => _StartHeight;
+		
+		/// <summary>
+		/// Gets the hash of the last block in the requested range.
+		/// </summary>
+		public uint256 StopHash => _StopHash;
+		
+		protected CompactFiltersQueryPayload(FilterType filterType, uint startHeight, uint256 stopHash)
+		{
+			if (filterType != FilterType.Basic)
+				throw new ArgumentException($"'{filterType}' is not a valid value.", nameof(filterType));
+			if (stopHash == null)
+				throw new ArgumentNullException(nameof(stopHash));
+
+			FilterType = filterType;
+			_StartHeight = startHeight;
+			_StopHash = stopHash;
+		}
+
+		protected CompactFiltersQueryPayload() { }
+
+		#region IBitcoinSerializable Members
+		public override void ReadWriteCore(BitcoinStream stream)
+		{
+			stream.ReadWrite(ref _FilterType);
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Wow this is awesome! will read more on this.

lontivero:
Bitcoin Core team members are discussing about what info to include into the filters, how many filter types should exist and how to verify the filters are correct. Additionally depending on the amount and type of data, the P value (used to set the FP rate) could need to be recalculated.
https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2018-May/015972.html

knocte:
Just curious, did BIP157 and BIP158 implementations ever got included into BitcoinCore?

lontivero:
> Just curious, did BIP157 and BIP158 implementations ever got included into BitcoinCore?

* BIP158: **yes**
* BIP157: **no**

NBitcoin/Protocol/Payloads/GetCompactFiltersPayload.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NBitcoin.Protocol
+{
+	public enum FilterType : byte
+	{
+		Basic = (0x00),
+		Extended = (0x01)
```
dangershony: What does extended filter hold?

lontivero: https://github.com/bitcoin/bips/commit/b65ca1c094c1efb193a79d39efac6e34d390254a#diff-367f1ff45d7efb5c6774e4112cd17489

lontivero:
@NicolasDorier I would like to continue with this PR but I hit a wall that kept me blocked for months. I have implemented the p2p messages and I can see the network traffic is okay and that Bitcoin Core receives and replay okay so, that part is correct and works perfect.

The problem I have is I am not sure how to implement the compact filters and compact filter headers. Considering the bip 157 recommends to synchronize the block headers first, should implement this using something like the `CompactFiltersChainBehavior`? or should I modify the `ChainBehavior`? Do you think it is correct to reuse the `SlimChain` to keep the chain of compact filter headers?

I would appreciate if you could take a look at the bip 157 and provide me some guidance. 

----------

Btw, the commits that starts with `[WIP]` are just garbage but I pushed them to give you an idea of what I am talking about.

lontivero:
It seems the fact this PR is 401 has bad karma and will not be authorized. cuack! 

Idk when i will find the time to continue with this so, what do you think about removing the [WIP] commits? Would you merge it?

lontivero:
@NicolasDorier I hate to break your zen but as every year, since 2018, I feel the need to go back to this PR. 

**What do you think about removing the [WIP] commits? Would you merge it?** Thank you.

NicolasDorier:
well I completely forgot this PR, if you think it's ready to merge let's merg eit.

NBitcoin.Tests/ProtocolTests.cs
```diff
@@ -342,6 +342,80 @@ public void CanHandshakeWithSeveralTemplateBehaviors()
 				}
 			}
 		}
+
+		//[Fact]
+		//[Trait("Protocol", "Protocol")]
```
NicolasDorier, line 347: Remove this so the CI can test it.

lontivero:
The PR contains only the protocol messages (payload). I removed the `CompactFiltersChainBehavior` and the integration test because I don't know how to implement it correctly.

knocte:
@lontivero just curious, how far is this for a light-node to not depend on, for exampe, electrum servers?

lontivero:
It is still far from that because the bip recommends to download the block headers chain first and then download the compact filters chain.

You can take a look at this two commits in order to have an idea about how to make a naive implementation of this:
* https://github.com/lontivero/NBitcoin/commit/aee05b855212d225f435e6b9bfa24fdddad71355
* https://github.com/lontivero/NBitcoin/commit/a1f348517f90b6b6cbb03ff7df3babb5a8488e05

PR #403, opened 2018-04-23
Implements VerifyMessage for p2wsh addresses

lontivero:
@nopara73 was right and p2wsh addresses can be used to verify messages too.


NicolasDorier:
pushed new NBitcoin version with all your PRs.

PR #404, opened 2018-04-23
Remove obsolete rpc methods

lontivero:
This removes those RPC call that are no available anymore in the bitcoin core reference implementation. Most of these are about fee estimations, for example:

+ **estimatefee** was removed in v0.17 and it doesn't anymore work, it simply throws an exception:
```c++
UniValue estimatefee(const JSONRPCRequest& request)
{
    throw JSONRPCError(RPC_METHOD_DEPRECATED, "estimatefee was removed in v0.17.\n"
        "Clients should use estimatesmartfee.");
}
```

+ The `estimatepriority` and `estimatesmartpriority` RPCs have been removed in 0.15.

It is clear that those using older versions of bitcoin core node could be using these rpc calls. This PR could be applied in any moment. So, here i have a question: *How many versions do you plan to ensure back compatibility, @NicolasDorier?* IMHO I think that 0.15 could be a good candidate as the latest supported version for rpc.


NicolasDorier:
Knowing that people can easily make extensions method of those if they really need it, it is not a big deal if we remove it.

PR #406, opened 2018-04-25
Exposes EC classes to external packages

xJonathanLEI:


bazooka70:
While we here at this subject, would anyone mind if I make a PR to make `BitWriter.cs` `BitReader`/`BitWriter` public?

I need `BitWriter` to retrieve the entropy back from a Mnemonic, and don't really want to use a copy.

lontivero:
@bazooka70 `FastBitArray` is public and could be useful.

NicolasDorier:
I just learnt that `FastBitArray` exists lol.

xJonathanLEI:
I made this PR because the project says:

>It relies on the BouncyCastle cryptography library instead of OpenSSL, yet replicates OpenSSL bugs to guarantee compatibility

So it means BouncyCastle has been modified to be Bitcoin-compatible. If this is true, using the "true" BouncyCastle wouldn't help?

As for the removal of .net461, it was only for local use. I wasn't able to build the project as I didn't have .NET Framework installed. I reverted the removal at a later commit. Anyway I know I shouldn't have included these changes here. If the exposure of EC classes *is* accepted. I'd close this one and make another PR without the removal and revert.

Great library!

lontivero:
So, @xJonathanLEI  it seems you want to contribute with the project, right? I think there are lots of meaningful to do. You could ask to the mantainer (@NicolasDorier) what would be the best place to start or you could complete the RPC client class (there are many missing rpc method to implement). More tests are always welcome in any project and are the best way to learn about the hard details, etc. Welcome!

NicolasDorier:
@xJonathanLEI NBitcoin just copy/pasted BouncyCastle and stripped useless stuff. The reason is that the official package is not maintained very well on nuget and would prevent me from supporting the runtime I want without begging the maintainer.

If the BouncyCastle package ever because better maintained I would depend on it.

NicolasDorier:
that said, it is a pain in the ass that you can't build without having net461. Will check this.

NicolasDorier:
thing I am afraid is: If your project use `Portable.BouncyCastle` but I depends on `BouncyCastle.NetCore` then you'll get conflict all around the code by depending on NBitcoin. To prevent this, I just copied the code over.

knocte:
So then maybe the solution for NBitcoin is have a conditional dependency? Then you would provide NBitcoin.Portable as the nuget package that depends on BouncyCastle.Portable instead of BouncyCastle.NetCore.

Anyway, Portable profiles are going away. If someone is having issues about using some library because he's linking to BouncyCastle.Portable, I think it's fine these days to tell him to switch from BouncyCastle.Portable to BouncyCastle.NetCore, as .NET Standard is "the future".

NicolasDorier:
but BouncyCastle.Portable also support standard.

Issue #396, opened 2018-04-16
TransactionBuilder builds tx w/ duplicated output


lontivero:
## Summary

TransactionBuilder builds a transaction with duplicated (identical) outputs. 

## Description

After creating a transaction with only one coin (spending all of it) and continue building it, the resulting transaction contains two identical outputs. The code snippets below shows the problem:

```c#
var rpc = Global.RpcClient;

// send money with a replaceable transaction
var fundingTxid = await rpc.SendToAddressAsync(key.GetP2wpkhAddress(network), new Money(1m, MoneyUnit.BTC), replaceable: true);

// Get the funding transaction
var fundingTx = await rpc.GetRawTransactionAsync(fundingTxid);

// Get the tx with the coin used as input in the fundingTx, the one that I want to double spend.
var prevTx = await rpc.GetRawTransactionAsync(fundingTx.Inputs[0].PrevOut.Hash);

// The already-spent coin
var coinToDoubleSpend = new Coin(prevTx, prevTx.Outputs[fundingTx.Inputs[0].PrevOut.N]); 

// Simply getting the key that I must to use.
var signingKeys = keyManager.GetSecrets("password", new [] { coinToDoubleSpend.ScriptPubKey });


// The transaction is created properly 
var tb = new TransactionBuilder();
var doubleSpendingTx = tb
     .AddCoins(coinToDoubleSpend)
     .AddKeys(signingKeys.ToArray())
     .Send(key.GetP2wpkhAddress(network), coinToDoubleSpend.Amount- Money.Coins(0.001m))
     .SendFees(Money.Coins(0.001m))
     .BuildTransaction(false);

// HERE ----+
//          |
//          v                                  
// NOTE: at this point the doubleSpendingTx contains only one output. OK
Assert.Equal(1, doubleSpendingTx.Outputs.Count);  // Pass

// In the original failing test this line is not commented, however the problem
// occurs with and without this line. 
// doubleSpendingTx.Inputs[0].Sequence = new Sequence(0x0);

// Sign the transaction
doubleSpendingTx = tb.ContinueToBuild(doubleSpendingTx).BuildTransaction(true);

// HERE ----+
//          |
//          v                                  
// ERROR: at this point the doubleSpendingTx contains two identical outpust.
// The same output is twice. NOT OK
Assert.Equal(1, doubleSpendingTx.Outputs.Count);  // Fails
```

Please let me know if more info is needed.

lontivero:
I will take a look at it too

[...]

PR #407, opened 2018-04-25
Improvements proposal

lontivero:
This PR contains a proposal for preventing the missuses reported in https://github.com/MetacoSA/NBitcoin/issues/396. I tried a couple of ideas but finally I went for the easiest (a flag).

### DIFF
--- a/NBitcoin.Tests/transaction_tests.cs
+++ b/NBitcoin.Tests/transaction_tests.cs
@@ -2892,6 +2892,29 @@ namespace NBitcoin.Tests
 		}
 
 		[Fact]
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Can you remove the changes on `TxOut` ?

It should be in another PR. Also I am still not sure it is a good idea.
The code is duplicated.

Later on, I want to support altcoins with different `TxOut` format so it makes implementers them a little more work which can go wrong.

lontivero:
I've removed the commits that changes the `TxOut` class and the test written to show the scenario to fix.

PR #413, opened 2018-04-29
Stratis Altcoin Support

mikedennis:
This is not ready to push yet but wanted to get @NicolasDorier and @dangershony feedback as I had to make some NBitcoin access level changes to get transaction serialization to work.

Also Stratis needs some RPC work before it can support the altcoin testing framework. I've logged these issues in the Stratis github which are needed to support the altcoin testing framework as is:

https://github.com/stratisproject/StratisBitcoinFullNode/issues/1355
https://github.com/stratisproject/StratisBitcoinFullNode/issues/1356

This still needs some more testing for sure but interested in any feedback you have with what I have currently.

### DIFF
--- a/NBitcoin.Altcoins/Stratis.cs
+++ b/NBitcoin.Altcoins/Stratis.cs
@@ -1,0 +1,660 @@
+﻿using NBitcoin.Altcoins.HashX11;
+using NBitcoin.Crypto;
+using NBitcoin.DataEncoders;
+using NBitcoin.Protocol;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+
+namespace NBitcoin.Altcoins
+{
+	public class Stratis : NetworkSetBase
+	{
+		public static Stratis Instance { get; } = new Stratis();
+
+		public override string CryptoCode => "STRAT";
+
+		private Stratis()
+		{
+
+		}
+
+#pragma warning disable CS0618 // Type or member is obsolete
+		public class StratisConsensusFactory : ConsensusFactory
+		{
+			private StratisConsensusFactory()
+			{
+			}
+
+			public static StratisConsensusFactory Instance { get; } = new StratisConsensusFactory();
+
+			public override BlockHeader CreateBlockHeader()
+			{
+				return new StratisBlockHeader();
+			}
+
+			public override Block CreateBlock()
+			{
+				return new StratisBlock(new StratisBlockHeader());
+			}
+
+			public override Transaction CreateTransaction()
+			{
+				return new StratisTransaction(this);
+			}
+
+			protected bool IsHeadersPayload(Type type)
+			{
+				var baseType = typeof(HeadersPayload).GetTypeInfo();
+				return baseType.IsAssignableFrom(type.GetTypeInfo());
+			}
+
+			public override bool TryCreateNew(Type type, out IBitcoinSerializable result)
+			{
+				result = null;
+				if (IsHeadersPayload(type))
+				{
+					result = CreateHeadersPayload();
+					return true;
+				}
+
+				return base.TryCreateNew(type, out result);
+			}
+
+			public HeadersPayload CreateHeadersPayload()
+			{
+				return new StratisHeadersPayload();
+			}
+		}
+
+		public class StratisHeadersPayload : HeadersPayload
+		{
+			class BlockHeaderWithTxCount : IBitcoinSerializable
+			{
+				public BlockHeaderWithTxCount()
+				{
+
+				}
+				public BlockHeaderWithTxCount(BlockHeader header)
+				{
+					_Header = header;
+				}
+				internal BlockHeader _Header;
+				#region IBitcoinSerializable Members
+
+				public void ReadWrite(BitcoinStream stream)
+				{
+					stream.ReadWrite(ref _Header);
+					VarInt txCount = new VarInt(0);
+					stream.ReadWrite(ref txCount);
+
+					// Stratis specific addition - unknown usage see Stratis source.
+					stream.ReadWrite(ref txCount);
+				}
+
+				#endregion
+			}
+
+			public override void ReadWriteCore(BitcoinStream stream)
+			{
+				if (stream.Serializing)
+				{
+					var heardersOff = Headers.Select(h => new BlockHeaderWithTxCount(h)).ToList();
+					stream.ReadWrite(ref heardersOff);
+				}
+				else
+				{
+					Headers.Clear();
+					List<BlockHeaderWithTxCount> headersOff = new List<BlockHeaderWithTxCount>();
+					stream.ReadWrite(ref headersOff);
+					Headers.AddRange(headersOff.Select(h => h._Header));
+				}
+			}
+		}
+
+		public class StratisBlockSignature : IBitcoinSerializable
+		{
+			protected bool Equals(StratisBlockSignature other)
+			{
+				return Equals(signature, other.signature);
+			}
+
+			public override bool Equals(object obj)
+			{
+				if (ReferenceEquals(null, obj)) return false;
+				if (ReferenceEquals(this, obj)) return true;
+				if (obj.GetType() != this.GetType()) return false;
+				return Equals((StratisBlockSignature)obj);
+			}
+
+			public override int GetHashCode()
+			{
+				return (signature?.GetHashCode() ?? 0);
+			}
+
+			public StratisBlockSignature()
+			{
+				this.signature = new byte[0];
+			}
+
+			private byte[] signature;
+
+			public byte[] Signature
+			{
+				get
+				{
+					return signature;
+				}
+				set
+				{
+					signature = value;
+				}
+			}
+
+			internal void SetNull()
+			{
+				signature = new byte[0];
+			}
+
+			public bool IsEmpty()
+			{
+				return !this.signature.Any();
+			}
+
+			public static bool operator ==(StratisBlockSignature a, StratisBlockSignature b)
+			{
+				if (System.Object.ReferenceEquals(a, b))
+					return true;
+
+				if (((object)a == null) || ((object)b == null))
+					return false;
+
+				return a.signature.SequenceEqual(b.signature);
+			}
+
+			public static bool operator !=(StratisBlockSignature a, StratisBlockSignature b)
+			{
+				return !(a == b);
+			}
+
+			#region IBitcoinSerializable Members
+
+			public void ReadWrite(BitcoinStream stream)
+			{
+				stream.ReadWriteAsVarString(ref signature);
+			}
+
+			#endregion
+
+			public override string ToString()
+			{
+				return Encoders.Hex.EncodeData(this.signature);
+			}
+		}
+
+		/// <summary>
+		/// A POS block header, this will create a work hash based on the X13 hash algos.
+		/// </summary>
+		public class StratisBlockHeader : BlockHeader
+		{
+			const int CurrentVersion = 7;
+
+			private readonly X13 x13;
+
+			public StratisBlockHeader() : base()
+			{
+				this.x13 = new X13();
+			}
+
+			protected internal override void SetNull()
+			{
+				nVersion = CurrentVersion;
+				hashPrevBlock = 0;
+				hashMerkleRoot = 0;
+				nTime = 0;
+				nBits = 0;
+				nNonce = 0;
+			}
+
+			private byte[] CalculateHash(byte[] data, int offset, int count)
+			{
+				byte[] hashData = data.Skip(offset).Take(count).ToArray();
+				uint256 hash = null;
+				if (this.nVersion > 6)
+					hash = Hashes.Hash256(hashData);
+				else
+				{
+					var x13 = new X13();
+					hash = new uint256(this.x13.ComputeBytes(hashData));
+				}
+				return hash.ToBytes();
+			}
+
+			protected override HashStreamBase CreateHashStream()
+			{
+				return BufferedHashStream.CreateFrom(CalculateHash);
+			}
+
+			public override uint256 GetPoWHash()
+			{
+				
+				return new uint256(this.x13.ComputeBytes(this.ToBytes()));
+			}
+		}
+
+		/// <summary>
+		/// A POS block that contains the additional block signature serialization.
+		/// </summary>
+		public class StratisBlock : Block
+		{
+			/// <summary>
+			/// A block signature - signed by one of the coin base txout[N]'s owner.
+			/// </summary>
+			private StratisBlockSignature blockSignature = new StratisBlockSignature();
+
+			public StratisBlock(StratisBlockHeader blockHeader) : base(blockHeader)
+			{
+			}
+
+			public override ConsensusFactory GetConsensusFactory()
+			{
+				return StratisConsensusFactory.Instance;
+			}
+
+			/// <summary>
+			/// The block signature type.
+			/// </summary>
+			public StratisBlockSignature BlockSignature
+			{
+				get { return this.blockSignature; }
+				set { this.blockSignature = value; }
+			}
+
+			/// <summary>
+			/// The additional serialization of the block POS block.
+			/// </summary>
+			public override void ReadWrite(BitcoinStream stream)
+			{
+				base.ReadWrite(stream);
+				stream.ReadWrite(ref this.blockSignature);
+
+				//this.BlockSize = stream.Serializing ? stream.Counter.WrittenBytes : stream.Counter.ReadBytes;
+			}
+		}
+		
+		private class StratisWitness
+		{
+			private TxInList _Inputs;
+
+			public StratisWitness(TxInList inputs)
+			{
+				_Inputs = inputs;
+			}
+
+			internal bool IsNull()
+			{
+				return _Inputs.All(i => i.WitScript.PushCount == 0);
+			}
+
+			internal void ReadWrite(BitcoinStream stream)
+			{
+				for (int i = 0; i < _Inputs.Count; i++)
+				{
+					if (stream.Serializing)
+					{
+						var bytes = (_Inputs[i].WitScript ?? WitScript.Empty).ToBytes();
+						stream.ReadWrite(ref bytes);
+					}
+					else
+					{
+						_Inputs[i].WitScript = WitScript.Load(stream);
+					}
+				}
+
+				if (IsNull())
+					throw new FormatException("Superfluous witness record");
+			}
+		}
+
+		public class StratisTransaction : Transaction
+		{
+			public StratisTransaction(ConsensusFactory consensusFactory)
+			{
+				_Factory = consensusFactory;
+			}
+
+			ConsensusFactory _Factory;
+			public override ConsensusFactory GetConsensusFactory()
+			{
+				return _Factory;
+			}
+
+			/// <summary>
+			/// POS Timestamp
+			/// </summary>
+			public uint Time { get; set; } = Utils.DateTimeToUnixTime(DateTime.UtcNow);
+
+			public override void ReadWrite(BitcoinStream stream)
+			{		
+				var witSupported = (((uint)stream.TransactionOptions & (uint)TransactionOptions.Witness) != 0) &&
+										stream.ProtocolCapabilities.SupportWitness;
+
+				if (stream.Serializing)
+					SerializeTxn(stream, witSupported);
+				else
+					DeserializeTxn(stream, witSupported);				
+			}
+
+			private void DeserializeTxn(BitcoinStream stream, bool witSupported)
+			{
+				byte flags = 0;
+
+				UInt32 nVersionTemp = 0;
+				stream.ReadWrite(ref nVersionTemp);
+
+				// POS time stamp
+				uint nTimeTemp = 0;
+				stream.ReadWrite(ref nTimeTemp);
+
+				TxInList vinTemp = new TxInList();
+				TxOutList voutTemp = new TxOutList();
+
+				/* Try to read the vin. In case the dummy is there, this will be read as an empty vector. */
+				stream.ReadWrite<TxInList, TxIn>(ref vinTemp);
+
+				var hasNoDummy = (nVersionTemp & NoDummyInput) != 0 && vinTemp.Count == 0;
+				if (witSupported && hasNoDummy)
+					nVersionTemp = nVersionTemp & ~NoDummyInput;
+
+				if (vinTemp.Count == 0 && witSupported && !hasNoDummy)
+				{
+					/* We read a dummy or an empty vin. */
+					stream.ReadWrite(ref flags);
+					if (flags != 0)
+					{
+						/* Assume we read a dummy and a flag. */
+						stream.ReadWrite<TxInList, TxIn>(ref vinTemp);
+						vinTemp.Transaction = this;
+						stream.ReadWrite<TxOutList, TxOut>(ref voutTemp);
+						voutTemp.Transaction = this;
+					}
+					else
+					{
+						/* Assume read a transaction without output. */
+						voutTemp = new TxOutList();
+						voutTemp.Transaction = this;
+					}
+				}
+				else
+				{
+					/* We read a non-empty vin. Assume a normal vout follows. */
+					stream.ReadWrite<TxOutList, TxOut>(ref voutTemp);
+					voutTemp.Transaction = this;
+				}
+				if (((flags & 1) != 0) && witSupported)
+				{
+					/* The witness flag is present, and we support witnesses. */
+					flags ^= 1;
+					StratisWitness wit = new StratisWitness(vinTemp);
+					wit.ReadWrite(stream);
+				}
+				if (flags != 0)
+				{
+					/* Unknown flag in the serialization */
+					throw new FormatException("Unknown transaction optional data");
+				}
+				LockTime lockTimeTemp = LockTime.Zero;
+				stream.ReadWriteStruct(ref lockTimeTemp);
+
+				this.Version = nVersionTemp;
+				this.Time = nTimeTemp; // POS Timestamp
+				vinTemp.ForEach(i => this.AddInput(i));
+				voutTemp.ForEach(i => this.AddOutput(i));
+				this.LockTime = lockTimeTemp;				
+			}
+
+			private void SerializeTxn(BitcoinStream stream, bool witSupported)
+			{
+				byte flags = 0;
+				var version = (witSupported && (this.Inputs.Count == 0 && this.Outputs.Count > 0)) ? this.Version | NoDummyInput : this.Version;
+				stream.ReadWrite(ref version);
+
+				// POS Timestamp
+				var time = this.Time;
+				stream.ReadWrite(ref time);
+
+				if (witSupported)
+				{
+					/* Check whether witnesses need to be serialized. */
+					if (HasWitness)
+					{
+						flags |= 1;
+					}
+				}
+				if (flags != 0)
+				{
+					/* Use extended format in case witnesses are to be serialized. */
+					TxInList vinDummy = new TxInList();
+					stream.ReadWrite<TxInList, TxIn>(ref vinDummy);
+					stream.ReadWrite(ref flags);
+				}
+				TxInList vin = this.Inputs;				
+				stream.ReadWrite<TxInList, TxIn>(ref vin);
+				vin.Transaction = this;
+				TxOutList vout = this.Outputs;
+				stream.ReadWrite<TxOutList, TxOut>(ref vout);
+				vout.Transaction = this;
+				if ((flags & 1) != 0)
+				{
+					StratisWitness wit = new StratisWitness(this.Inputs);
+					wit.ReadWrite(stream);
+				}
+				LockTime lockTime = this.LockTime;
+				stream.ReadWriteStruct(ref lockTime);
+			}
+
+			public static StratisTransaction ParseJson(string tx)
+			{
+				JObject obj = JObject.Parse(tx);
+				StratisTransaction stratTx = new StratisTransaction(Stratis.StratisConsensusFactory.Instance);
+				DeserializeFromJson(obj, ref stratTx);
+
+				return stratTx;
+			}
+
+			private static void DeserializeFromJson(JObject json, ref StratisTransaction tx)
+			{
+				tx.Version = (uint)json.GetValue("version");
+				tx.Time = (uint)json.GetValue("time");
+				tx.LockTime = (uint)json.GetValue("locktime");
+
+				var vin = (JArray)json.GetValue("vin");
+				for (int i = 0; i < vin.Count; i++)
+				{
+					var jsonIn = (JObject)vin[i];
+					var txin = new TxIn();
+					tx.Inputs.Add(txin);
+
+					var script = (JObject)jsonIn.GetValue("scriptSig");
+					if (script != null)
+					{
+						txin.ScriptSig = new Script(Encoders.Hex.DecodeData((string)script.GetValue("hex")));
+						txin.PrevOut.Hash = uint256.Parse((string)jsonIn.GetValue("txid"));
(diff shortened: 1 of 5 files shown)


NicolasDorier:
Run the tests on a stratis node by following example in `NodeBuilderEx.cs`, and run the tests with traits `Altcoins=Altcoins`.

mikedennis:
Yes I have tried that and Stratis daemon needs a few more RPC methods added in order to pass the tests.

NicolasDorier:
ah? I am surprised I thought it got everything already.

NicolasDorier:
How is it going?

mikedennis:
Haven't had a chance to work on this much. Will try to find some time this week to write some tests and see if it works.

mikedennis:
First test successful - I've been able to create a transaction on stratis testnet using this code. The resulting transaction is here: https://chainz.cryptoid.info/strat-test/tx.dws?838640.htm

NicolasDorier:
Did you run the altcoins tests in NBitcoin? This will find issues quite easily.

NicolasDorier:
You certainly have a `generate` somewhere right? else you can't do proper tests.

NicolasDorier:
@D3m0nKingx how are you testing anything without a `generate`? I remember the stratisfullnode for stratis had a call to generate

D3m0nKingx:
@NicolasDorier that's why the Fullnode is being built cause stratisX was just an interim wallet for the time being. And I'm not doing any testing in this scope. Just informing :P

NicolasDorier:
So what is missing on the fullnode ?

D3m0nKingx:
This is what I can see atm: 
https://raw.githubusercontent.com/stratisproject/StratisBitcoinFullNode/master/src/NBitcoin/RPC/RPCClient.cs

and

https://raw.githubusercontent.com/stratisproject/StratisBitcoinFullNode/master/src/NBitcoin/RPC/RPCClient.Wallet.cs

But somethings like gettxout are just the calls to the methods. so still things to get implemented. I'm not part of the dev team, I think @dangershony will know more about it.

NicolasDorier:
I don't think I am using `gettxout` at all in the Altcoin tests.

mikedennis:
These are the RPC methods that altcoin test framework is using but are missing from C# full node https://github.com/stratisproject/StratisBitcoinFullNode/issues/1356  You are correct there is indeed a generate method that we use in our tests but it hasn't been RPC'd yet. I don't think it's worth spending any more time on the C++ node (StratisX), just thought I'd give it a try and see if it worked out of the box which unfortunately it didn't. Instead will focus on C# fullnode moving forward.

NBitcoin/Transaction.cs
```diff
@@ -1083,7 +1083,8 @@ public enum TransactionOptions : uint
 		Witness = 0x40000000,
 		All = Witness
 	}
-	class Witness
+
+	public class Witness
```
NicolasDorier: Don't use public here, at worst copy the Witness implementation in Stratis code.

NicolasDorier: Or maybe use InternalsVisibleTo("NBitcoin.Altcoins")

mikedennis: 👍

NicolasDorier:
I am quite excited about this. You are the first non bitcoind fork supporting it!

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,517 @@
...
+				return StratisConsensusFactory.Instance;
+			}
+
+			public override void ReadWrite(BitcoinStream stream)
+			{
+				base.ReadWrite(stream);
+				stream.ReadWrite(ref this.blockSignature);
+			}
+
+			public static bool BlockSignature = false;
```
dangershony: Why do you need this boolean? `BlockSignature`

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,517 @@
...
+			private StratisBlockSignature blockSignature = new StratisBlockSignature();
+
+			public StratisBlockSignature BlockSignatur
+			{
+				get { return this.blockSignature; }
+				set { this.blockSignature = value; }
+			}
+		}
+
+		private class StratisWitness
```
dangershony: We currently don't support witness so I suggest not adding witness at all, you just need to serialize the trx as is and att the `stream.ReadWrite(ref nTimeTemp);`

dangershony: hmm well if you got it working already and its working fine then leave it as is.

NBitcoin.TestFramework/WellknownNodeDownloadData.cs
```diff
@@ -318,6 +314,39 @@ public class PolisNodeDownloadData
 			};
 		}
 
+		/// <summary>
+		/// Currently this is using the c++ stratisx full node. Only the linux version is available at this time.
+		/// Should eventually be changed to use C# full node once RPC interfaces are more mature.
```
dangershony: Yeah you probably want to use the fullnode, also because we have more RPC coverage on the C# node.

dangershony:
>I am quite excited about this. You are the first non bitcoind fork supported in NBitcoin.  

Yes me too indeed, also because we will be able to use many of the NBitcoin dependent projects much easier.

NBitcoin.Altcoins/Stratis.cs
mikedennis: I think this was in the original Stratis code. Looks like it has been cleaned up since I wrote this so will take another look but my initial impression is that I should be able to remove it.

mikedennis:
Getting closer! Below is the current status of tests with the C# fullnode.
image

`CanCalculateTransactionHash()` and `CorrectCoinMaturity()` need a couple of more missing RPC methods added to StratisFullNode.

I also found that I didn't have Stratis Block Headers implemented properly when I ran the tests. So added the code for that. In order to do that there were a couple of changes for access level in Block.cs.

I also had to add some code specific to Stratis in the Altcoins tests due to difference from bitcoin in the way the Stratis wallet works. Stratis wallets are always encrypted and not created by default. There might be a better way of doing this.

I still need to investigate the `CanSyncWithPoW()` and `CanSyncWithoutPow()` tests as they never seem to run to completion (or maybe I'm not patient enough!)

NicolasDorier:
You don't have to care about `CanSyncWithPoW` but `CanSyncWithoutPow` must work and finish in less than 5 sec normally

dangershony:
All RPC methods needed to complete this should now be merged in to Stratis Full Node

NicolasDorier:
The altcoins tests pass for Stratis?

mikedennis:
Getting close to having them pass. Having issue with ```CorrectCoinMaturity()``` which we are seeing different behaviour in Stratis than btc core. I think it might be a bug in Stratis wallet code, it's being discussed in this Stratis PR https://github.com/stratisproject/StratisBitcoinFullNode/pull/2487 

Also having issue with getting ```CanSync...``` tests passing. I think I may still have an error with the Stratis block header definition in AltCoins. I've stepped in and I can see it fetching the headers but the checks never pass so it just keeps fetching headers over and over again. I need to debug further to narrow down the issue.

NicolasDorier:
need also rebase

dangershony:
@mikedennis this might be your issue
https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin/P2P/Protocol/Payloads/HeadersPayload.cs#L36

mikedennis:
Thanks @dangershony that might have taken me a long time to find! Yes I'm not accounting for that right now, will give that a try and see if I make more progress.

dangershony:
This requires a problematic changes to the low level NBitcoin payload messages, so perhaps involving @NicolasDorier is a good idea.

NicolasDorier:
That's fine, you can override the way you readwrite any structure with the `ConsensusFactory`.
In such case, create your own payload type deriving from NBitcoin's one and override the ReadWrite method.

mikedennis:
Below is current state of tests.

Still having problems syncing headers.  I've tried to add StratisHeaderPayload to the Stratis Consensus Factory. I am now getting the node disconnected after "GetHeaders" message is sent. Maybe I've done something wrong with the HeaderPayload override in consensus factory?

Also there is a new test which look like is going to require more RPC work in Stratis.

image

fellowserf:
Sent from Yahoo7 Mail on Android U ti g ku
 
  On Fri, 19 Oct 2018 at 3:11 AM, mikedennis<notificationsjugithub.com> wrote:   
Thanks @dujuscangershony that might Thxh6have taken me a long time to find! Yes I'm not accounting for that right now, will give that a try and see if I make more progress.up

NicolasDorier:
Can you take a look at DoesRPCCapabilities well advertised? I am using this to scan what RPC can do.
My hint is that it is because I am using RPC batching, though I remember having implemented it on stratis full node.

mikedennis:
Yes I don't think Stratis RPC server currently supports RPC batching. It's throwing an "Internal Error" exception when it receives a batch.

mikedennis:
For the chain syncing test - looks like nbitcoin doesn't like the header payload that Stratis sends it. Here is a snip out of the stratis log:

```
[2018-10-28 21:33:10.1791 8] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<ReceiveMessagesAsync>d__29.MoveNext [1-[::ffff:127.0.0.1]:54422] Received message: 'getheaders: GetHeadersPayload'
[2018-10-28 21:33:10.1791 8] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadAndParseMessageAsync (protocolVersion:SENDHEADERS_VERSION,)
[2018-10-28 21:33:10.1791 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeer.ProcessMessageAsync ()
[2018-10-28 21:33:10.1791 8] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadMessageAsync (protocolVersion:SENDHEADERS_VERSION,)
[2018-10-28 21:33:10.1791 8] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadMagicAsync (magic.Count:4,)
[2018-10-28 21:33:10.1791 8] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadBytesAsync (buffer.Count:1,offset:0,bytesToRead:1,)
[2018-10-28 21:33:10.1946 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeer.SendMessageAsync (payload:HeadersPayload,)
[2018-10-28 21:33:10.1946 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.SendAsync (payload:HeadersPayload,)
[2018-10-28 21:33:10.1946 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<SendAsync>d__31.MoveNext [1-[::ffff:127.0.0.1]:54422] Sending message: 'headers: HeadersPayload'
[2018-10-28 21:33:10.1946 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.SendAsync (data.Count:8225,)
[2018-10-28 21:33:10.1946 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<SendAsync>d__32.MoveNext (-)
[2018-10-28 21:33:10.1946 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<SendAsync>d__31.MoveNext (-)
[2018-10-28 21:33:10.1946 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeer+<SendMessageAsync>d__99.MoveNext (-)
[2018-10-28 21:33:10.1946 19] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeer+<ProcessMessageAsync>d__95.MoveNext (-)
[2018-10-28 21:33:10.8414 12] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerCollection.GetEnumerator ()
[2018-10-28 21:33:10.8414 12] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerCollection.GetEnumerator (-)
[2018-10-28 21:33:10.8414 12] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeer.Behavior ()
[2018-10-28 21:33:10.8414 12] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeer.Behavior (-)
[2018-10-28 21:33:13.5758 22] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<ReadBytesAsync>d__35.MoveNext [1-[::ffff:127.0.0.1]:54422] (-)[STREAM_END]
[2018-10-28 21:33:13.6235 22] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<ReadBytesAsync>d__35.MoveNext (-)$exception:System.OperationCanceledException: The operation was canceled.
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadBytesAsync(Byte[] buffer, Int32 offset, Int32 bytesToRead, CancellationToken cancellation)
[2018-10-28 21:33:13.6235 22] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<ReadMagicAsync>d__34.MoveNext (-)$exception:System.OperationCanceledException: The operation was canceled.
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadBytesAsync(Byte[] buffer, Int32 offset, Int32 bytesToRead, CancellationToken cancellation)
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadMagicAsync(Byte[] magic, CancellationToken cancellation)
[2018-10-28 21:33:13.6235 22] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<ReadMessageAsync>d__33.MoveNext (-)$exception:System.OperationCanceledException: The operation was canceled.
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadBytesAsync(Byte[] buffer, Int32 offset, Int32 bytesToRead, CancellationToken cancellation)
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadMagicAsync(Byte[] magic, CancellationToken cancellation)
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadMessageAsync(ProtocolVersion protocolVersion, CancellationToken cancellation)
[2018-10-28 21:33:13.6235 22] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<ReadAndParseMessageAsync>d__36.MoveNext (-)$exception:System.OperationCanceledException: The operation was canceled.
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadBytesAsync(Byte[] buffer, Int32 offset, Int32 bytesToRead, CancellationToken cancellation)
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadMagicAsync(Byte[] magic, CancellationToken cancellation)
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadMessageAsync(ProtocolVersion protocolVersion, CancellationToken cancellation)
   at Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection.ReadAndParseMessageAsync(ProtocolVersion protocolVersion, CancellationToken cancellation)
[2018-10-28 21:33:13.6235 22] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeerConnection+<ReceiveMessagesAsync>d__29.MoveNext [1-[::ffff:127.0.0.1]:54422] Receiving cancelled.
[2018-10-28 21:33:13.6235 22] TRACE: Stratis.Bitcoin.P2P.Peer.NetworkPeer.Disconnect (reason:'Receiving cancelled.',exception:<NULL>)
```

dangershony:
@mikedennis this may not be the reason, it could be the socket dropped for what ever reason.
In other tests headers are parsed successfully right?

mikedennis:
@dangershony On the nbitcoin side it generates a cancellation token which drops the connection. Need to figure out why. I suspect it's something to do with receiving the stratis headers payload.

mikedennis:
I wrote a version of  ```ScanRPCCapabilities()``` that doesn't use batch and reran ```DoesRPCCapabilitiesWellAdvertised()``` test.

Fails on this block of code. Stratis does not yet support Bech32 addresses
```
var address = rpc.GetNewAddress(new GetNewAddressRequest()
{
	AddressType = AddressType.Bech32
});
// If this fail, rpc support segwit bug you said it does not
Assert.Equal(rpc.Capabilities.SupportSegwit, address.ScriptPubKey.IsWitness);
```
Below are the RPC capabilities:
image

mikedennis:
For the disconnection issue for ```CanSyncWithoutPOW()``` This is the exception that is being thrown in NBitcoin:

```NBitcoin.Altcoins.Stratis+StratisHeadersPayload is not a payload```

I think this is because the enumeration of valid payloads is done on the NBitcoin assembly but doesn't look into NBitcoin.Altcoins. I believe this error is thrown by ```PayloadAttribute.cs```
```
		internal static string GetCommandName(Type type)
		{
			string result;
			if(!_TypeToName.TryGetValue(type, out result))
				throw new ArgumentException(type.FullName + " is not a payload");
			return result;
		}
```

File: NBitcoin.Tests/AltcoinTests.cs [hunk 5]
```diff
@@ -241,6 +251,7 @@ public void CorrectCoinMaturity()
 			using (var builder = NodeBuilderEx.Create())
 			{
 				var node = builder.CreateNode();
+				this.EnsureWallet(node);
```
NicolasDorier: Don't do this, you are putting burden on everybody making a test, and not only on this project. 

`NodeDownloadData.AdditionalRegtestConfig` can be used if you need to pass more config stuff.

mikedennis: Unfortunately configuration to do so does not yet exist in Stratis. I created an issue in Stratis github as linked below.

dangershony:
Is ready to be pushed? I see appveoyer tests are passing

NicolasDorier:
I don't know, you need to try the altcoins tests with Stratis full node.

NicolasDorier:
1. You can specify some custom configuration file in the `NodeDownloadData`.
2. Batching is indeed used, and must be supported. BTCPay (actually NBXplorer) can't work without it.
3. There is a way to have your own BlockHeader type recognized, look what I did for Elements/Liquid.

mikedennis:
@NicolasDorier RPC Batching support and creating wallet through config require some changes to Stratis Node - I'm working on those on Stratis side. Hopefully will have something soon.

I was able to get StratisHeaderPayload to work and successfully pass CanSyncWithoutPoW()  by making a small change to PayloadAttribute in NBitcoin. See commit https://github.com/MetacoSA/NBitcoin/pull/413/commits/04c8f7c2c946f0ea457b5ed0e9acf6f7b3244fef

NicolasDorier:
@mikedennis ok. Maybe instead we could use the consensus factory to create the P2P messages based on the name. Though it might involve more change.

NBitcoin/UnsignedList.cs
```diff
@@ -20,7 +20,7 @@ public UnsignedList(Transaction parent)
 		public Transaction Transaction
 		{
 			get;
-			internal set;
+			set;
 		}
```
NicolasDorier: why?

NBitcoin/Block.cs
```diff
@@ -236,7 +236,7 @@ public virtual uint256 GetPoWHash()
 			return GetHash();
 		}
 
-		public uint256 GetHash()
+		public virtual uint256 GetHash()
```
NicolasDorier: why? you can override `CreateHashStream`

NBitcoin.Altcoins/HashX11/X13.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
```
NicolasDorier: I thought we already had X13?

NBitcoin/UnsignedList.cs
mikedennis: Yes, I tried reverting this back to internal set and my code still builds. Must have been for an iteration that I didn't end up needing. I will revert this back to internal.

NBitcoin/Block.cs
mikedennis: I believe the Stratis code was overriding GetHash(), I'll look and see if I can do it via GetHashStream instead.

NBitcoin.Altcoins/HashX11/X13.cs
mikedennis: I didn't see it anywhere, unless I missed it somewhere else in NBitcoin?

mikedennis:
RE: Create P2P message from consensus factory

Do you mean add the new type to the PayloadAttribute list from within the ConsensusFactory? Could do that if you prefer but would involve adding a public method to PayloadAttribute for adding a new command for type that I could call in the StratisConsensusFactory instance initialization. Let me know if you prefer this and I can change it.

ChekaZ:
I got the code working with another X13 + nTime (PoS Time) Coin. But I actually had to change the X13 implementation or it wouldnt work. Could we consider to change this one and just extend the current X11 implementation?

NicolasDorier:
@ChekaZ make a separate PR for adding X13, then stratis will rebase.

mikedennis:
All altcoin tests now pass.

image

Should I squash commits?

mikedennis:
> I am surprised that the `CanSyncWithPow` is passing though... you did not forgot to select Stratis as the target of the test right? :p

It passed as soon as I got `CanSyncWithoutPoW` working. Haven't looked into why exactly it works but for sure I was running the right node. Will brush up on my git skills to see if I can rebase this pile of commits.

I also haven't ran the linux or osx versions yet, only windows. I will set something up to make sure they are good as well.

mikedennis:
Ok the commit history is now much cleaner. I think it's ready for a final review.

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+			}
+
+			internal void SetNull()
+			{
+				signature = new byte[0];
+			}
+
+			public bool IsEmpty()
+			{
+				return !this.signature.Any();
```
lontivero: Given signature can be assigned and `null` is a valid values then this method can throw NRE.

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+
+			public static bool operator ==(StratisBlockSignature a, StratisBlockSignature b)
+			{
+				if (System.Object.ReferenceEquals(a, b))
+					return true;
+
+				if (((object)a == null) || ((object)b == null))
+					return false;
+
+				return a.signature.SequenceEqual(b.signature);
```
lontivero: This is okay but in many places we use `Utils.ArrayEqual`.

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+			public void ReadWrite(BitcoinStream stream)
+			{
+				stream.ReadWriteAsVarString(ref signature);
+			}
+
+			#endregion
+
+			public override string ToString()
+			{
+				return Encoders.Hex.EncodeData(this.signature);
```
lontivero: This can throw an `ArgumentNullException` when `this.signature` is `null`. What should `ToString` return in case `signature` is empty?

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+				hashPrevBlock = 0;
+				hashMerkleRoot = 0;
+				nTime = 0;
+				nBits = 0;
+				nNonce = 0;
+			}
+
+			private byte[] CalculateHash(byte[] data, int offset, int count)
+			{
+				byte[] hashData = data.Skip(offset).Take(count).ToArray();
```
lontivero: I would change this and use the extension method `SafeSubarray` here. In this case it would look like `byte[] hashData = data.SafeSubarray(offset, count);`

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+
+			protected override HashStreamBase CreateHashStream()
+			{
+				return BufferedHashStream.CreateFrom(CalculateHash);
+			}
+
+			public override uint256 GetPoWHash()
+			{
+				
+				return new uint256(this.x13.ComputeBytes(this.ToBytes()));
```
lontivero: I am not familiar with Stratis so, this is a question: doesn't it depend on the `nVersion`?

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+			}
+
+			/// <summary>
+			/// POS Timestamp
+			/// </summary>
+			public uint Time { get; set; } = Utils.DateTimeToUnixTime(DateTime.UtcNow);
+
+			public override void ReadWrite(BitcoinStream stream)
+			{		
+				var witSupported = (((uint)stream.TransactionOptions & (uint)TransactionOptions.Witness) != 0) &&
```
lontivero: This is okay of course but it looks to c++. My suggestion is:

```suggestion
				var witSupported = stream.TransactionOptions.HasFlag(TransactionOptions.Witness) &&
```

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+				JObject obj = JObject.Parse(tx);
+				StratisTransaction stratTx = new StratisTransaction(Stratis.StratisConsensusFactory.Instance);
+				DeserializeFromJson(obj, ref stratTx);
+
+				return stratTx;
+			}
+
+			private static void DeserializeFromJson(JObject json, ref StratisTransaction tx)
+			{
+				tx.Version = (uint)json.GetValue("version");
```
lontivero: This is only a matter of style. I think this looks better:
```suggestion
				tx.Version = json.Value<uint>("version");
```

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+				}
+
+				var vout = (JArray)json.GetValue("vout");
+				for (int i = 0; i < vout.Count; i++)
+				{
+					var jsonOut = (JObject)vout[i];
+					var txout = new TxOut();
+					tx.Outputs.Add(txout);
+
+					var btc = (decimal)jsonOut.GetValue("value");
```
lontivero: If it is possible to avoid using decimal for representing values I would recommend to avoid it. Sooner or later there will be problems because of that.

NBitcoin.Altcoins/Stratis.cs
```diff
@@ -0,0 +1,660 @@
...
+		protected override NetworkBuilder CreateMainnet()
+		{
+			NetworkBuilder builder = new NetworkBuilder();
+			// a large 4-byte int at any alignment.
+			var messageStart = new byte[4];
+			messageStart[0] = 0x70;
+			messageStart[1] = 0x35;
+			messageStart[2] = 0x22;
+			messageStart[3] = 0x05;
+			var magic = BitConverter.ToUInt32(messageStart, 0); //0x5223570;
```
lontivero: Why is this? why not simply do:
```suggestion
			var magic = 0x5223570;
```

NBitcoin.Altcoins/Stratis.cs
mikedennis: Hmmm, in looking at this I question that as well as CalculateHash checks the version. I double checked the stratis code (see below) and it matches what I have here so maybe we need @dangershony to weigh in on whether this is on purpose or not.
https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/NBitcoin/BlockStake.cs#L295-#L303

NBitcoin.Altcoins/Stratis.cs
mikedennis: Cleaned this up a bit. Still left the individual bytes in, just in case someone is comparing it to the Stratis code base, but made it a little less ugly.

mikedennis:
Ok I cleaned up the code a bit according to your suggestions. Thanks! They were all good suggestions, I had done a lot of "cut and paste" from the Stratis code base. All Altcoin tests still pass.

NBitcoin.Altcoins/Stratis.cs
mikedennis: My guess is Stratis POW is always using X13 so doesn't need to look at version but will get Dan to confirm.

NBitcoin.TestFramework/WellknownNodeDownloadData.cs
```diff
@@ -844,6 +846,41 @@ public class ChaincoinNodeDownloadData
...
+				Windows = new NodeOSDownloadData()
+				{
+					DownloadLink = "https://github.com/mikedennis/StratisBitcoinFullNode/releases/download/{0}/stratis-{0}-win64.zip",
+					Archive = "stratis-{0}-win64.zip",
+					Executable = "stratis-{0}-win64/Stratis.StratisD.exe",
+					Hash = "7B0ABEA75B032D8FAF3BEE071A892E2864A31A8ECC42F7AA4300CC51B1CA1D5A"
+				},
+				Linux = new NodeOSDownloadData()
+				{
+					DownloadLink = "https://github.com/mikedennis/StratisBitcoinFullNode/releases/download/{0}/stratis-{0}-linux64.tgz",
```
NicolasDorier: not officially released?

NBitcoin.Tests/NodeBuilderEx.cs
```diff
@@ -49,12 +47,13 @@ public static NodeBuilder Create([CallerMemberName] string caller = null)
...
 			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcoinplus.v2_7_0, Altcoins.AltNetworkSets.Bitcoinplus.Regtest, caller);
 
-			var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_17_0, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
+			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_17_0, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Liquid.v3_14_1_21, Altcoins.AltNetworkSets.Liquid.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcore.v0_15_2, Altcoins.Bitcore.Instance.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Gincoin.v1_1_0_0, Altcoins.Gincoin.Instance.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Koto.v2_0_0, Altcoins.Koto.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Chaincoin.v0_16_4 , Altcoins.AltNetworkSets.Chaincoin.Regtest, caller);
+			var builder = NodeBuilder.Create(NodeDownloadData.Stratis.v3_0_0, Altcoins.AltNetworkSets.Stratis.Regtest, caller);
```
NicolasDorier: put back Bitcoin Core

NBitcoin.TestFramework/WellknownNodeDownloadData.cs
mikedennis: All my Stratis changes have been pushed to master now but binaries have not yet been released.

NicolasDorier:
Should we wait next official version before pushing?

mikedennis:
We can or you can push now and I can do a new PR to update it once binaries are published for the release.

NBitcoin.TestFramework/WellknownNodeDownloadData.cs
bokobza: We're going to build some releases coming from the Stratis repo.

PR #415, opened 2018-05-01
Golomb-Rice filter compatibility

lontivero:
This PR tests the GRFilters against the [BIP158 test vectors](https://github.com/bitcoin/bips/blob/master/bip-0158/testnet-20.csv) that were included in the bip158 about a month ago, after we pushed the golomb-rice filter to NBitcoin.

## Why
In order to achieve binary compatibility with the btcd node and neutrino when we serialize/deserialize filters, I had to change the filters internal representation. Previously we used a BitArray-like class called FastBitArray which was created for performance reasons. That class is removed in this PR in favor of a bigendian bit array data structure.

## Serialization
This PR provides a format for serializing grfilters as `[varint(N) | filter_data_bytes ]` where `|` means concatenation.

## Builders
Additionally, we include a set of useful methods for building grfilters such as:
* **BuildBasicFilter** for building basic filters for a given block and,
* **BuildExtendedFilter** for building suplementary filters with witness data for a given block.

The `GolombRiceFilterBuilder` class provides now convenient methods to make easier to build filters, such as:
* AddScriptPubkey,
* AddScriptSig,
* AddWitness
* AddOutPoint
* AddTxId

## Filters header
Filters can now calculate their header in order to verify the chain of filters is not broken (idem to chain of block headers)

## Missing
Classes/methods need documentation. However I don't want to move forward without feedback.

### DIFF
--- a/NBitcoin/BIP158/GolombRiceFilter.cs
+++ b/NBitcoin/BIP158/GolombRiceFilter.cs
@@ -97,11 +97,10 @@ namespace NBitcoin
 		/// <param name="key">Key used for hashing the datalements.</param>
 		/// <param name="data">Data elements to be computed in the list.</param>
 		/// <returns></returns>
-		internal static List<ulong> ConstructHashedSet(byte P, byte[] key, IEnumerable<byte[]> data)
+		internal static List<ulong> ConstructHashedSet(byte P, int n, byte[] key, IEnumerable<byte[]> data)
 		{
 			// N the number of items to be inserted into the set.
 			var dataArrayBytes = data as byte[][] ?? data.ToArray();
-			var n = dataArrayBytes.Length;
 
 			// The list of data item hashes.
 			var values = new List<ulong>();
@@ -156,7 +155,7 @@ namespace NBitcoin
 			if (data == null || !data.Any())
 				throw new ArgumentException("data can not be null or empty array.", nameof(data));
 
-			var hs = ConstructHashedSet(P, key, data);
+			var hs = ConstructHashedSet(P, N, key, data);
 
 			var lastValue1 = 0UL;
 			var lastValue2 = hs[0];
@@ -497,7 +496,7 @@ namespace NBitcoin
 		public GolombRiceFilter Build()
 		{
 			var n = _values.Count;
-			var hs = GolombRiceFilter.ConstructHashedSet(_p, _key, _values);
+			var hs = GolombRiceFilter.ConstructHashedSet(_p, n, _key, _values);
 			var filterData = Compress(hs, _p);
 
 			return new GolombRiceFilter(filterData, n, _p);
--- a/NBitcoin.Tests/Bip158_tests.cs
+++ b/NBitcoin.Tests/Bip158_tests.cs
@@ -127,7 +127,7 @@ namespace NBitcoin.Tests
 			// per block.
 			const byte P = 20;
 			const int blockCount = 100;
-			const int maxBlockSize = 4000_000;
+			const int maxBlockSize = 4_000_000;
 			const int avgTxSize = 250;                  // Currently the average is around 1kb.
 			const int txoutCountPerBlock = maxBlockSize / avgTxSize;
 			const int avgTxoutPushDataSize = 20;        // P2PKH scripts has 20 bytes.


NicolasDorier:
Let me know when ready to merge so I review

nopara73:
If you want to use `HashSet` to prevent duplicated entries you must create a dedicated class and implement equality.

NBitcoin.Tests/NBitcoin.Tests.csproj
```diff
@@ -161,6 +161,9 @@
     <Content Include="data\block169482.txt">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
+    <Content Include="data\testnet-20.csv">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
```
nopara73: I'd rather hardcode it. What do you think @NicolasDorier ?

NBitcoin/BIP158/BitStream.cs
```diff
@@ -4,62 +4,165 @@
...
+			_buffer[lastIndex] |= (byte)(b >> (8 - _remainCount));
+
+			AddZeroByte();
+			_buffer[lastIndex + 1] = (byte)(b << _remainCount);
+		}
+
+		public bool ReadBit()
+		{
+			if (_buffer.Length == 0)
+				throw new InvalidOperationException("The stream is empty");
```
nopara73: Can you add punctuation, like this: (`"The stream is empty."`) so to comply with Microsoft's best practices? https://docs.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#use-grammatically-correct-error-messages

NBitcoin/BIP158/GolombRiceFilter.cs
```diff
@@ -73,6 +54,13 @@ internal static List<ulong> ConstructHashedSet(byte P, int N, byte[] key, IEnume
 			return values;
 		}
 
+		internal uint256 GetHeader(uint256 previousHeader)
+		{
+			var curFilterHashBytes = Hashes.Hash256(ToBytes()).ToBytes();
+			var prvFilterHashBytes =previousHeader.ToBytes();
```
nopara73: space

lontivero:
> @NicolasDorier:  HashSet will only prevent duplicated arrays (reference wise), not content. I am not sure it is useful.

and

> @nopara73   If you want to use HashSet to prevent duplicated entries you must create a dedicated class and implement equality.

You're right. That's why I've created with a `ByteArrayComparer` instance allowing us to prevent duplicated arrays _based on their elements_ instead of the array reference. See: https://github.com/MetacoSA/NBitcoin/pull/415/files#diff-916615c5885ddc38ca3f9edea8fbf2a3R236

I've tested it with the block height=926485 which pushes duplicated elements into the filter.

NicolasDorier:
I did not know you could pass a comparer to Hashset.

NBitcoin.Tests/NBitcoin.Tests.csproj
lontivero: I've given the file a better name.

lontivero:
> Let me know when ready to merge so I review

@NicolasDorier I think this is ready to merge. Could you take a look at it? Also, if you think it is ready and merge it, could you generate a new `nuget package` version?

NicolasDorier:
Yes, reviewing now.

NBitcoin/BIP158/BitStream.cs
```diff
@@ -4,62 +4,165 @@
...
-			_buffer = bitArray;
-			_position = 0;
+			var newBuffer = new byte[buffer.Length];
+			Buffer.BlockCopy(buffer, 0, newBuffer, 0, buffer.Length);
+			_buffer = newBuffer;
+			_remainCount = buffer.Length == 0 ? 0 : 8;
 		}
 
-		public bool ReadBit()
+		private void AddZeroByte()
```
NicolasDorier: @lontivero why not using `Array.Resize` ?

NBitcoin/BIP158/BitStream.cs
```diff
@@ -4,62 +4,165 @@
...
+			{
+				throw new InvalidOperationException("End of stream reached.");
+			}
+
+			b |= (byte)(_buffer[0] >> _remainCount);
+			_buffer[0] <<= (8 - _remainCount);
+			return b;
 		}
 
 		public byte[] ToByteArray()
```
NicolasDorier: `ToArray()` should make a copy. `GetBuffer()` should not.

PR #421, opened 2018-05-15
New GetRawTransactionInfo method

lontivero:
This PR adds a new `GetRawTransactionInfo` method to the `RpcClient`
class because the exisiting `GetRawTransaction` method returns simply a
`Transaction` instance and sometimes developers also need useful
informations such as `BlochHash` and/or `Confirmations` for checking
whether the tx is in the blockchain or not; in which block it is and how
mature it is.

Note this new method returns redundant information that can
be extracted directly from the also-returned transaction. IMO this is
not something wrong (I can remove it too)

Having two methods with similar names and porpouse doesn't sound a good
idea IMO, maybe we could mark `GetRawTransaction(Async)` as `Obsolete`
and remove them in some version in the future.

### DIFF
--- a/NBitcoin/RPC/RPCClient.cs
+++ b/NBitcoin/RPC/RPCClient.cs
@@ -1167,6 +1167,31 @@ namespace NBitcoin.RPC
 			return tx;
 		}
 
+		public TransactionInfo GetRawTransactionInfo(uint256 txid)
+		{
+			return GetRawTransactionInfoAsync(txid).GetAwaiter().GetResult();
+		}
+
+		public async Task<TransactionInfo> GetRawTransactionInfoAsync(uint256 txId)
+		{
+			var request = new RPCRequest(RPCOperations.getrawtransaction, new object[]{ txId.ToString(), true });
+			var response = await SendCommandAsync(request);
+			var json = response.Result;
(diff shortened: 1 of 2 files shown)


nopara73:
Thanks @lontivero, I was too lazy to open an issue for this, but you figured out my thoughts :D

NBitcoin/RPC/RPCClient.cs
```diff
@@ -1562,6 +1587,21 @@ public class Bip9SoftFork
 		public List<Bip9SoftFork> Bip9SoftForks { get; set; }
 	}
 
+	public class TransactionInfo
```
nopara73: Shouldn't it be called rather `RawTransactionInfo`? I think `TransactionInfo` is too generic.

lontivero: Sure. I will do it.

lontivero:
Please @NicolasDorier  do NOT merge yet (i've found a defect: transaction time is null when it is confirmed). Fixing now

NicolasDorier:
I can now?

lontivero:
Yes, it's done. The original test was extended to cover the buggy scenario too so, it is ready.

PR #425, opened 2018-05-17
Support Feathercoin

ChekaZ:
+ 6 out of 7 Tests work. ( CanSyncWithPoW doesnt )
+ Tests needed to be modified a bit ( Min relay fee & Block creation, FTC needs 120 Confirmations until mature )

- Missing -> Binary links + Hash in WellKnownData


NicolasDorier:
I changed the CoinMaturity setting of FeatherCoin from 100 to 120.
Because all tests everywhere assume 101, you might need to replace with

```
node.Generate(101);
```

with
```
node.Generate(builder.Network.Consensus.CoinbaseMaturity + 1);
```

In tests in NBXplorer for example.

PR #434, opened 2018-05-25
Update Polis Ext Prefixes

cronos-polis:


cronos-polis:
Hey @NicolasDorier We are going to start using this Library for one project, what we love is the altcoin support.

PR #439, opened 2018-05-31
Update bip 158 test vectors

lontivero:
The bip 158 test vectors are now in json format (using the original file) and there is a new tested scenario. Additionally, the vectors use the new serialization which consists of N as a varint. This means
an empty filter is serialized as "00" instead of "" (empty). 
**NOTE**: The NBitcoin code was already using this serialization format, that's why no changes in the code were required. 
See this BIP PR for more info: https://github.com/bitcoin/bips/commit/fe1d3f84ba260bb39a49861564c6bbf1f371f801#diff-36d11b70d5a30a00085e9bc4699b35b1

By last, there were a missing changes after the dotnet update to 2.1 that are fixed here too.


lontivero:
Btw, the CI builds fail because neither travis nor appveyor support netcoreapp2.1 yet. We had the same problem with Wasabi Wallet today (See: https://github.com/zkSNACKs/WalletWasabi/pull/262#issuecomment-393416267)

NicolasDorier:
ah shit, we should dockerize the tests

PR #444, opened 2018-06-07
Crypto.Pbkdf2: make class public

knocte:
In the same way SCrypt is already public, having
Pbkdf2 be public can also be useful (for example
to implement the WarpWallet algorithm without
depending on CryptSharp).

### DIFF
--- a/NBitcoin/Crypto/Cryptsharp/Pbkdf2.cs
+++ b/NBitcoin/Crypto/Cryptsharp/Pbkdf2.cs
@@ -50,7 +50,8 @@ namespace NBitcoin.Crypto
 	/// Stream derivedKeyStream = new Pbkdf2(new HMACSHA512(key), salt, 1000);
 	/// </code>
 	/// </example>
-	internal class Pbkdf2 : Stream
+	[Obsolete("This might disappear in favour of .NET BCL's Rfc2898DeriveBytes class")]
+	public class Pbkdf2 : Stream
 	{
 		#region PBKDF2
 		byte[] _saltBuffer, _digest, _digestT1;
@@ -74,7 +75,7 @@ namespace NBitcoin.Crypto
 		/// </param>
 		/// <param name="iterations">The number of iterations to apply.</param>
 #if USEBC || WINDOWS_UWP || NETCORE
-		public Pbkdf2(IMac hmacAlgorithm, byte[] salt, int iterations)
+		internal Pbkdf2(IMac hmacAlgorithm, byte[] salt, int iterations)
 		{
 			NBitcoin.Crypto.Internal.Check.Null("hmacAlgorithm", hmacAlgorithm);
 			NBitcoin.Crypto.Internal.Check.Null("salt", salt);
@@ -138,7 +139,7 @@ namespace NBitcoin.Crypto
 		/// <param name="derivedKeyLength">The desired length of the derived key.</param>
 		/// <returns>The derived key.</returns>
 #if USEBC || WINDOWS_UWP || NETCORE
-		public static byte[] ComputeDerivedKey(IMac hmacAlgorithm, byte[] salt, int iterations,
+		internal static byte[] ComputeDerivedKey(IMac hmacAlgorithm, byte[] salt, int iterations,
 											   int derivedKeyLength)
 		{
 			NBitcoin.Crypto.Internal.Check.Range("derivedKeyLength", derivedKeyLength, 0, int.MaxValue);
--- a/NBitcoin/NBitcoin.csproj
+++ b/NBitcoin/NBitcoin.csproj
@@ -12,7 +12,7 @@
 		<RepositoryType>git</RepositoryType>
 	</PropertyGroup>
 	<PropertyGroup>
-		<Version Condition=" '$(Version)' == '' ">4.1.1.10</Version>
+		<Version Condition=" '$(Version)' == '' ">4.1.1.11</Version>
 		<LangVersion>7.2</LangVersion>
 	</PropertyGroup>
 	<PropertyGroup>


NBitcoin/Crypto/Cryptsharp/Pbkdf2.cs
```diff
@@ -138,7 +138,7 @@ public byte[] Read(int count)
 		/// <param name="derivedKeyLength">The desired length of the derived key.</param>
 		/// <returns>The derived key.</returns>
 #if USEBC || WINDOWS_UWP || NETCORE
-		public static byte[] ComputeDerivedKey(IMac hmacAlgorithm, byte[] salt, int iterations,
+		internal static byte[] ComputeDerivedKey(IMac hmacAlgorithm, byte[] salt, int iterations,
```
lontivero: This change makes the class useful for your project but it is not usable for others using NETCORE or UWP.

It is clear that making `ComputeDerivedKey` comes with problems. However, if I am working on a netcore app, i will have the `Pbkdf2` class available but without one of its main methods to invoke! After all, `pbkdf` means password based key derivation function, and `ComputeDerivedKey` is all what this class is about.

NBitcoin/Crypto/Cryptsharp/Pbkdf2.cs
```diff
@@ -74,7 +74,7 @@ internal class Pbkdf2 : Stream
 		/// </param>
 		/// <param name="iterations">The number of iterations to apply.</param>
 #if USEBC || WINDOWS_UWP || NETCORE
-		public Pbkdf2(IMac hmacAlgorithm, byte[] salt, int iterations)
+		internal Pbkdf2(IMac hmacAlgorithm, byte[] salt, int iterations)
```
lontivero: Given the class is now public but its only constructor is internal, if i am in a netcore project then how can I create and new instance?

knocte:
@lontivero I marked those as internal because IMac was internal (otherwise it would give a compiler error). Do you want me to also mark IMac as public, in order to address your .NETCore concerns?

lontivero:
@NicolasDorier is the one that decides whether something is acceptable or not so. IMO unless that require a long chain of changes, it would be good to have an similar interface for different platforms, yes.

knocte:
Ok, let's wait for his feedback then.

NicolasDorier:
Mh I would prefer keeping this internal to not get conflicts with people referencing CryptoSharp.
However, I am wondering why this is stripped for NETCore, have you looked? maybe we don't need this in .NET Core because it is inside the base library.

knocte:
> to not get conflicts with people referencing CryptoSharp.

I don't think there would be conflicts because you have placed these classes inside a NBitcoin.Crypto namespace (and IMac is inside NBitcoin.BouncyCastle.Crypto one); plus, as I said in the PR description and commit message, Scrypt class is already public (which also comes from CryptSharp).

> However, I am wondering why this is stripped for NETCore, have you looked? maybe we don't need this in .NET Core because it is inside the base library.

Can you check @lontivero ?

lontivero:
.NET already has the `Rfc2898DeriveBytes` class in both core and net. I have never used it by the way, but I am almost sure it is available in both versions.

lontivero:
Oh wait, I've just remembered why I have never use `Rfc2898DeriveBytes`, because it used HMACSHA1 and bip32 requires HMACSHA256. So, I don't think .net has something like that in the base library, I should google for a while.

knocte:
What does this mean for this PR? That it's fine already and .NET Core would see the overloads then? (If merged.)

NicolasDorier:
@knocte why would you need this public ? Using the base library seems best. (btw, it seems that NBitcoin does not, but it should)

knocte:
I'm lost, where's Pbkdf2 in the BCL and how does Rfc2898DeriveBytes relate to it?

NicolasDorier:
Pbkdf2 is `Rfc2898DeriveBytes` + `HMACSHA256`, which can be easily used in .NET Core. (right @lontivero ?)

So better for you to just use this than NBitcoin.

knocte:
Ah ok let me test this.

lontivero:
Yes. PBKDF1 and PBKDF2 specifications are part of the [RFC 2898](https://www.ietf.org/rfc/rfc2898) and that's why the .net class is called `Rfc2898DeriveBytes`. Given it is an old rfc, at that time MD5 and SHA1 were the default hashing algorithms, it seems. That's why I thought it was only for those hashing algos. However the  `Rfc2898DeriveBytes` also allow to specify the hashing algo to use.

So, @knocte to make all this crystal clear: Pbkdf2 ==  Rfc2898DeriveBytes

Now, going back to NBitcoin, we only use Pbkdf2 for implementing Bip32, I think. In that case we could remove the existing class and replace it by the one in the .net base library, yes.

lontivero:
It seems to be pretty new (added in .NET 4.7.2). Well it is a pity.

knocte:
Can we merge this then? I could add an [Obsolete("Will get removed once NBitcoin depends on 4.7.2 or higher")] attribute if you guys prefer.

lontivero:
It seems we won't be able to use the class that comes with the base library because it doesn't accept a `salt` argument with length less than 8 bytes.

I have a `Pbkdf2` class with exactly the same interface than the BCL's `Rfc2898DeriveBytes` so  I only had to change one line to test this. Sadly, in BIP38 we use `salt = HASH256(address)[0..3]` and then, it is not accepted by `Rfc2898DeriveBytes` that fails with:

```
Error Message:
   System.ArgumentException : Salt is not at least eight bytes.
Parameter name: salt
```

We can see this in the source code here: https://referencesource.microsoft.com/#mscorlib/system/security/cryptography/rfc2898derivebytes.cs,135

knocte:
Mmm, then, conclusion?
Seems like this PR might be the least worse of all evils.

NicolasDorier:
It is possible that the 8 bytes limitation is math related and silently padded by Pbkdf2.

NicolasDorier:
@knocte yeah put Obsolete for now.

knocte:
> @knocte yeah put Obsolete for now.

Done @NicolasDorier

NicolasDorier:
@knocte I merged, but can you remove the warnings with #pragma when we build?

knocte:
> @knocte I merged, but can you remove the warnings with #pragma when we build?

Sorry for the late reply. I just cloned the repo in a Windows machine with .NET Core 2.1 and didn't see any warnings, did you already fix this?

NicolasDorier:
probably

knocte:
It seems that NBitcoin, when compiled for .NET Standard, is following the NETCORE ifdef.

I thought I was not going to need this as public inside NETCORE because I thought NetStandard didn't map with NETCORE, however when compiling a .NET Standard lib that uses NBitcoin, the symbol is not seen. Can I mark these as public also then, even inside NETCORE?

knocte:
(in a new PR, that is)

NicolasDorier:
Maybe better fixing. Netstandrd should not define `NETCORE`. For which environment do you need netstandard for information?

PR #450, opened 2018-06-21
Fix builder with no custom parameters

lontivero:
@NicolasDorier this PR fixes a defect when we use custom (non-default) value for the `M` parameter (something we do in Wasabi Wallet). 

We have added more tests covering the previously uncovered code.

I have never before asked you to do this but, could you merge it and generate a new version with this PR, please?

Thank you very much.


NicolasDorier:
done `.13`

PR #452, opened 2018-06-25
Performnce improvemnt for GR Filters

lontivero:
Previous version trew InvalidOperationExceptions for signaling the end
of the Stream. This version simply check the availability of bits.


nopara73:
@NicolasDorier Could you roll out a NuGet while we're sleeping (during the next 9 hours)?

NicolasDorier:
done

PR #454, opened 2018-06-26
Update Feathercoin.cs

ChekaZ:
- CookieAuth


NicolasDorier:
So the bug for cookieAuth was only on testnet?

PR #457, opened 2018-07-02
Golomb rice filter performance improvement

lontivero:
This PR fixes the allocation problem detected by @NicolasDorier which was hurting the performance. Just to have an idea about the improvement, these are the numbers.

**Before**
```
Filters generation time: 00:00:24.0559429
Filters matching time: 00:00:25.8740358
Filters matching false time: 00:00:01.5151029
```

**After**
```
Filters generation time: 00:00:02.9757793
Filters matching time: 00:00:00.4871420
Filters matching false time: 00:00:01.3413390
```


nopara73:
Wow!

PR #465, opened 2018-07-08
Schnorr signature/verification

lontivero:
# Schnorr signatures
This PR contains the first version of the `Schnorr signature and verification algorithms` as defined in the BIP proposal https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki.

### Batch verification
We not implement signatures batch verifications becuase I don't know how to implement the step:
Let y = c<sup>(p+1)/4</sup> mod p

### Integration into NBitcoin existing classes
We didn't integrate this new types into the existing classes because it is not clear how to do it without having conflicts with the `ECDSA` methods. The `ECDSA signature/verification methods` are in the `ECKey` and `Key` classes. 

We prefer let this decision on the hands of the maintainer @NicolasDorier. 

### Security considerations
A secured signature implementation should be constant time execution and this is not. Probably the ECDSA signature algorithm in BC is not constant time neither.

### Jacobi
We didn't review the correctness of the `BigInteger.Jacobi` method because we didn't try to understand the math involved in that method. Instead, we ran that method in parallel with the `Jacobi` method used in the BIP:

---
**Python**
```python
p = 73

def jacobi(x):
    return pow(x, (p - 1) // 2, p)

for n in range(0, p):
    print(jacobi(n)) 
```
**C#**
```c#
var p = 73;

for(var n=0; i<p; i++){
    Console.WriteLine(BigInteger.Jacobi(n, p));
}
```

Both methods' results are the same.
---

### What next
We would like to implement the the following algorithms soon:
* Signatures batch verification
* Signature aggregations
* Blind Schnorr signatures

### Merge
This PR can be merge in order to start play with Schnorr signatures. It is not dangeroues and it would be good to have some feedback about it in real world situations.


NBitcoin/Crypto/SchnorrSignature.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Threading.Tasks;
+using NBitcoin.BouncyCastle.Asn1.X9;
+using System.Security.Cryptography;
+
+namespace NBitcoin.Crypto
+{
+	public class SchnorrSigner
+	{
+
+        public byte[] Sign(uint256 m, byte[] secret)
```
NicolasDorier: `secret` should be of type `uint256`.

NBitcoin/Crypto/SchnorrSignature.cs
```diff
@@ -0,0 +1,86 @@
...
+            if r >= p or s >= n:
+                return False
+            e = sha256(sig[0:32] + bytes_point(pubkey) + msg)
+            R = point_add(point_mul(G, s), point_mul(pubkey, n - e))
+            if R is None or jacobi(R[1]) != 1 or R[0] != r:
+                return False
+            return True
+        */
+
+        public bool Verify(uint256 m, byte[] pubkey, byte[] sig)
```
NicolasDorier: `pubkey` should be of type `PubKey`. `sig` should have its own type like `SchnorrSignature`.

NBitcoin/BouncyCastle/math/BigInteger.cs
```diff
@@ -1523,47 +1523,47 @@ internal bool RabinMillerTest(int certainty, Random random, bool randomlySelecte
...
-		//				b = a1;
-		//			}
-		//			return totalS;
-		//		}
+		
+						int a1Lsw = a1.magnitude[a1.magnitude.Length - 1];
+						if ((bLsw & 3) == 3 && (a1Lsw & 3) == 3)
+							totalS = -totalS;
+		
+		//				a = b.Mod(a1);
```
NicolasDorier: should this be commented?

NBitcoin/BouncyCastle/math/BigInteger.cs
```diff
@@ -1523,47 +1523,47 @@ internal bool RabinMillerTest(int certainty, Random random, bool randomlySelecte
...
+						int bLsw = b.magnitude[b.magnitude.Length - 1];
+						if ((e & 1) != 0 && ((bLsw & 7) == 3 || (bLsw & 7) == 5))
+							totalS = -totalS;
+		
+						// TODO Confirm this is faster than later a1.Equals(One) test
+						if (a.BitLength == e + 1)
+							break;
+						BigInteger a1 = a.ShiftRight(e);
+		//				if (a1.Equals(One))
 		//					break;
```
NicolasDorier: should this be commented?

NBitcoin/Crypto/SchnorrSignature.cs
lontivero: There are two overloaded `Sign` methods, one receives a `Key` instance and the other one receives a `BigInteger` instance.

NBitcoin/Crypto/SchnorrSignature.cs
lontivero: Done

NBitcoin/BouncyCastle/math/BigInteger.cs
lontivero: No, the whole method was commented and the Java implementation is very different to the c# one. Anyway, I've compared the results of this implementation against the BIP implementation (`pow(x, (y - 1) // 2, y)`) and it seems to be correct in all the cases.

I've removed the commented garbage anyway.

NBitcoin/BouncyCastle/math/BigInteger.cs
lontivero: Yes, because it uses the remainder what is the same as mod.

NBitcoin/Crypto/SchnorrSignature.cs
```diff
@@ -0,0 +1,100 @@
...
+			var keyPrefixedM = Utils.BigIntegerToBytes(Xr, 32).Concat( P.GetEncoded(true), m.ToBytes(false) ); 
+			var e = new BigInteger(1, Hashes.SHA256(keyPrefixedM));
+
+			var s = k.Add(e.Multiply(secret)).Mod(Secp256k1.N);
+			return new SchnorrSignature(Xr, s);
+		} 
+
+		public bool Verify(uint256 m, PubKey pubkey, SchnorrSignature sig)
+		{
+			return Verify(m, pubkey.ToBytes(), sig);
```
NicolasDorier: you are doing too much work here. From `PubKey` you can already have the `ECKey`. Here you are converting in byte[] for later parsing it again!

NBitcoin/Crypto/SchnorrSignature.cs
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		public bool Verify(uint256 m, byte[] pubkey, SchnorrSignature sig)
```
NicolasDorier: I think `byte[] pubkey` overload should not exist

NBitcoin/Crypto/SchnorrSignature.cs
```diff
@@ -0,0 +1,100 @@
...
+		public static SchnorrSignature Parse(string hex)
+		{
+			var bytes = Encoders.Hex.DecodeData(hex);
+			return new SchnorrSignature(bytes);
+		}
+
+		public SchnorrSignature(byte[] bytes)
+		{
+			if(bytes.Length != 64)
+				throw new ArgumentException("Invalid schnorr signature lenght.");
```
NicolasDorier: typo => `length`

NBitcoin/Crypto/SchnorrSignature.cs
lontivero: I've removed the overloaded method.

NBitcoin/Crypto/SchnorrSignature.cs
lontivero: fixed.

NBitcoin/Crypto/SchnorrSignature.cs
lontivero: removed.

PR #467, opened 2018-07-10
Added GetMemPoolInfo RPC call

misnor:
Adding functionality so the RPC Client can call GetMemPoolInfo.

### DIFF
--- a/NBitcoin/RPC/RPCClient.cs
+++ b/NBitcoin/RPC/RPCClient.cs
@@ -1076,12 +1076,12 @@ namespace NBitcoin.RPC
 
 			return new MemPoolInfo()
 			{
-				Size = Int32.Parse((string)response.Result["size"]),
-				Bytes = Int32.Parse((string)response.Result["bytes"]),
-				Usage = Int32.Parse((string)response.Result["usage"]),
-				MaxMemPool = Double.Parse((string)response.Result["maxmempool"]),
-				MemPoolMinFee = Double.Parse((string)response.Result["mempoolminfee"]),
-				MinRelayTxFee = Double.Parse((string)response.Result["minrelaytxfee"])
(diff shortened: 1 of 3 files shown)


NBitcoin/RPC/RPCClient.cs
```diff
@@ -1065,6 +1065,26 @@ public int GetBlockCount()
...
+			return this.GetMemPoolAsync().GetAwaiter().GetResult();
+		}
+
+		public async Task<MemPoolInfo> GetMemPoolAsync()
+		{
+			var response = await SendCommandAsync(RPCOperations.getmempoolinfo);
+
+			return new MemPoolInfo()
+			{
+				Size = Int32.Parse((string)response.Result["size"]),
```
NicolasDorier: Can you pass `CultureInfo.InvariantCulture` as second parameter of those `Parse` methods? Else it will explode for some users.

NBitcoin/RPC/RPCClient.cs
```diff
@@ -1065,6 +1065,26 @@ public int GetBlockCount()
 			return (int)(await SendCommandAsync(RPCOperations.getblockcount).ConfigureAwait(false)).Result;
 		}
 
+		public MemPoolInfo GetMemPool()
```
nopara73: Could you create its async pair, too?

nopara73:
ACK, the code looks good.

NBitcoin/RPC/RPCClient.cs
misnor: Of course! There are 14 other `Parse` (mostly `uint`) without the CultureInfo provided in `RPCClient.cs`, should I add them there as well or just focus on what was added?

NicolasDorier:
included in 4.1.1.26

PR #473, opened 2018-07-16
ECDSA Blinding signature protocol

lontivero:
This PR is for proposing an Blinding Signature Algorithm over using ECC. 

## Goal
Blinding signatures are useful in many project involving cryptocurrencies (Wasabi wallet is one of those) and for that reason I propose to include this classes as part of NBitcoin.

## Protocol
```c#
// Requester represents the participant that requests the signature from the signer
var requester = new ECDSABlinding.Requester();

var signerKey = new Key();  // The key used for signing the message
var r = new Key();          // A value used to derivate a r.Pubkey that has to be sent to the requester 

// Signer represents the one that has to sign the message
var signer = new ECDSABlinding.Signer(signerKey, r);

// 1. Signer sends the r.Pubkey and the signer.Pubkey to the requester 
// 2. requester blinds the message and sends it to the signer
var blindedMessage = requester.BlindMessage(message, r.PubKey, signerKey.PubKey);

// 3. Signer signs the blinded message and returns that blindedSignature to the requester
var blindSignature = signer.Sign(blindedMessage);

// 4. The requester "unblinds" the blinded signature and gets the unblinded signature
var unblindedSignature = requester.UnblindSignature(blindSignature);

// 5. Anyone how konws the message, the and the signer.PubKey can verify the signature
var isValid = ECDSABlinding.VerifySignature(message, unblindedSignature, signerKey.PubKey);
```

## References
The algorithm details are specified on the following resources:
 * [A Blind Mixing Scheme for Bitcoin based on an Elliptic Curve Cryptography
Blind Digital Signature Algorithm](https://arxiv.org/pdf/1510.05833.pdf)
* [Blind ECDSA Signatures](https://en.wikipedia.org/wiki/Blind_signature#Blind_ECDSA_signatures[4])
* [A blind digital signature scheme using elliptic curve digital signature algorithm](https://pdfs.semanticscholar.org/399d/d5695dda5c730d58240e675f8b7823cd2adb.pdf)

### DIFF
--- a/NBitcoin.Tests/DeterministicSignatureTests.cs
+++ b/NBitcoin.Tests/DeterministicSignatureTests.cs
@@ -254,75 +254,75 @@ namespace NBitcoin.Tests
 		public void BlindingSignature()
 		{
 			// Test with known values 
-			var requester = new ECdsaBlinding.Requester();
-			var signingKey = new Key(Encoders.Hex.DecodeData("31E151628AED2A6ABF7155809CF4F3C762E7160F38B4DA56B784D9045190CFA0"));
-			var verificationKey = new Key(Encoders.Hex.DecodeData("B7E151628AED2A6ABF7158809CF4F3C762E7160F38B4DA56A784D9045190CFEF"));
-			var signer = new ECdsaBlinding.Signer(signingKey, verificationKey);
+			var requester = new ECDSABlinding.Requester();
+			var r = new Key(Encoders.Hex.DecodeData("31E151628AED2A6ABF7155809CF4F3C762E7160F38B4DA56B784D9045190CFA0"));
+			var key = new Key(Encoders.Hex.DecodeData("B7E151628AED2A6ABF7158809CF4F3C762E7160F38B4DA56A784D9045190CFEF"));
+			var signer = new ECDSABlinding.Signer(key, r);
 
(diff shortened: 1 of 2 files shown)


NicolasDorier:
That's cool! Can you rename `ECdsaBlinding` to `ECDSABlinding`?

nopara73:
https://en.wikipedia.org/wiki/Blind_signature#Blind_ECDSA_signatures[4]

> Another scheme for blind signatures is possible by using elliptic curve cryptography, allowing shorter key size, faster computations and avoids the unwanted mathematical properties of RSA described above.

What's the difference between Schnorr blinding and this kind of blinding?

lontivero:
> What's the difference between Schnorr blinding and this kind of blinding?

@nopara73 this signature algorithm uses ECC and schnorr signatures does not. This means that in this signature scheme we add and multiply points that belong to the seckp256k1 curve. On the other hand in the schnorr signature algorithm there are no operations involving EC points in any curve. That means that schnorr signatures should be faster than ECDSA-based signatures (ec operations are a bit slow). 

> That's cool! Can you rename ECdsaBlinding to ECDSABlinding?

@NicolasDorier sure. I've realize that I can improve the naming in other elements too. I will fix this asap.

PR #474, opened 2018-07-16
Prevent loosing money by parsing small values

lontivero:
This PR is for fixing the issue https://github.com/zkSNACKs/WalletWasabi/issues/489 when someone parses values that involves fractions of satoshis. For example:

```
Money.Parse("0.00000000000001");  will be 0
Money.Parse("12.66548765432");  will be 12.6654876 loosing the 0.00000005432
```

This is a **breaking change** so, please review this and let us know if it okay for you @NicolasDorier

### DIFF
--- a/NBitcoin.Tests/util_tests.cs
+++ b/NBitcoin.Tests/util_tests.cs
@@ -352,6 +352,9 @@ namespace NBitcoin.Tests
 
 				// Attempted 63 bit overflow should fail
 				Assert.False(Money.TryParse(prefix + "92233720368.54775808", out ret));
+
+				// Attempted parsing a very small value (loosing money)  should fail
+				Assert.False(Money.TryParse(prefix + "0.00000000000001", out ret));
 			}
 		}
 
--- a/NBitcoin.Tests/util_tests.cs
+++ b/NBitcoin.Tests/util_tests.cs
@@ -355,6 +355,9 @@ namespace NBitcoin.Tests
 
 				// Attempted parsing a very small value (loosing money)  should fail
 				Assert.False(Money.TryParse(prefix + "0.00000000000001", out ret));
+				Assert.False(Money.TryParse(prefix + "0.000000001", out ret));
+				Assert.True(Money.TryParse(prefix + "0.00000000", out ret));
+				Assert.True(Money.TryParse(prefix + "0.00000000000", out ret));
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


lontivero:
@knocte `Money.Parse` doesn't throw a new exception. However, the constructors and the factory methods such as `Money.Satoshis`, `Money.Cents`, `Money.Coins` and others now can throw a new `ArgumentOutOfRangeException`. This means that:

* this is the breaking change that can break existing projects and,
* you cannot loose money if you don't explicitly say so.

About the latest point, this example shows the point:
```c#
decimal value = 10 / 3m;
var m1 = Money.Satoshis(value);  // Fails with OOR exception
var m2 = Money.Satoshis((long)value);  // With the cast YOU are explicitly removing the fractional part 
```

Anyway, you are probably right on this but before writing more tests lets see what our benevolent dictatorship has to say about it.

nopara73:
Doesn't Parse should throw `FormatException`? It feels to me, like that's the convention. I may be wrong on this, though.

lontivero:
@nopara73 you are right, that's the exception is is throwing. It is the same exception it was throwing before.

NBitcoin.Tests/util_tests.cs
```diff
@@ -352,6 +352,9 @@ public void util_ParseMoney()
 
 				// Attempted 63 bit overflow should fail
 				Assert.False(Money.TryParse(prefix + "92233720368.54775808", out ret));
+
+				// Attempted parsing a very small value (loosing money)  should fail
+				Assert.False(Money.TryParse(prefix + "0.00000000000001", out ret));
```
NicolasDorier: Can you add test for `Assert.False(Money.TryParse("0.000000001"))` ? (Should fail with 9 decimal)

And `Assert.True(Money.TryParse("0.0000000000000"))` (Should work with 0)

lontivero: Done!

NicolasDorier:
Sorry I needed to revert that, it breaks in lots of my project. :(

Making a any decimal operation is very likely to fire this problem, it broke BTCPay :(

lontivero:
Ack. It was clear to me that this was going to break existing projects.

NicolasDorier:
I knew it would, but did not anticipated that almost any divide operation would break.

PR #475, opened 2018-07-17
Add hex string constructor to ExtPubKey

osagga:
Now you can construct `ExtPubKey` with a hex string in addition to the byte array constructor. This behavior is needed for the symmetry between `ExtPubKey` and `ExtKey`

### DIFF
--- a/NBitcoin/BIP32/ExtPubKey.cs
+++ b/NBitcoin/BIP32/ExtPubKey.cs
@@ -72,11 +72,22 @@ namespace NBitcoin
 		{
 		}
 
+        	/// <summary>
+		/// Constructor. Creates a new extended public key from the specified extended public key bytes.
+		/// </summary>
 		public ExtPubKey(byte[] bytes)
 		{
 			if(bytes == null)
 				throw new ArgumentNullException(nameof(bytes));
 			this.ReadWrite(bytes);
+		}
+        
+		/// <summary>
+		/// Constructor. Creates a new extended public key from the specified extended public key bytes, from the given hex string.
+		/// </summary>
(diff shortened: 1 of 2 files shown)


osagga:
I mean people already can use string seeds with `ExtKey`, I think you mean `ExtPubKey`. But I would argue the same with a `byte[]` input. It could be that people pass a private seed input to `ExtPubKey` and getting wrong results back.

NicolasDorier:
Ok,

1. Can you fix the indent?
2. Do not call it a `seed`, it is not a seed, it is the extpubkey serialized as bytes. `hex` is better naming
3. Can you add a small `///<summary>` comment like I did on `ExtKey` on the byte[] and the new string ctor?

osagga:
Done

NBitcoin/BIP32/ExtPubKey.cs
```diff
@@ -72,12 +72,23 @@ internal ExtPubKey()
 		{
 		}
 
+        	/// <summary>
```
nopara73: Unneccessary whitespace.

NicolasDorier:
And unecessary `Constructor`, of course it is a constructor! :D

NBitcoin/BIP32/ExtPubKey.cs
nopara73: https://github.com/MetacoSA/NBitcoin/pull/477

osagga: It's really weird, the formatting seems to be inconsistent on Linux (I didn't see the whitespace in my editor), and the same with the indentation problem I fixed above. Good catch!

PR #489, opened 2018-07-27
[WIP] Bitcoin Private

ch4ot1c:
- ForkId is set as 42 (for `42 << 8`), but we also use `(1U << 16)` for SCRIPT_VERIFY_FORKID. This should be added in a way that doesnt break BTG or BCH.

- The Genesis blocks in `BPrivate.cs` are pretty big; let me know if I should remove them for now.

Thank You

### DIFF
--- a/NBitcoin.Altcoins/BPrivateTransaction.cs
+++ b/NBitcoin.Altcoins/BPrivateTransaction.cs
@@ -1,53 +1,26 @@
-﻿using System;
+using System;
 using System.Linq;
-using System.Collections.Generic;
-using System.Text;
 using NBitcoin.Crypto;
 
 namespace NBitcoin.Altcoins
 {
-    public class ForkIdTransaction : Transaction, IHasForkId
-	{
-#pragma warning disable CS0618 // Type or member is obsolete
-		public ForkIdTransaction(uint forkId, bool supportSegwit, ConsensusFactory consensusFactory)
+  	public class BPrivateTransaction : ForkIdTransaction
+  	{
+		public BPrivateTransaction(ConsensusFactory consensusFactory) : base(42, false, consensusFactory)
 		{
-			_ForkId = forkId;
-			_SupportSegwit = supportSegwit;
-			_Factory = consensusFactory;
-		}
-#pragma warning restore CS0618 // Type or member is obsolete
-
-		ConsensusFactory _Factory;
-		public override ConsensusFactory GetConsensusFactory()
-		{
-			return _Factory;
-		}
-
-		private readonly bool _SupportSegwit;
-		public bool SupportSegwit
-		{
-			get
-			{
-				return _SupportSegwit;
-			}
-		}
-
-		private readonly uint _ForkId;
-		public uint ForkId
-		{
-			get
-			{
-				return _ForkId;
-			}
+			// BTCP is a fork of Zclassic
+			// FORKID_BTCP = 42: https://github.com/BTCPrivate/BitcoinPrivate/blob/4031ff02ec7c56bcafa085b01100cbddfcd33ea3/src/script/interpreter.h#L41
+			// No Segwit (For now)
 		}
 
 		public override uint256 GetSignatureHash(Script scriptCode, int nIn, SigHash nHashType, Money amount, HashVersion sigversion, PrecomputedTransactionData precomputedTransactionData)
 		{
 			uint nForkHashType = (uint)nHashType;
-			if((nHashType & SigHash.ForkId) != 0)
+			if((nForkHashType & (uint)SigHash.ForkId) != 0)
 				nForkHashType |= ForkId << 8;
 
-			if((SupportSegwit && sigversion == HashVersion.Witness) || (nHashType & SigHash.ForkId) != 0)
+			/* TODO Segwit compat
+			if(false && (SupportSegwit && sigversion == HashVersion.Witness) || (nHashType & SigHash.ForkId) != 0)
 			{
 				if(amount == null)
 					throw new ArgumentException("The amount of the output being signed must be provided", "amount");
@@ -58,13 +31,13 @@ namespace NBitcoin.Altcoins
 				if((nHashType & SigHash.AnyoneCanPay) == 0)
 				{
 					hashPrevouts = precomputedTransactionData == null ?
-								   GetHashPrevouts() : precomputedTransactionData.HashPrevouts;
+									GetHashPrevouts() : precomputedTransactionData.HashPrevouts;
 				}
 
 				if((nHashType & SigHash.AnyoneCanPay) == 0 && ((uint)nHashType & 0x1f) != (uint)SigHash.Single && ((uint)nHashType & 0x1f) != (uint)SigHash.None)
 				{
 					hashSequence = precomputedTransactionData == null ?
-								   GetHashSequence() : precomputedTransactionData.HashSequence;
+									GetHashSequence() : precomputedTransactionData.HashSequence;
 				}
 
 				if(((uint)nHashType & 0x1f) != (uint)SigHash.Single && ((uint)nHashType & 0x1f) != (uint)SigHash.None)
@@ -101,9 +74,7 @@ namespace NBitcoin.Altcoins
 
 				return GetHash(sss);
 			}
-
-
-
+			*/
 
 			if(nIn >= Inputs.Count)
 			{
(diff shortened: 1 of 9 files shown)


NBitcoin.Altcoins/ForkIdTransaction.cs
```diff
@@ -47,7 +47,7 @@ public override uint256 GetSignatureHash(Script scriptCode, int nIn, SigHash nHa
 			if(UsesForkId(nHashType))
 				nForkHashType |= ForkId << 8;
 
-			if((SupportSegwit && sigversion == HashVersion.Witness) || UsesForkId(nHashType))
```
NicolasDorier: Keep ` || UsesForkId(nHashType)` please, I don't want to test again bcash and bgold.

NBitcoin/Script.cs
```diff
@@ -121,14 +121,18 @@ public enum ScriptVerify : uint
 		/// <summary>
 		/// Some altcoins like BCash and BGold requires ForkId inside the sigHash
 		/// </summary>
-		ForkId = (1U << 29),
+		ForkId = (1U << 16),
```
NicolasDorier: Do not change this value.

NBitcoin.Altcoins/ForkIdTransaction.cs
```diff
@@ -47,7 +47,7 @@ public override uint256 GetSignatureHash(Script scriptCode, int nIn, SigHash nHa
...
+			if((SupportSegwit && sigversion == HashVersion.Witness) || (UsesForkId(nHashType) && ForkId != 42))
```
NicolasDorier: Please do not change this code. I have no way to know that it did not break bgold or bcash.

NBitcoin/Script.cs
```diff
@@ -117,6 +117,13 @@ public enum ScriptVerify : uint
 		/// Public keys in segregated witness scripts must be compressed
 		/// </summary>
 		WitnessPubkeyType = (1U << 15),
+		
+		/// <summary>
+		/// SCRIPT_VERIFY_FORKID flag for BPrivate
+		/// https://github.com/BTCPrivate/BitcoinPrivate/blob/4031ff02ec7c56bcafa085b01100cbddfcd33ea3/src/script/interpreter.h#L130
+		/// https://github.com/BTCGPU/BTCGPU/commit/777899b682c3d59702a4223c8239705a74616a63#diff-b81dfdd8a5bd80fe9f82b5a40c4c991eR126
+		/// </summary>
+		BTCPForkId = (1U << 16),
```
NicolasDorier: Remove this, this is not used anywhere, and `Script.cs` should not be polluted by altcoin specific logic.

NBitcoin/Network.cs
```diff
@@ -1459,13 +1459,41 @@ public uint Magic
 			}
 		}
 
+		private bool isBTCPForkEnabled(int nHeight)
+		{
+			return nHeight > 272991; // nForkStartHeight
+		}
+
+		// TODO - use
+		public Money GetRewardBTCP(int nHeight)
```
NicolasDorier: Remove this, no altcoin specific logic in `Network.cs`.

NBitcoin.Altcoins/ForkIdTransaction.cs
ch4ot1c: How would you prefer I skip over this conditional for BTCP?

NBitcoin/Script.cs
ch4ot1c: Note that this is used as a magic number in https://github.com/MetacoSA/NBitcoin/blob/4b989108b44d7add50f28b777e0495d09949f311/NBitcoin/Policy/StandardTransactionPolicy.cs#L194

NBitcoin/Network.cs
ch4ot1c: Done

NBitcoin/Script.cs
ch4ot1c: (Done)

NBitcoin.Altcoins/ForkIdTransaction.cs
NicolasDorier: I would override `GetSignatureHash` with a class deriving from `ForkIdTransaction`, say BPrivateTransaction.

NicolasDorier: Or Making UseForkId virtual, and only overriding this in BPrivateTransaction

NBitcoin.Altcoins/BPrivateTransaction.cs
```diff
@@ -0,0 +1,149 @@
...
+			// No Segwit (For now)
+		}
+
+		public override uint256 GetSignatureHash(Script scriptCode, int nIn, SigHash nHashType, Money amount, HashVersion sigversion, PrecomputedTransactionData precomputedTransactionData)
+		{
+			uint nForkHashType = (uint)nHashType;
+			if(UsesForkId(nHashType))
+				nForkHashType |= ForkId << 8;
+
+			/* TODO Segwit compat
```
NicolasDorier: Remove the comments

NicolasDorier:
If I compare with what you did before, a better idea seemed to me to make `UseForkId` virtual in the ForkIdTransaction and only override this with the specific condition of BPrivate.

ch4ot1c:
The above is a more semantically correct approach; BCC and BTG tests still all pass. I'm not sure where `1U << 29` comes from, but these changes don't modify the code's function at all.

Let me know if the virtual method is still what you require instead.

Thanks for the review

NBitcoin/ScriptEvaluationContext.cs
```diff
@@ -1679,9 +1679,9 @@ bool IsDefinedHashtypeSignature(byte[] vchSig)
 			}
 
 			var temp = ~(SigHash.AnyoneCanPay);
-			if((ScriptVerify & ScriptVerify.ForkId) != 0)
```
NicolasDorier: No altcoin logic in NBitcoin code. ForkId is an exception for something multiple coin will use. If this does not handle your case then you need to refactor NBitcoin in a more general way which make very obvious to me that there is no change of behavior for what existed before.

NBitcoin/Script.cs
```diff
@@ -119,16 +119,24 @@ public enum ScriptVerify : uint
 		WitnessPubkeyType = (1U << 15),
 
 		/// <summary>
-		/// Some altcoins like BCash and BGold requires ForkId inside the sigHash
+		/// Verify ForkId's presence when evaluating the script
+		/// https://github.com/Bitcoin-ABC/bitcoin-abc/blob/d46be71cdae64b50207a78b2804968f919cac408/src/script/bitcoinconsensus.h#L63
+		/// https://github.com/BTCGPU/BTCGPU/blob/0f0bb60bbd2a6aee990f813387980ee9a5f248df/src/script/interpreter.h#L122
+		/// https://github.com/BTCPrivate/BitcoinPrivate/blob/4031ff02ec7c56bcafa085b01100cbddfcd33ea3/src/script/interpreter.h#L130
 		/// </summary>
-		ForkId = (1U << 29),
```
NicolasDorier: do not change this

NicolasDorier:
You should not change anything in NBitcoin library, all your changes must be located inside `NBitcoin.Altcoins`. Please I don't want to take the risk of breaking something because of the support of another alt.

ch4ot1c:
You're running me in circles here.

There is already fork related code in NBitcoin: https://github.com/MetacoSA/NBitcoin/blob/4b989108b44d7add50f28b777e0495d09949f311/NBitcoin/Policy/StandardTransactionPolicy.cs#L194

Also, overriding `UsesForkId` alone doesn't fix the issue, since it needs to evaluate to true on ForkIdTransaction line47, and false on line 50. `96f83f8` should be sufficient.

NicolasDorier:
I know, `ForkId` has been done so other altcoins could easily customize the "forkId". By not doing so on BPrivate, you don't fit into what other altcoins does. This is a problem at BPrivate level, and your code is changing logic inside NBitcoin.

What you can do is to try to refactor `IHasForkId/ForkIdTransaction` to be more flexible, so that the logic which is specific to your coin stay contained inside `NBitcoin.Altcoins`.

This put additional work for you, but this is the price to pay for not doing things like other altcoin does.

NicolasDorier:
In other words, you first need to refactor NBitcoin code to be more flexible. (This would need to be in separate PR so I can properly evaluate if you do not break anything),

Then you submit your PR for including BPrivate which use this extensibility point.

But you should never touch the NBitcoin code with altcoin specific logic.

NicolasDorier:
need rebase

PR #493, opened 2018-07-31
Schnorr signatures batch verification

lontivero:
This PR is for adding the _Schnorr signatures batch verification_ algorithm described in the [Sipa's Schnorr signature BIP proposal](https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki#Batch_verification).

The same algorithm is available for python [here](https://gist.github.com/lontivero/1d75b11e9243c0f2a75466db05776623) with a more readable syntax in case the BouncyCastle nested and chained methods are too hard to understand. 

More unit tests are needed. In fact, more tests are needed for all the BIP.


NicolasDorier:
Awesome!

PR #495, opened 2018-08-01
Minimal changes to add BTCP

ch4ot1c:
Perform `GetSignatureHash` correctly for BTCP's ForkId `42`.

### DIFF
--- a/NBitcoin.Altcoins/ForkIdTransaction.cs
+++ b/NBitcoin.Altcoins/ForkIdTransaction.cs
@@ -47,7 +47,7 @@ namespace NBitcoin.Altcoins
 			if(UsesForkId(nHashType))
 				nForkHashType |= ForkId << 8;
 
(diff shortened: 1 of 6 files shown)


NBitcoin.Altcoins/ForkIdTransaction.cs
```diff
@@ -47,7 +47,7 @@ public override uint256 GetSignatureHash(Script scriptCode, int nIn, SigHash nHa
 			if(UsesForkId(nHashType))
 				nForkHashType |= ForkId << 8;
 
-			if((SupportSegwit && sigversion == HashVersion.Witness) || UsesForkId(nHashType))
+			if((SupportSegwit && sigversion == HashVersion.Witness) || (UsesForkId(nHashType) && ForkId != 42))
```
NicolasDorier: As I said, do not do this. Make your own class which derive from ForkId, and overrride `UsesFork` to always returns true.

PR #496, opened 2018-08-02
[WIP] Add mogwaicoin, please add.

darkfriend77:
I haven't added the C# native neoscrypt implementation, because it will need to check unsafe code for compilation, if this isn't a problem I can add the source, otherwise it is on our github (https://github.com/mogwaicoin/NeoScryptCSharp).

I hope this fits your pull request requirements, if not let me know what is missing.


darkfriend77:
Please add or tell us what we have to change ... to have the PR merged.

NBitcoin.Altcoins/Mogwai.cs
```diff
@@ -0,0 +1,237 @@
...
+				return new MogwaiBlockHeader();
+			}
+			public override Block CreateBlock()
+			{
+				return new MogwaiBlock(new MogwaiBlockHeader());
+			}
+		}
+
+#pragma warning disable CS0618 // Type or member is obsolete
+		public class MogwaiBlockHeader : BlockHeader
```
NicolasDorier: you do not need this. Mogwai block format is same as bitcoin

NBitcoin.Altcoins/Mogwai.cs
```diff
@@ -0,0 +1,237 @@
...
+			}
+
+			protected override HashStreamBase CreateHashStream()
+			{
+				return BufferedHashStream.CreateFrom(CalculateHash);
+			}
+
+		}
+
+		public class MogwaiBlock : Block
```
NicolasDorier: you do not need this.

NBitcoin.Altcoins/Mogwai.cs
```diff
@@ -0,0 +1,237 @@
...
+		public static Mogwai Instance { get; } = new Mogwai();
+
+		public override string CryptoCode => "MOG";
+
+		private Mogwai()
+		{
+
+		}
+
+		public class MogwaiConsensusFactory : ConsensusFactory
```
NicolasDorier: you do not need this.

darkfriend77:
thx for the reply, I removed unused code. In last commit.

NBitcoin.Altcoins/Mogwai.cs
darkfriend77: okay, thank you for the hint I removed it in latest commit. It was copy paste from dash ... :-)

NBitcoin.Altcoins/Mogwai.cs
darkfriend77: Thx, for the hint, I removed it in the latest commit, If you could accept merge request now, we would be happy. :-)

NicolasDorier:
You tried to run the Altcoins tests with your node ?

darkfriend77:
yes, I had issues with CanSignTransactions, CanSyncWithoutPoW, CanSyncWithPoW, CorrectCoinMaturity ... so I figured out it was because we had already activated DGW (drkGravityWaves Difficulty Algo) in reg test, which avoided generating more then 24 blocks without a diff increase.

So I changed that in regtest, but I also thought I would change BIP34, which seems to be wha did broke it again ... I gona fix that

darkfriend77:
I will do a clean pull request once I got all test working. Sry for the lost time.

PR #498, opened 2018-08-03
Add WordList.GetWords overloaded version

lontivero:
This PR adds an `WordList.GetWords()` method overload that returns the full list of words. These words are needed to implement an user friendly way to introduce the seed in wallets through autocompletion. Without a way to retrieve the list of world, wallets have to keep their own list (hardcoded or downloaded from internet into a file) 

This is what I want to do in Wasabi Wallet:
https://github.com/zkSNACKs/WalletWasabi/issues/440#issuecomment-410120805

### DIFF
--- a/NBitcoin/BIP39/Wordlist.cs
+++ b/NBitcoin/BIP39/Wordlist.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
@@ -347,9 +348,9 @@ namespace NBitcoin
 			return _Name;
 		}
 
-		public string[] GetWords()
+		public ReadOnlyCollection<string> GetWords()
 		{
-			return (string[])_words.Clone();
+			return new ReadOnlyCollection<string>(_words);
 		}
 
 		public string[] GetWords(int[] indices)
--- a/NBitcoin/BIP39/Wordlist.cs
+++ b/NBitcoin/BIP39/Wordlist.cs
@@ -347,6 +347,11 @@ namespace NBitcoin
 			return _Name;
 		}
 
+		public string[] GetWords()
+		{
+			return (string[])_words.Clone();
+		}
+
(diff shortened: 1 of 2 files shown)


NBitcoin/BIP39/Wordlist.cs
```diff
@@ -347,6 +347,11 @@ public override string ToString()
 			return _Name;
 		}
 
+		public string[] GetWords()
+		{
+			return (string[])_words.Clone();
```
nopara73: Why not return an IEnumerable?

nopara73: There's an unspoken contract between developers that nobody should modify a returned IEnumerable, unless you were the one who returned it: )

lontivero: This method is overloaded version of an already-existing one and for that reason i think it should return the same type that the original method. 
Given we return a clone, we shouldn't care if the caller modifies its elements.

knocte: It doesn't matter that it wraps an already existing one, using IEnumerable allows to modify later the method to, for example, return a List instead of an Array, without changing the API.

NicolasDorier:
There is also the read-only collection. If you want to give a try to it, let's use this. Else, I don't mind merging that.

knocte:
The most efficient way (to not require high memory usage) of returning the elements would be using `yield`. However, you would not be able to use `yield` with ReadOnlyCollection AFAIK (while you could with IEnumerable).

NicolasDorier:
The problem with IEnumerable is that you can't use indexer `[]` which is very convenient for BIP39.

lontivero:
@NicolasDorier it is your decision. I know there is a ReadOnlyCollection collection however I try to respect the maintainer's style and design trade offs so, if you used an array then I use an array. However, IMO, an array is perfectly okay here.

NicolasDorier:
ok, @lontivero can you use a `ReadOnlyCollection` ? The other overload is OK because we need to create a new array anyway. But this one is just doing a copy. If this is pain in the ass, let's keep array.

lontivero:
Done!

Issue #502, opened 2018-08-07
Node/network payload code fails


misternode:
I have tried this on several coins, and several nodes (all of which are functioning).  Vis Studio 2017 15.7.6, with .NET Core 2.1.

results in: throw new InvalidOperationException("The node is not in a connected state");  (in 		public TPayload ReceivePayload<TPayload>(CancellationToken cancellationToken = default(CancellationToken))
)

How do I get your library to work?

```
using NBitcoin;
using NBitcoin.Protocol;
using System;
using System.Threading;

namespace ConsoleApp1
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");

            var node = Node.Connect(Network.Main, "52.14.153.105:8333");

            node.VersionHandshake();
        }
    }
}
```

misternode:
I HAVE tried my own nodes.  It fails on EVERY node.

knocte:
Are you sure the port is open from your location?

misternode:
Yes. 100%.

NicolasDorier:
I ran exactly the same code and it worked.

```
static void Main(string[] args)
{   
    var node = Node.Connect(Network.Main, "52.14.153.105:8333");
    Console.WriteLine("Handshaking");
    node.VersionHandshake();
    Console.WriteLine("Connected");
}
```

What OS ?

misternode:
Windows 10, 32 bit.

The TCP connection is made, but no data is sent.  .NET throws InvalidOperationException("The node is not in a connected state") and drops the connection with no data transfered.

NicolasDorier:
Can you try with a local node of yours? You run `bitcoind.exe -regtest` then use `Node.ConnectToLocal(Network.RegTest)`.

rnewsham:
Ran regtest locally and using ConnectToLocal().  Same issue.

misternode:
Just to be thorough, also installed .NET Core 2.1 on linux.

Same results:  Unhandled Exception: System.InvalidOperationException: The node is not in a connected state

NicolasDorier:
Trying to reproduce

NicolasDorier:
```bash
mkdir temp
cd temp
dotnet --version  #2.1.4
dotnet new console
dotnet add package NBitcoin
vim Program.cs
```

Copy pasting 
```bash
using System;
using NBitcoin.Protocol;
using NBitcoin;

namespace temp
{
    class Program
    {
[...]

PR #503, opened 2018-08-08
Add Mogwaicoin

darkfriend77:
commented out neoscrypt for csharp
added mogwaicoin node download


darkfriend77:
Okay, cleaned up code and all tests okay, except CanSyncWithoutPoW & CanSyncWithPoW ... they fail with InvalidOperationException("The node is not in a connected state") ...

NicolasDorier:
`CanSyncWithoutPoW` should not fail.

NicolasDorier:
Can you try to run the same tests on your machine but using Bitcoin instead of  Mogwaicoin?

This might be a dup of https://github.com/MetacoSA/NBitcoin/issues/502

darkfriend77:
I did the same tests with Bitcoin and they succeed all .... Mogwaicoin Tests fail during node versionhandshake,

*var payload = listener.ReceivePayload<Payload>(cancellationToken);*

nodenotconnected00
nodenotconnected01
nodenotconnected02
nodenotconnected03png

darkfriend77:
Okay somehow this code shows the same Node not connected error for bitcoin too ...
```
using NBitcoin.Tests;
using System;

namespace TestNode
{
    class Program
    {
        static void Main(string[] args)
        {
			{
				Console.WriteLine(builder.Network.NetworkType);
				Console.WriteLine(builder.Network.NetworkSet);
				Console.ReadKey();
				builder.StartAll();
				var nodeClient = node.CreateNodeClient();
				nodeClient.VersionHandshake();
			}
		}
    }
}
 
```

image

darkfriend77:
The versionhandshake fixed now, we had the magic bytes inverted.

CanSyncWithoutPoW still fails ...

Meldung: Assert.Equal() Failure
Expected: typeof(NBitcoin.BlockHeader)
Actual:   typeof(NBitcoin.Altcoins.Mogwai+MogwaiBlockHeader)

this seems a legit fail isn't it?

NicolasDorier:
Remove `MogwaiBlockHeader`, you don't use it.

darkfriend77:
```
		public class MogwaiBlockHeader : BlockHeader
		{
			// https://github.com/mogwaicoin/mogwai/blob/master/src/primitives/block.cpp
			static byte[] CalculateHash(byte[] data, int offset, int count)
			{
				var output = new byte[32];
				// reference https://github.com/mogwaicoin/NeoScryptCSharp for correct neoscrypt native c# hashing
				// haven't added source because it will need to have unsafe code checked for the project.
				NeoScrypt.NeoScrypt.neoscrypt(data, ref output, 0x0);
				return output;
			}

			protected override HashStreamBase CreateHashStream()
			{
				return BufferedHashStream.CreateFrom(CalculateHash);
			}

		}
```

NicolasDorier:
It is not what you committed, sorry I don't have time to help you. Moreover, the test has a comment.

darkfriend77:
It's okay you don't need to help I will fix it. For clarification, I have Neoscrypt C# implementation done, so I want to have all Test succeed. I think you wouldn't like unsafe code commited in a PR, that's why I have commented out Neoscrypt and removed the algo in my PR, but for my purpose I want to have it, of course removing MogwaiBlockHeader fix the test, but not if you I want to use the Neoscrypt algo. So I will do it like dash and readd the code to have CanSyncWithoutPoW succeed.

NicolasDorier:
There is difference between PoWHash and BlockHash in some crypto currency.
If neoscrypt is only PoWHash, I advise you to not worry about `CanSyncWithPow`.

darkfriend77:
For Mogwaicoin it is like Dash and for Dash all tests work, however we have hashPrevBlock & hashMerkleRoot as little endian in our blockheader not like dash and bitcoin big endian ... is there an easy way to change that. Then I think last test will succeed.

kodxana:
@darkfriend77 Can you contact me on discord? Madiator2011#2793 i have some questions for you ;)

NicolasDorier:
It is possible to fix, but I don't have time to tell you how nor to test your integration sorry :/

darkfriend77:
You can merge it. It passes the tests.

https://github.com/MetacoSA/NBitcoin/pull/503#issuecomment-412061298

I will not worry about the CanSyncWithPoW :-)

PR #510, opened 2018-08-16
Make sure ECDSA signatures use LowR values

lontivero:
This PR is for making sure ECDSA signatures use lor R values making the DER-formatted signature length = 70 bytes. This same change is based on [the Andrew Chow's PR](https://github.com/bitcoin/bitcoin/pull/13666) in bitcoin core code base. If well that PR is not merged yet, the discussion there is not about the correctness or security of this change but about implementation details and conflicts with other soon-to-be-merged PRs instead.

I think this change is safe but it would be good @NicolasDorier review it deeply. 

Btw, I've tested it with the test below but I didn't include it in this PR because I think it is unnecessary. Anyway, I can add it to the project if someone think it is a good idea.
```c#
[Fact]
public void Signatures_use_low_R()
{
    var rnd = new Random();
    for(var i=0; i < 1_000_000; i++)
    {
        var key = new Key();
        var msgLen = rnd.Next(10, 1000);
        var msg = new byte[msgLen];
        rnd.NextBytes(msg);

        var sig = key.Sign(Hashes.Hash256(msg));
        Assert.True(sig.IsLowR && sig.ToDER().Length <= 70);
    }
}
```

### DIFF
--- a/NBitcoin/BouncyCastle/crypto/signers/ECDsaSigner.cs
+++ b/NBitcoin/BouncyCastle/crypto/signers/ECDsaSigner.cs
@@ -1,3 +1,4 @@
+using System;
 using NBitcoin.BouncyCastle.Crypto.Parameters;
 using NBitcoin.BouncyCastle.Math;
 using NBitcoin.BouncyCastle.Math.EC;
@@ -19,9 +20,11 @@ namespace NBitcoin.BouncyCastle.Crypto.Signers
 		protected ECKeyParameters key = null;
 		protected SecureRandom random = null;
 
+		protected bool forceLowR = false;
+
 		/**
-         * Default configuration, random K values.
-         */
+		* Default configuration, random K values.
+		*/
 		public ECDsaSigner()
 		{
 			this.kCalculator = new RandomDsaKCalculator();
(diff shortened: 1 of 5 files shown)


NBitcoin/BouncyCastle/crypto/signers/ECDsaSigner.cs
```diff
@@ -99,16 +103,22 @@ public virtual BigInteger[] GenerateSignature(byte[] message)
...
-					// 5.3.3
-					r = p.AffineXCoord.ToBigInteger().Mod(n);
+						// 5.3.3
+						r = p.AffineXCoord.ToBigInteger().Mod(n);
+					}
+					while(r.SignValue == 0);
+					rBytes = r.ToByteArrayUnsigned();
 				}
-				while(r.SignValue == 0);
+				while(forceLowR && rBytes[0] >= 0x80);
```
NicolasDorier: nit: Replace `rBytes[0]` by `r.ToByteArrayUnsigned()[0]` so the array is created only when strictly needed, and you dont' need to declare `rBytes` above.

NicolasDorier:
Cool I wanted that.

lontivero:
Yes are right. It's done. I've added a test too.

nopara73:
@achow101 if you have a minute you could also take a quick look at this PR (though it's already merged.)

achow101:
Looks correct to me.

PR #513, opened 2018-08-22
Download bitcoin binaries from bitcoincore.org

lontivero:
Yes, this is my first, and I hope also the last, political PR. Not sure if it worth the effort.


nopara73:
ACK. Could you do it for Wasabi, too?

NicolasDorier:
Cool, but please do not remove 0.16.0, instead just add 0.16.2

PR #532, opened 2018-09-16
Dystem support

danystem:
Hi Nicolas,

I have added support for Dystem ([https//dystem.io](url)) - this appears to be the first PIVX fork added (which unfortunately is quite far behind Bitcoin Core & Dash).

This doesn't have the generate RPC command implemented, so most of the tests against regtest network are currently failing. I looking at implementing this in the core code base currently. However NBitcoin does appear to be working in the most part as I can connect to nodes across the network and issues commands.

Also to note the following:

- I have added support for Quark hashing algorithm alongside X11.
- Also modified the RPC client slightly to accept nulls on the fork flags.

Looking forward to hearing your thoughts?

Dan 
dan@dystem.io


danystem:
Will reopen this to previous mistake.

PR #533, opened 2018-09-16
Dystem support

danystem:
Hi Nicolas,

I have added support for Dystem (https//dystem.io) - this appears to be the first PIVX fork added (which unfortunately is quite far behind Bitcoin Core & Dash).

This doesn't have the generate RPC command implemented, so most of the tests against regtest network are currently failing. I looking at implementing this in the core code base currently. However NBitcoin does appear to be working in the most part as I can connect to nodes across the network and issues commands.

Also to note the following:

I have added support for Quark hashing algorithm alongside X11.
Also modified the RPC client slightly to accept nulls on the fork flags.

Looking forward to hearing your thoughts?

Dan
dan@dystem.io

PS - Sorry made a bit of a faux pas with my previous PR.

### DIFF
--- a/NBitcoin.Altcoins/AltcoinNetworkSets.cs
+++ b/NBitcoin.Altcoins/AltcoinNetworkSets.cs
@@ -13,6 +13,7 @@ namespace NBitcoin.Altcoins
 		public static Dash Dash { get; } = Dash.Instance;
 		public static Mogwai Mogwai { get; } = Mogwai.Instance;
 		public static Dogecoin Dogecoin { get; } = Dogecoin.Instance;
+	    public static Dystem Dystem { get; } = Dystem.Instance;
 		public static Litecoin Litecoin { get; } = Litecoin.Instance;
 		public static Feathercoin Feathercoin { get; } = Feathercoin.Instance;
 		public static Viacoin Viacoin {get; } = Viacoin.Instance;
@@ -31,6 +32,7 @@ namespace NBitcoin.Altcoins
 			yield return Feathercoin;
 			yield return Viacoin;
 			yield return Dogecoin;
+			yield return Dystem;
 			yield return BCash;
 			yield return BGold;
 			yield return Polis;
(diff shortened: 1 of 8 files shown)


eabz:
Nice! I was just starting to port a PIVX fork.

eabz:
@NicolasDorier 
Hey Nicolas. Sorry to spam you, but is there a chance we can merge this. 
Trying to work out on our wallet, but we need some other algos.

danystem:
Hey @NicolasDorier appreciate your probably busy but any idea when you might get a chance to review this?

NicolasDorier:
I was in holiday. If it can't be tested chances are that it will not work properly with NBXplorer and BTCPay. Can you just implement the missing call?

NBitcoin.Altcoins/AltcoinNetworkSets.cs
```diff
@@ -13,6 +13,7 @@ public class AltNetworkSets
 		public static Dash Dash { get; } = Dash.Instance;
 		public static Mogwai Mogwai { get; } = Mogwai.Instance;
 		public static Dogecoin Dogecoin { get; } = Dogecoin.Instance;
+	    public static Dystem Dystem { get; } = Dystem.Instance;
```
NicolasDorier: fix indent

danystem:
Indent fixed dude.

NicolasDorier:
Seems good. Implement `generate` though, because else you will never be able to properly test it.

danystem:
I'm on the case with this now sir! :)

BenjaminNitschke:
@danystem cool stuff, great work, we might be able to port pivx/colx based on that code as well, would be quite useful.

danystem:
@BenjaminNitschke there are a few outstanding issues with Dystem & pivx forks that I am having to solve in the C++ code base, to bring more in line with core.

1. _getheaders_ returns an _inv_ with vectors for blocks which then have to be retrieved with _getdata_. It appears the way btc/dash does it is returns a headers response containing a batch of headers in one go. This makes it too slow to get the chain. 

2. _generate_ is not yet implemented in the Pivx fork code base, this means that most of the tests against the core instance currently fail.

Any thoughts / assistance would be great appreciated.

@NicolasDorier would you be happy to merge this in the mean time? I think at least getting the Quark implementation in there is a useful addition?

PR #537, opened 2018-09-23
[Zclassic] Initial changes - Add Zclassic.cs

ch4ot1c:


NicolasDorier:
Can you rename to `ZClassicNodeDownloadData` (case)

NicolasDorier:
Can you change all Zclassic to ZClassic for consistency.

ch4ot1c:
Both `Zcash` and `Zclassic` are usually lowercase. I can change it but rather not

PR #549, opened 2018-10-05
Add method get list unspent with query_options

Ilchuk-Mihail:


Ilchuk-Mihail:
For some reason - The job exceeded the maximum time limit for jobs, and has been terminated.

knocte:
I think you should retry the build, the tests hang for about once every 10 times or so.

knocte:
mmm still happening, so somehow this PR is making something fail on macOS; you would need to run it on your own to try to determine which of the tests is hanging

NicolasDorier:
Can you use named parameter instead for passing RPC args?

NicolasDorier:
Can you, instead of having an optional arguments, keeping the empty argument overload?

Because changing as optional args might will break stuff that got compiled with another version during runtime.

Ilchuk-Mihail:
@NicolasDorier Not sure, if I understand you correctly. Can you please, add more detail ?

NicolasDorier:
```csharp
public async Task<UnspentCoin[]> ListUnspentAsync();

public async Task<UnspentCoin[]> ListUnspentAsync(ListUnspentOptions options = null, params BitcoinAddress[] addresses);
```

If I call `rpcClient.ListUnspentAsync()` which one of those two get called? because it is ambiguous, it might break some code. Remove the `= null`

Ilchuk-Mihail:
Ahhh ... Sorry, I missed that. But, very strange why the compiler is not issuing an error.

PR #551, opened 2018-10-08
Test againts bitcoin core 0.17.0

lontivero:
Silly PR.


nopara73:
For the record, `signrawtransaction` was removed and the regtest and testnet `bitcoin.conf` entries now need `regtest.` and `test.` prefixes. There might be more things, these definitely will be issues.

NicolasDorier:
Yep... should have been sily PR, but as @nopara73 it is pain in the ass.

NicolasDorier:
I added a commit to make it work

PR #553, opened 2018-10-10
Expand SipHash::SIPROUND for gaining performance

lontivero:
This PR is for improve a bit the `SipHash` algorithm. This is something that would be good to have for those applications that use GolombRice filters (as Wasabi Wallet) because every item you want to match using GRfilters has to be siphashed first. This means if your wallet contains hundreds, or even thousands, addresses, they have to be siphash-ed.


lontivero:
Btw, ShipHash is currently invoked millions times in Wasabi wallet. That's why any improvement in performance here is good for us. I will modify GolombRiceFilter class in order to allow us gain much more in performance but this change would help anyway.

PR #557, opened 2018-10-12
Fix consensuslib DllNotFoundException

lontivero:


NBitcoin.Tests/script_tests.cs
```diff
@@ -404,18 +404,20 @@ private void EnsureHasLibConsensus()
...
+				libConsensusDll = "libbitcoinconsensus.so";
 			}
 			else
 			{
 				throw new NotSupportedException("Unknown operating system");
 			}
 
 			var bitcoinBinFolderPath = Path.GetDirectoryName("TestData");
-			var libConsensusPath = Path.Combine(bitcoinPath, "../../lib", libConsensusDll);
+			var libConsensusPath = Path.Combine(bitcoinPath, "../../lib/", libConsensusDll);
```
NicolasDorier: The slash should not matter.

lontivero: i know. the problem is that everything works okay locally and that why i have to test silly things here.

NicolasDorier:
crazy shit right

NicolasDorier:
The `libbitcoinconsensus.so` is actually a symlink. Not the library itself.

lontivero:
I am getting a different error now. That's good. The problem seems to be that the shared libraries has to be in one of the folders where shared libraries use to be `/usr/lib` for example.

lontivero:
btw, this is the same error I was fighting (and I gave up) in OSX.

lontivero:
okay. Currently it finds the dll but it fails trying to load one of its dependencies: _/usr/lib/x86_64-linux-gnu/libstdc++.so.6_

```
[xUnit.net 00:01:07.1499583]     NBitcoin.Tests.script_tests.script_CHECKMULTISIG23 [FAIL]
Failed   NBitcoin.Tests.script_tests.script_CHECKMULTISIG23
Error Message:
 System.DllNotFoundException : Unable to load shared library './libbitcoinconsensus.dll' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by lib/libbitcoinconsensus.dll)
Stack Trace:
   at NBitcoin.Script.VerifyScriptConsensus(Byte[] scriptPubKey, UInt32 scriptPubKeyLen, Byte[] txTo, UInt32 txToLen, UInt32 nIn, ScriptVerify flags, BitcoinConsensusError& err)
   at NBitcoin.Script.VerifyScriptConsensus(Script scriptPubKey, Transaction tx, UInt32 nIn, ScriptVerify flags, BitcoinConsensusError& err) in /home/travis/build/MetacoSA/NBitcoin/NBitcoin/Script.cs:line 927
   at NBitcoin.Tests.script_tests.AssertValidScript(Script scriptPubKey, Transaction tx, Int32 n, ScriptVerify verify) in /home/travis/build/MetacoSA/NBitcoin/NBitcoin.Tests/script_tests.cs:line 783
   at NBitcoin.Tests.script_tests.script_CHECKMULTISIG23() in /home/travis/build/MetacoSA/NBitcoin/NBitcoin.Tests/script_tests.cs:line 738
File.Delete("./lib/./libbitcoinconsensus.dll");
```

NicolasDorier:
`File.Delete("./lib/./libbitcoinconsensus.dll");` this seems wrong

lontivero:
It is just what `Console.WriteLine($"File.Delete(.....")` line that I added just to see what was going on, after all the CI is a black box.

NicolasDorier:
I mean the path seemed wrong

D3m0nKingx:
This **/usr/lib/x86_64-linux-gnu/libstdc++.so.6** is a symlink to **/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.21**

@lontivero maybe just put a copy of the actual libraries into /usr/lib

NBitcoin/Script.cs
```diff
@@ -891,7 +891,7 @@ public static bool VerifyScript(Script scriptSig, Script scriptPubKey, Transacti
 
 #if !NOCONSENSUSLIB
 
-		public const string LibConsensusDll = "libbitcoinconsensus.dll";
+		public const string LibConsensusDll = "./libbitcoinconsensus.dll";
```
knocte: I would suggest trying just [DllImport("bitcoinconsensus")]. I believe MS.NET has default logic of trying "lib" and ".dll" prefixes, while Mono uses .so for Linux and .dylib on macOS.

knocte: I meant `lib` prefix and `.dll` suffix.

D3m0nKingx:
> `File.Delete("./lib/./libbitcoinconsensus.dll");` this seems wrong

@NicolasDorier that should be complete path `path/to/lib/libbitcoinconsensus.dll` by putting that second ' . ' linux treats it as if trying to execute a program same with the first ' . ' being there.

lontivero:
I didn't look at it again but at least I know what's going on. I have to download a docker image with travis in order to work on it locally. Basically, the problem is `libbitcoinconsensus` has a dependency that has to be installed in one of the shared libraries paths. However it seems that when a library has a dependency, both need to be in the same shared library folder. I am quite sure this is what was happening on OSX too.

knocte:
> However it seems that when a library has a dependency, both need to be in the same shared library folder.

Unless you make use of LD_LIBRARY_PATH? This would be in Linux. I think in macOS the equivalent is called DYLD_LIBRARY_PATH.

NicolasDorier:
@lontivero I think @knocte has a good idea. Maybe by setting LD_LIBRARY_PATH and DYLD_LIBRARY_PATH with `Environment.SetEnvironentVariable` it may works. Though you seem to have an issue with a dependency as well.

lontivero:
Yes, but i would prefer to copy the libbitcoinconsensus to the right directory instead. Not sure yet

NicolasDorier:
You may need to run `ldconfig` ? This project is also install libconsensus https://github.com/libbitcoin/libbitcoin-consensus

We can follow same steps maybe?

lontivero:
@NicolasDorier it seems to work on Windows (and in my local Linux env). Unfortunately travis says that it detected an abuse and doesn't build the project. Once I can verify it works I will consolidate all these commits and clean the code.

NicolasDorier:
anything I can do? it seems I can't build this PR on travis either.

NicolasDorier:
Let me know when you cleaned up, I'll try by merging.

NicolasDorier:
let's try!

NicolasDorier:
nop https://travis-ci.org/MetacoSA/NBitcoin/builds/452302369

PR #564, opened 2018-10-17
Some small performance gain in GRFilter matching

lontivero:


nopara73:
I don't understand what's happening here, but I think this PR caused some issues. It seems like Wasabi filters sometimes aren't matching when they should.

NicolasDorier:
if that's the case, test case missing

PR #566, opened 2018-10-20
Revert "Some small performance gain in GRFilter matching (#564)"

nopara73:
This reverts commit 96d85d64806aab7552506c6e7296383487da2b54.

The commit resulted in false negative filter matches, which should never happen (only false positives are tolerated.)

### DIFF
--- a/NBitcoin/BIP158/BitStream.cs
+++ b/NBitcoin/BIP158/BitStream.cs
@@ -87,21 +87,16 @@ namespace NBitcoin
 			return true;
 		}
 
-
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Ping @lontivero also we need to update tests, tests did not catch that.

NicolasDorier:
Fixed in 4.1.1.59

nopara73:
@lontivero @NicolasDorier The work is is in the queue now: https://github.com/zkSNACKs/WalletWasabi/issues/751

PR #586, opened 2018-11-12
Disable Consensus library for Linux and MacOS

lontivero:


knocte:
I thought it was only failing on macOS?

lontivero:
No. After updating to 0.17.0 it began failling on Linux too.

PR #588, opened 2018-11-14
Add simple starting point to write Property based test

joemphilips:
1. Currently, it is only for very basic stuff (Address and TX serialization)
2. It is quite slow (I'm not sure how to make it fast, using different seed generation algo for KeyGenerator might work.)

Any feedback appreciated.
As a future work, I am planning to implement something like `ConsensusAcceptableTransactionGenerator` and `MempoolAcceptableTransactionGenerator` .

### DIFF
--- a/NBitcoin.Tests/Generators/ScriptGenerator.cs
+++ b/NBitcoin.Tests/Generators/ScriptGenerator.cs
@@ -1,0 +1,104 @@
+using System;
+using System.Linq;
+using FsCheck;
+using NBitcoin;
+using static NBitcoin.Tests.Generators.CryptoGenerator;
+
+namespace NBitcoin.Tests.Generators
+{
+  public class ScriptGenerator
+  {
+
+    #region script sig
+    // -------- legacy -------
+    // 1. p2pkh scriptSig
+    public static Gen<Tuple<Script, Script>> p2pkhScriptSigSet(Key privKey, TransactionSignature sig) =>
+      p2pkhScriptPubKey().Zip(p2pkhScriptSig(privKey.PubKey, sig));
+
+    public static Gen<Script> p2pkhScriptSig(PubKey pk, TransactionSignature sig) =>
+      Gen.Fresh(() => PayToPubkeyHashTemplate.Instance.GenerateScriptSig(sig, pk));
+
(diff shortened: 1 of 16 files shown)


NicolasDorier:
Cool, can you rebase? you are adding a file that I deleted (Readme.txt)

NBitcoin.Tests/Generators/ScriptGenerator.cs
```diff
@@ -0,0 +1,104 @@
...
+        return multiSignatureScriptSig();
+      }
+
+      throw new Exception("Unknown Script PubKey");
+    }
+
+    #endregion
+
+    #region scriptPubKey
+    public static Gen<Script> p2pkhScriptPubKey() =>
```
NicolasDorier: `p2pkhScriptPubKey` => `P2PKHScriptPubKey`

NicolasDorier:
Can you fix the cases ? In C#, we use PascalCase.

joemphilips:
Looks good now?

Where should I put explanation about these tests? The explanation is like following (BTY, I'd like to hear your thoughts about these categories if you have any opinion.)

```
"PropertyTest" are Property based test. i.e. It generates some test data automatically and test against all those data.
For PropertyTest, we split tests into following categories.
(These are an exactly same with the one described in ref: https://fsharpforfunandprofit.com/posts/property-based-testing-2/)

1. Commutativity
2. BidirectionalConversion
3. Immutability
4. Idempotence
5. Induction
6. Verification
7. Oracle
```

joemphilips:
done.

NicolasDorier:
It makes me think that I probably should add a BenchmarkDotnet project to identify where NBitcoin is slow.

NicolasDorier:
I fixed the perf issue on c8dd9ae58624c7fabeb3a257560bb84dd3b34a1b there was too much amount above 21M.

NicolasDorier:
Can you take a look at `Assert.True(tx.HasWitness);` I tried removing the `MaxTest = 10`, then things start breaking as it seems your generator is generating non segwit tx

NicolasDorier:
Found the issue

NicolasDorier:
2fb4f3416ed77fd83681fc54920cc22cf3d002b3

PR #590, opened 2018-11-15
Improve genreator

joemphilips:
* Genrate `SIGHASH_ANYONECANPAY` sighash properly
* do not compute actual hash when genrating uint256 and u160
 * ~This was what causing error in SegwitTransaction property tests. (It seems some there were some odd behavior when pubkey in multisig scripts were exactly the same, which don't need any fix since it doesn't happen in the real world)~ I was wrong, It was simply because WitScript was empty, so revert to what you did before.
* run 100 tests as usual for tx property based tests.

### DIFF
--- a/NBitcoin.Tests/Generators/CryptoGenerator.cs
+++ b/NBitcoin.Tests/Generators/CryptoGenerator.cs
@@ -16,6 +16,11 @@ namespace NBitcoin.Tests.Generators
       return Arb.From(PrivateKey());
     }
 
+    public static Arbitrary<ExtKey> ExtKeysArb()
+    {
+      return Arb.From(ExtKey());
+    }
+
     public static Arbitrary<List<Key>> KeysListArb()
     {
       return Arb.From(PrivateKeys(15));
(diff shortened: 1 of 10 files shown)


joemphilips:
I also improved randomness for scriptSig generator. See commit message for details

joemphilips:
VSCode, I don't know why it is ignoring `.editorconfig` , I will make clear.

NicolasDorier:
You need to open the folder with the solution, not only the project. Maybe it is why?

joemphilips:
Nop. What I found was

1. omnisharp formatter does respect .editorconfig properly. (via `"omnisharp.useEditorFormattingSettings"` , when I opend the )
2 But this was only when I ran `Format Document` command.
3. I was switching off auto formatting since it changes existing codebase
4. So it inserts spaces according to the default setting of vscode ( `"editor.insertSpaces": true` )

So the solution is either one of
1. include `.vscode/settings.json` into repo which specifies`" editor.insertSpaces": false`
2. specify a formatter option more precisely in `.editorconfig.json` , and check the code base format on CI. (enable Devs to auto format)

I think the solution 1 is good enough for temporary workaround, since I think most devs use VS or VSCode. I did it for my own local repo but I am not sure if you want to include it in this repo or not.

PR #596, opened 2018-11-20
 Implement BIP 174 Partially Signed Bitcoin Transaction

joemphilips:
* [x] Implement methods for (de)serialization
* [x] Implement methods to Update PSBT
* [x] Implement a method  to Combine PSBT
* [x] Implement a method to Finalize PSBT
* [x] Implement a method to Extract TX from PSBT
* [ ] Implement RPC


NBitcoin/PubKey.cs
```diff
@@ -348,8 +348,9 @@ public override bool Equals(object obj)
 			PubKey item = obj as PubKey;
 			if(item == null)
 				return false;
-			return ToHex().Equals(item.ToHex());
+			return Equals(item);
 		}
+		public bool Equals(PubKey pk) => ToHex().Equals(pk.ToHex());
```
lontivero: Before this change `Equals` was comparing the PubKey hex string representation what is not great. I think this is a good opportunity for changing that and compare the internal `byte[] vch` array.

Another point here is that `public` methods should check for NULLs. The `IEquatable<PubKey>.Equals()` doesn't check it so, it can throws NRE if `myPubKey.Equals((PubKey)null)`

NBitcoin/Utils.cs
```diff
@@ -416,6 +416,32 @@ public static bool ArrayEqual(byte[] a, int startA, byte[] b, int startB, int le
...
 
+		public static bool NullSafeEquals<T>(T a, T b) where T : IEquatable<T> =>
+			(a == null && b == null) || ((a != null && b != null) && a.Equals(b));
+
+		public static bool DictEqual<T, U>(Dictionary<T, U> x, Dictionary<T, U> y)
+		{
+			if (null == y)
+				return null == x;
+			if (null == x)
+				return false;
```
lontivero: These four lines are not really necessary because you use `object.ReferenceEquals()` that does the same.

NBitcoin/Transaction.cs
```diff
@@ -1352,6 +1356,9 @@ public TxOutList Outputs
 		//Since it is impossible to serialize a transaction with 0 input without problems during deserialization with wit activated, we fit a flag in the version to workaround it
 		protected const uint NoDummyInput = (1 << 27);
 
+		public bool Equals(Transaction other) =>
```
lontivero: Q: If I have the same transaction twice, with and without witnesses, should this method return true or false? @NicolasDorier? 

I am thinking on this scenario: my application maintains its own mempool and that means that for each transaction it receives, it checks if it is already in the list of transactions using this `Equals(Transaction)` method. It could be possible(?) to receive the same transaction from two different nodes using different serialization format. In this case I think this should return true, otherwise i would end with duplicated transactions.

NBitcoin/BIP174/PartiallySignedTransaction.cs
```diff
@@ -439,13 +439,16 @@ public override bool Equals(object obj)
 		}
 		public bool Equals(PSBTInput other) =>
 			IsReferencingSamePrevOut(other) &&
-			Utils.DictEqual(partial_sigs, other.partial_sigs) &&
```
lontivero: I think it could worth to simplify this by comparing the serialized versions. That could be a bit slower the first time but at the same time it could be easier to read and maintain given the amount of code that could be removed.

joemphilips: You mean serialize the whole `PSBTInput` ? What do you think the current one which uses predicate to be passed on `DictEqual` ?

lontivero: I think your implementation is okay. I just wanted to put this tradeof on the table: simplicity vs small performance gain. I always prefer simplicity if performance is not something critical. But this is just my opinion.

NBitcoin/PubKey.cs
NicolasDorier: Also, this is ugly. Should compare the bytes directly or underlying objects, this is less wasteful.

NBitcoin/Transaction.cs
NicolasDorier: We should remove this. Lot's of code depends on the fact that we are comparing reference here. We should keep it that way.

NBitcoin/NBitcoin.csproj
```diff
@@ -58,6 +58,7 @@
 		<PackageReference Include="System.Buffers" Version="4.5.0" />
 		<PackageReference Include="System.Net.Requests" Version="4.3.0" />
 		<PackageReference Include="System.Net.Http" Version="4.3.3" />
+		<PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
NicolasDorier: We should not have to add a dependency

NBitcoin/Transaction.cs
```diff
@@ -553,7 +553,7 @@ private int GetSpecialSize(uint nSize)
 		#endregion
 	}
 
-	public class TxOut : IBitcoinSerializable, IDestination
+	public class TxOut : IBitcoinSerializable, IDestination, IEquatable<TxOut>
```
NicolasDorier: Don't do this, this will not make expected result for crypto currencies with TxOut with other field (like Elements)

joemphilips:
Thanks for review.
I've took @lontivero 's advice and just compare bytes for Equality checking of my new classes.
It still makes `Script` , `WitScript` and `PubKey` makes `IEquatable`(which does not affect PSBT Equality logic anymore) but I guess this is nice-to-have so I have left it in this way.

NBitcoin/PubKey.cs
```diff
@@ -348,8 +348,9 @@ public override bool Equals(object obj)
 			PubKey item = obj as PubKey;
 			if(item == null)
 				return false;
-			return ToHex().Equals(item.ToHex());
+			return Equals(item);
 		}
+		public bool Equals(PubKey pk) => pk != null && ToBytes().SequenceEqual(pk.ToBytes());
```
NicolasDorier: Can you use `NBitcoin.Utils.ArrayEqual(ToBytes(), pk.ToBytes())` it is a bit better.

NBitcoin/BIP32/KeyPath.cs
```diff
@@ -40,6 +40,23 @@ public KeyPath(string path)
 				.ToArray();
 		}
 
+		public static KeyPath FromBytes(byte[] data)
+		{
+			if (data.Length % 4 != 0)
```
lontivero: It is a common practice for public methods to check argument for NULL to prevent NREs. Q: should a zero-length byte array be allowed here?

NBitcoin/BIP32/KeyPath.cs
```diff
@@ -40,6 +40,23 @@ public KeyPath(string path)
...
 
+		public static KeyPath FromBytes(byte[] data)
+		{
+			if (data.Length % 4 != 0)
+				throw new ArgumentOutOfRangeException("data length is not suited for KeyPath");
+			var depth = data.Length / 4;
+			uint[] result = new uint[depth];
+			for (int i = 0;  i < depth; i++)
+			{
+				result[i] = BitConverter.ToUInt32(data, i * 4);
```
NicolasDorier: Use `Utils.ToUInt32` with little endian so we are sure to not have issues.

NBitcoin/Crypto/ECKey.cs
```diff
@@ -65,10 +65,10 @@ public ECDomainParameters DomainParameter
 		}
 
 
-		public ECDSASignature Sign(uint256 hash)
+		public ECDSASignature Sign(uint256 hash, bool UseLowR = true)
```
NicolasDorier: parameters are `camelCase`, why is it useful?

NBitcoin/BIP32/KeyPath.cs
```diff
@@ -40,6 +40,23 @@ public KeyPath(string path)
...
+			}
+
+			return new KeyPath(result);
+		}
+
+		public byte[] ToBytes() =>
+			Indexes.Select(i => BitConverter.GetBytes(i)).Aggregate((a, b) => a.Concat(b)).ToArray();
```
lontivero: This is okay because the `Indexes` has just a couple of elements but for collections with more elements it could be good to use `SpanExtensions .AsBytes()`, `CopyTo`, or `Cast/TryCast` (I don't know if these last methods are still available)

NBitcoin/PubKey.cs
```diff
@@ -351,8 +351,9 @@ public override bool Equals(object obj)
 			PubKey item = obj as PubKey;
 			if(item == null)
 				return false;
-			return ToHex().Equals(item.ToHex());
+			return Equals(item);
 		}
+		public bool Equals(PubKey pk) => pk != null && Utils.ArrayEqual(ToBytes(), pk.ToBytes());
```
lontivero: No need to make a copy of the internal representation to compare them. Use simply the `vch` field or `.ToBytes(true)`

NBitcoin/BIP32/KeyPath.cs
joemphilips: I think it should be allowed to represent path for the master key. I will add test

NBitcoin/BIP32/KeyPath.cs
joemphilips: Sorry for my ignorance but what do you mean by `SpanExtensions` ? I could not found anything besides [internal module of Kestrel](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.internal.system.spanextensions.asbytes?view=aspnetcore-2.0)

NicolasDorier: Meh yes that's fine span does not work on some target, let's keep it like this now.

NBitcoin/ScriptEvaluationContext.cs
```diff
@@ -1934,9 +1934,9 @@ public bool CheckSig(TransactionSignature signature, PubKey pubKey, Script scrip
 			return CheckSig(signature.ToBytes(), pubKey.ToBytes(), scriptPubKey, checker, (int)hashVersion);
 		}
 
-		public bool CheckSig(byte[] vchSig, byte[] vchPubKey, Script scriptCode, Transaction txTo, int nIn)
+		public bool CheckSig(byte[] vchSig, byte[] vchPubKey, Script scriptCode, Transaction txTo, int nIn, int sigVersion = 0, TxOut spentOutput = null)
```
NicolasDorier: Can you make one new

`public bool CheckSig(byte[] vchSig, byte[] vchPubKey, Script scriptCode, Transaction txTo, int nIn, int sigVersion, TxOut spentOutput)`

and call that from `public bool CheckSig(byte[] vchSig, byte[] vchPubKey, Script scriptCode, Transaction txTo, int nIn)` 

instead of doing optional parameters? Optional parameters are ok, but it has a slight chance to break other people code at runtime for code depending on a previous version of NBitcoin.

NBitcoin/StandardScriptTemplate.cs
```diff
@@ -317,6 +317,11 @@ public TransactionSignature[] GetMultisigSignatures()
 		{
 			return PayToMultiSigTemplate.Instance.ExtractScriptSigParameters(new Script(Pushes.Select(p => Op.GetPushOp(p)).ToArray()));
 		}
+
+		public PubKey[] GetMultisigPubKeys()
+		{
+			return PayToMultiSigTemplate.Instance.ExtractScriptPubKeyParameters(RedeemScript).PubKeys;
```
NicolasDorier: `return PayToMultiSigTemplate.Instance.ExtractScriptPubKeyParameters(RedeemScript)?.PubKeys;`

NicolasDorier:
Minor nits, you want I wait for RPC before merging this?

NBitcoin/BIP32/KeyPath.cs
NicolasDorier: netstandard2.0

joemphilips:
fixed nits.
> Minor nits, you want I wait for RPC before merging this?

I guess this PR is already too big so I want you to merge it. I will make a separate PR for the RPC when it is necessary.

NicolasDorier:
That's awesome work! I will think about making this class more easy to use. Merging it for now!

NicolasDorier:
Waiting for your RPC tests as well

PR #598, opened 2018-11-23
Rudimentary Pedersen commitment data structure

zeptin:
This is mostly a wrapper around `ECPoint` to make it public, with some helper code to ease working with the commitments.

I intend to add code for the CT range proofs in the future when it's ready.

Reference for value of H generator point:
https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/secp256k1/src/modules/rangeproof/main_impl.h

### DIFF
--- a/NBitcoin/Crypto/PedersenCommitment.cs
+++ b/NBitcoin/Crypto/PedersenCommitment.cs
@@ -1,0 +1,85 @@
+﻿using NBitcoin.BouncyCastle.Asn1.X9;
+using NBitcoin.BouncyCastle.Math;
+using NBitcoin.BouncyCastle.Math.EC;
+
+namespace NBitcoin.Crypto
+{
+	/// <summary>
+	/// A data structure for holding and manipulating Pedersen commitments,
+	/// as used in Confidential Transactions etc.
+	/// </summary>
+	public class PedersenCommitment
+	{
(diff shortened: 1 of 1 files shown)


lontivero:
This needs UTs

NBitcoin/Crypto/PedersenCommitment.cs
```diff
@@ -0,0 +1,85 @@
...
+		// The actual commitment, which is simply a point on the secp256k1 curve.
+		private readonly ECPoint Commitment;
+
+		public PedersenCommitment(BigInteger x, BigInteger a)
+		{
+			this.x = x;
+			this.a = a;
+
+			// Pedersen commitment C = xG + aH
+			this.Commitment = G.Multiply(x).Add(H.Multiply(a));
```
NicolasDorier: I think you need to Modulo it to N. It is normally not mentioned as it suppose it is a group operation in papers.

lontivero: He is adding two point in the curve, no modular arithmetic.

lontivero:
Close this PR @NicolasDorier. There is another one with the same commits and some few extra goodnesses. If I have time I will create a NBitcoin.Experimental with this PR and also some really crazy ideas that could be useful some day like the key aggregation scheme over ecdsa.

PR #600, opened 2018-11-26
Fix bug in `PaytoWitPubkeyHashTemplate.ExtractWitScriptParamters`

joemphilips:
Currently, If a signature is empty in a p2wpkh witness, `ExtractWitScriptParameters` returns null.
This is because in p2wpkh case, "empty signature" is represented by "empty witness item", not by push of 0.
So fix the logic for determining an empty signature.

### DIFF
--- a/NBitcoin.Tests/script_tests.cs
+++ b/NBitcoin.Tests/script_tests.cs
@@ -1014,6 +1014,11 @@ namespace NBitcoin.Tests
 			});
 			Assert.Equal(expected, actual);
 
+			var witScriptWithoutSig = PayToWitPubKeyHashTemplate.Instance.GenerateWitScript(null, pubkey);
+			var actualParam = PayToWitPubKeyHashTemplate.Instance.ExtractWitScriptParameters(witScriptWithoutSig);
+			Assert.NotNull(actualParam);
+			Assert.Equal(pubkey, actualParam.PublicKey);
+
 			var scriptSig = new Script("304402206b782f095f52f12133a96c078b558458b84c925afdb620d96c5f5bbf483e28d502206206796ff45d80216b83c77bafc4e7951fdb10a5bf3e4041c0e6c0938079b22b01 2103");
 			var redeem = new Script(Encoders.Hex.DecodeData("2103a65786c1a48d4167aca08cf6eb8eed081e13f45c02dc6000fd8f3bb16242579aac"));
 			expected = new WitScript(new Script("304402206b782f095f52f12133a96c078b558458b84c925afdb620d96c5f5bbf483e28d502206206796ff45d80216b83c77bafc4e7951fdb10a5bf3e4041c0e6c0938079b22b01 2103 2103a65786c1a48d4167aca08cf6eb8eed081e13f45c02dc6000fd8f3bb16242579aac"));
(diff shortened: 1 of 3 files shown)


NicolasDorier:
This is tricky. Can you add a test that

```csharp
var expected = new WitScript("0 03a65786c1a48d4167aca08cf6eb8eed081e13f45c02dc6000fd8f3bb16242579a");
```

returns a `null` signature?

joemphilips:
That will cause `System.FormatException` in `WitScript` constructor. Because first 0 can not be parsed as hex string.
https://github.com/MetacoSA/NBitcoin/blob/e4326c981428c3d28823d1f9f39f48e25e8531b9/NBitcoin/Transaction.cs#L997

It could be parsed if there are two zeros.
`"00 03a65786c1a48d4167aca08cf6eb8eed081e13f45c02dc6000fd8f3bb16242579a"`

And this will result to a push of `0x00` in a first witness item. (thus in a form that old `ExtractWitScriptParams` expects.)

I can add special case in WitScript constructor so that "0" could be recognized as "empty item" if you prefer.
(I don't understand what's the point of being able to instantiate signature-empty WitScript like this, so I think the current code is fine though.)

NicolasDorier:
```csharp
new Script("0 03a65786c1a48d4167aca08cf6eb8eed081e13f45c02dc6000fd8f3bb16242579a").ToWitScript();
```

joemphilips:
done

NBitcoin/Transaction.cs
```diff
@@ -994,7 +994,14 @@ public WitScript(string script)
 			_Pushes = new byte[parts.Length][];
 			for (int i = 0; i < parts.Length; i++)
 			{
-				_Pushes[i] = Encoders.Hex.DecodeData(parts[i]);
+				if (parts[i] == "0")
+				{
+					_Pushes[i] = new byte[0];
```
NicolasDorier: Don't do this. the `0` syntax is just short for `OP_0` which is only valid for a `Script`, not a `witScript`, because witScript should not have any script instructions.

NBitcoin.Tests/script_tests.cs
```diff
@@ -1014,6 +1014,18 @@ public void CanParseAndGenerateSegwitScripts()
 			});
 			Assert.Equal(expected, actual);
 
+			// should be able to parse if signature is empty in p2wpkh.
+			var witScriptWithoutSig = PayToWitPubKeyHashTemplate.Instance.GenerateWitScript(null, pubkey);
+			var actualParam = PayToWitPubKeyHashTemplate.Instance.ExtractWitScriptParameters(witScriptWithoutSig);
+			Assert.NotNull(actualParam);
+			Assert.Equal(pubkey, actualParam.PublicKey);
+			var script = new Script("0 03a65786c1a48d4167aca08cf6eb8eed081e13f45c02dc6000fd8f3bb16242579a").ToWitScript();
+			var witScript = new WitScript("0 03a65786c1a48d4167aca08cf6eb8eed081e13f45c02dc6000fd8f3bb16242579a");
```
NicolasDorier: this should not work

joemphilips:
Sorry I have misunderstood what you've said. fixed.

NicolasDorier:
I confused myself, I though `OP_0` was pushing `0`. But no... it push empty arrray.

PR #601, opened 2018-11-26
Tiny improvements

lontivero:


NicolasDorier:
That is terrible idea. Because `Shutdown` is blocking you are blocking a thread on the threadpool.

`Task.Run(SafeCloseSocketAsync)` would be better with `SafeCloseSocketAsync` using Async version of `Shutdown`

PR #602, opened 2018-11-27
Add generatetoaddress RPC

### LINKED ISSUE #572: Add alias to generate RPC with generatetoaddress RPC
If we do it soon enough, maybe most of the ecosystem who updates NBitcoin in time won't feel the pain.  

https://github.com/bitcoin/bitcoin/issues/14299

lontivero:
This PR adds `GenerateToAddress` and `GenerateToAddressAsync` new methods. Also add this to the RPCCapabilities scan and uses the new methods by default when call the soon-to-be-deprecated `Generate` rpc method.

Closes #572

### DIFF
--- a/NBitcoin.Tests/RPCClientTests.cs
+++ b/NBitcoin.Tests/RPCClientTests.cs
@@ -1121,6 +1121,35 @@ namespace NBitcoin.Tests
 			}
 		}
 
+		[Fact]
+		public async Task CanGenerateBlocks()
+		{
+			using (var builder = NodeBuilderEx.Create())
+			{
+				var node = builder.CreateNode();
+				node.CookieAuth = true;
+				node.Start();
+				var rpc = node.CreateRPCClient();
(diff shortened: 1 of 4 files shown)


NicolasDorier:
TIL we have `GetSegwitAddress`. Why does the test would pass? This is not supported right now right?

lontivero:
@NicolasDorier the builds fail for different reasons:

* AppVeyor fails because it seems it cannot connect with QBitNinga (test: `NBitcoin.Tests.RepositoryTests.CanRequestTransactionOnQBit`)

* travis fails but the point is it is not building the latest commit https://github.com/MetacoSA/NBitcoin/pull/602/commits/fc2402c9edc24d3b07b42191c127e3bed27a4c24 but a different one  (look at the commit they are compiling https://github.com/MetacoSA/NBitcoin/commit/5ee795b676e03c8226071e3a426b4e5f456eb630).

PR #611, opened 2018-12-07
Add new GetMempoolEntry RPC method

lontivero:
Do not merge yet please, I want to improve a bit the properties' names

PR #612, opened 2018-12-08
Add new GetMempoolEntry RPC method 2

nopara73:
Replacement of: https://github.com/MetacoSA/NBitcoin/pull/611

I added XML comments and removed an obsolated entry.

### DIFF
--- a/NBitcoin.TestFramework/NodeBuilder.cs
+++ b/NBitcoin.TestFramework/NodeBuilder.cs
@@ -573,7 +573,7 @@ namespace NBitcoin.Tests
 			{
 				batch.SendRawTransactionAsync(tx);
 			}
-			rpc.SendBatch();
+			batch.SendBatch();
 		}
 
 		object l = new object();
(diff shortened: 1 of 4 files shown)


NBitcoin.TestFramework/NodeBuilder.cs
```diff
@@ -573,7 +573,7 @@ public void Broadcast(params Transaction[] transactions)
 			{
 				batch.SendRawTransactionAsync(tx);
 			}
-			rpc.SendBatch();
+			batch.SendBatch();
```
NicolasDorier: wait how it could even work before?

NicolasDorier:
I would `throwIfNotFound` to `false` by default

NBitcoin.TestFramework/NodeBuilder.cs
lontivero: It didn't work before, in fact it threw an exception saying "use PrepareBatch".

NicolasDorier:
shit indeed, keep consistent.

PR #614, opened 2018-12-09
UPDATED: Comments and messages only. Removed all code changes.

iainholder:
Update fixing some spelling and typos. Conservative changes, done mainly by hand. Two biggest single changes are zeroes -> zeros and advertize -> advertise (the only correct spelling in US and GB English). No changes from US->GB or GB->US.

### DIFF
--- a/NBitcoin.Altcoins/BCash.cs
+++ b/NBitcoin.Altcoins/BCash.cs
@@ -817,7 +817,7 @@ namespace BCashAddr
 		}
 
 		/// <summary>
-		/// Converts an array of 8-bit integers into an array of 5-bit integers, right-padding with zeroes if necessary
+		/// Converts an array of 8-bit integers into an array of 5-bit integers, right-padding with zeros if necessary
 		/// </summary>
 		/// <param name="data"></param>
 		/// <returns></returns>
@@ -828,7 +828,7 @@ namespace BCashAddr
 
 		/// <summary>
 		/// Converts an array of 5-bit integers back into an array of 8-bit integers
-		/// removing extra zeroes left from padding if necessary.
+		/// removing extra zeros left from padding if necessary.
 		/// Throws a ValidationError if input is not a zero-padded array of 8-bit integers
 		/// </summary>
 		/// <param name="data"></param>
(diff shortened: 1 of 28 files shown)


knocte:
> I would prefer if you could reverse the exception message fixes. Those can break applications.

I disagree with this, if someone is parsing exception messages they're doing it wrong and their application should break.

NicolasDorier:
Yes, parsing messages is wrong. Also the errors he fixed are about exception happening only when your code has an obvious bug. (`ArgumentException`)

nopara73:
"If a change results in user programs breaking, it's a bug in the kernel. We never EVER blame the user programs."

https://lkml.org/lkml/2012/12/23/75

knocte:
I don't think that quote can be framed properly when someone changes a typo on an error message. The Linux kernel is written in C and probably all error codes are integer in that case...

nopara73:
I think it's more important in Bitcoin. We're playing with people's money after all.

nopara73:
I'm fine with breaking stuff only when Visual Studio complains. Of course this is a .net privilege, because linux gurus' command line cist editors won't tell them about anything. But let's not overdo it.

knocte:
Parsing error messages is simply wrong. If you lose money because you use crappy software, it's the user's problem :) .NET has an exceptions model so if any library wants to consume errors and do different things depending on the error, they will need to make sure that the library they consume has different exception types for each error condition. E.g.: some months ago I found that key decription in Nethereum threw a generic "Exception" type message; instead of checking the exception's message from my wallet, what I did was a PR to Nethereum to change the exception being thrown by a new exception type just created for this purpose. This is how error handling works in .NET, if you do it differently you're doing it wrong ;-)

nopara73:
@knocte It doesn't matter what is the best practice in consuming an API. What matters is what is the best practice in writing an API. Not changing the user space unless it is unavoidable is the best practice. We did not follow this best practice today, so what we did is "simply wrong."

knocte:
I didn't explain best practices of consuming an API, I explained how error handling works in .NET, not a best practice, but the guideline. Anyway upgrading a dependency might always break things unless you have tests so if someone upgrades without testing or having tests, their workflow is also "simply wrong".

NicolasDorier:
NBitcoin is mainly stable, I rarely break anything and what I break get ample time to migrate via `[Obsolete]` with proper migration tips that I remove after one or two years.

The exception message, while technically breaking, is so unlikely to happen that there is no need to bother about it. Especially that those errors are not runtime exception, those are error for invalid parameters. If your code is firing `ArgumentException` you already have a bug.

PR #623, opened 2018-12-24
Update PSBT, More strict sanity check, and rpc test, delete useless PropertyTest

joemphilips:
Mostly done, but I'm having weird bug in `WalletProcessPSBT` rpc.


joemphilips:
Ready for review

NicolasDorier:
@lontivero @nopara73 , you will dog food PSBT once a new package is published with this?

I don't have time to see if the class design is good or not, so I think after some dog fooding, we will end up modifying and breaking some code. Better sooner than later.

NicolasDorier:
@joemphilips last changes:

* Rename Psbt member to PSBT (because it is abreviation it is OK)
* In the RPC client (like the `decodepsbt`), call the real RPC commands, I understand that it does exactly the same thing as what you do in `PSBT.Parse`, but people expect those methods to use RPC under the hood. I think you can comment the method saying that this is the same result as `PSBT.Parse` is ok. If you prefer you can as well remove them completely as they are quite useless for NBitcoin users.

NBitcoin.Tests/PSBTTests.cs
```diff
@@ -148,24 +152,35 @@ public void CanUpdate()
...
 
-			var SignedPSBTWithTXs = PSBTWithTXs
-				.TrySignAll(keys);
+			// Operation must be idempotent.
+			tmp = PSBTWithTXs.Clone()
+				.AddCoins(coins)
+				.AddTransactions(funds);
+			Assert.Equal(PSBTWithTXs, tmp, ComparerInstance);
+
+			var ClonedPSBT = PSBTWithTXs.Clone();
```
knocte: local variables should be camelCase, not PascalCase

NBitcoin.Tests/PSBTTests.cs
```diff
@@ -148,24 +152,35 @@ public void CanUpdate()
...
+
+			ClonedPSBT.TrySignAll(keys[0]);
+			PSBTWithTXs.TrySignAll(keys[1], keys[2]);
+
+			var WhollySignedPSBT = ClonedPSBT.Combine(PSBTWithTXs);
```
knocte: same here

NBitcoin.Tests/testresult
```diff
@@ -0,0 +1,18 @@
+.NET Core 向け Microsoft (R) Build Engine バージョン 15.9.20+g88f5fadfbe
```
knocte: missing .gitignore entry?

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -858,6 +861,21 @@ public async Task<SignRawTransactionResponse> SignRawTransactionWithWalletAsync(
...
 
+		public WalletProcessPSBTResponse WalletProcessPSBT(PSBT psbt, bool sign = true, SigHash hashType = SigHash.All, bool bip32derivs = false)
+			 => WalletProcessPSBTAsync(psbt, sign, hashType, bip32derivs).GetAwaiter().GetResult();
+		public async Task<WalletProcessPSBTResponse> WalletProcessPSBTAsync(PSBT psbt, bool sign = true, SigHash sighashType = SigHash.All, bool bip32derivs = false)
+		{
+			if (psbt == null)
+				throw new ArgumentNullException(nameof(psbt));
+
+			var response = await SendCommandAsync(RPCOperations.walletprocesspsbt, psbt.ToBase64(), sign, SigHashToString(sighashType), bip32derivs ).ConfigureAwait(false);
+			var result = response.Result as JObject;
```
knocte: if you're not comparing "result" against null later, you should use a normal (static) cast instead of the keyword "as" (dynamic cast)

joemphilips:
I have removed all dummy RPCs.

NicolasDorier:
4.1.1.74 is pushed with PSBT!

lontivero:
> @lontivero @nopara73 , you will dog food PSBT once a new package is published with this?

No by the moment, @NicolasDorier.

PR #625, opened 2018-12-28
Rename Blinding signature clases and improves API

lontivero:
This PR comes after dogfooding it in Wasabi Wallet. It does:

* Rename the whole system from **ECDSABlinding** to **SchnorrBlinding**.
* Rename **BlindSignature** to **UnblindedSignature** because it is the signature after being unblinded
* Uses **uint256** in API interface instead of **BigIntegers** because the numbers are always less than 2^256
* Simplifies the API thanks to new methods:
```c#
var message = Encoders.ASCII.DecodeData("Hello, World!");
var requester = new SchnorrBlinding.Requester();
var signer = new SchnorrBlinding.Signer(new Key());
var blindedMessage = requester.BlindMessage(message, signer.R.PubKey, signer.Key.PubKey);
var blindSignature = signer.Sign(blindedMessage);
var unblindedSignature = requester.UnblindSignature(blindSignature);

Assert.True( signer.VerifyUnblindedSignature(unblindedSignature, message) );
```

* Tests: I added a brute force test that runs in about a second (I think it will not delay the tests). I also ran that test in a 10_000_000 loop that took a couple of hours and it past green.

### DIFF
--- a/NBitcoin.Tests/DeterministicSignatureTests.cs
+++ b/NBitcoin.Tests/DeterministicSignatureTests.cs
@@ -14,6 +14,7 @@ using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Xunit;
 using NBitcoin.BouncyCastle.Asn1;
+using System.Security;
 
 namespace NBitcoin.Tests
 {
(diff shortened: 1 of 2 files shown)


knocte:
PR has mixed tabs and spaces

lontivero:
I don't see where I could mix tabs and spaces.

lontivero:
Oh I see what you mean. I used spaces for the file that was using spaces and tabs for the file that was using tabs. The thing is that I cannot replace the tabs by spaces or the other way neither because in that case the reviewer will not be able to see what I changed easily.

We can make another PR with those, and other similar, changes but I prefer to keep this PR focused.

PR #628, opened 2018-12-31
Pedersen commitment

lontivero:
Continues https://github.com/MetacoSA/NBitcoin/pull/598

### DIFF
--- a/NBitcoin.Tests/PedersenCommitmentTests.cs
+++ b/NBitcoin.Tests/PedersenCommitmentTests.cs
@@ -1,0 +1,41 @@
+using Xunit;
+using NBitcoin;
+using NBitcoin.Crypto;
+using NBitcoin.BouncyCastle.Math;
+
+namespace NBitcoin.Tests
+{
+	public class PedersenCommitmentTests
+	{
+		[Fact]
+		[Trait("UnitTest", "UnitTest")]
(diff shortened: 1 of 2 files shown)


zeptin:
Thanks @lontivero, I have been too busy to carry on with this.

NicolasDorier:
Is there any test vector on other projects with pedersen commitment done by some crypto gurus like sipa?

lontivero:
I don't know. As I wrote in my commit's description this is not a production ready implementation because there are no test vectors, and there are some cases that are not controlled like negative value numbers  and provably corner cases that I ignore.

I just reviewed that `H` was correct (it is equal to `H = Secp256k1.Curve.DecodePoint(new[] { (byte)0x02 }.Concat(Hashes.SHA256(G.GetEncoded(false))));`), added a few important validations, added some methods/operations and wrote a test to demonstrate how this works (documentation).

IMO the PR is useful for playing with Pedersen commitments and working on proof of concepts because the math looks correct.

NicolasDorier:
Closing this with my port of Secp256k1, you can port the pedersen commitment directly from https://github.com/ElementsProject/secp256k1-zkp with all the tests.

I did not ported pedersen commitment, but this should be easy enough if you intend to do.

zeptin:
Thanks

PR #631, opened 2019-01-04
PSBT scriptcoin test

joemphilips:
This is part of #627 and It depends on #630 .

Behavior for Adding `ScriptCoin` was not ideal, so Improved it with a test.

PR #632, opened 2019-01-04
Add two RPCs `getaddressinfo` and `walletcreatefundedpsbt`

joemphilips:
This is part of #627 , and it depends on #630 and #631 

Both tests depend on `importmulti` , but this does not support witness script in v0.17.
So ignore the case for p2wsh and p2sh-p2wsh.
maybe it is good to not merge until master bumps to v0.18 :(

### DIFF
--- a/NBitcoin.Tests/RPCClientTests.cs
+++ b/NBitcoin.Tests/RPCClientTests.cs
@@ -1434,6 +1434,8 @@ namespace NBitcoin.Tests
 		{
 			using (var builder = NodeBuilderEx.Create())
 			{
+				if (!builder.NodeImplementation.Version.Contains("0.17"))
+					throw new Exception("Test must be updated!");
 				var nodeAlice = builder.CreateNode();
 				var nodeBob = builder.CreateNode();
 				var nodeCarol = builder.CreateNode();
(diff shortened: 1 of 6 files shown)


NicolasDorier:
Need rebase, and fix the build

NBitcoin/BIP174/PartiallySignedTransaction.cs
```diff
@@ -1493,25 +1512,36 @@ public PSBT CoinJoin(PSBT other)
 			return result;
 		}
 
-		public PSBT TryFinalize(out bool result)
+		/// <summary>
+		/// If this method throws an error, that is a bug.
+		/// </summary>
+		/// <param name="errors"></param>
+		/// <returns></returns>
+		public PSBT TryFinalize(out InvalidOperationException[] errors)
```
knocte: TryFoo() methods in C# usually work the other way around: they return bool, and have an out parameter for the thing to return

NicolasDorier: Good point

joemphilips: Will create another PR for this.

joemphilips:
Added assertion for bitcoind version. 
I've also created an another PR depends on this. #633 which renames public methods as @knocte pointed out.

NicolasDorier:
next time for fixing the nits of a PR just push more commit to your branch instead of a new PR :p

PR #633, opened 2019-01-05
Rename PSBT's public methods to not use a word "Try"

joemphilips:
Since it violates C# convention.
Some internal methods still use it but I suppose this doesn't matter.

Depends on #632


NicolasDorier:
need rebase

NicolasDorier:
You are done with your changes or should I wait to release a new version?

PR #638, opened 2019-01-10
Money: warn on ctors with too limited types

knocte:
Devs using these constructors should know that their systems
are buggy! Let's be friendly and give them a warning.


NicolasDorier:
good point

PR #639, opened 2019-01-14
add SendAll method to TransactionBuilder

joemphilips:
For sweeping the funds in the wallet.
(Not very sure this is the best way to implement because I don't get the whole picture of TransactionBuilder. Feedback is appreciated)

### DIFF
--- a/NBitcoin.Tests/transaction_tests.cs
+++ b/NBitcoin.Tests/transaction_tests.cs
@@ -3427,13 +3427,17 @@ namespace NBitcoin.Tests
 				.AddCoins(RandomCoin(Money.Coins(0.5m), k))
 				.AddCoins(RandomCoin(Money.Coins(0.5m), k))
 				.AddKeys(k)
-				.SendAll(dest, rate);
+				.SendAll(dest);
+			
+			var fee = builder.EstimateFees(rate);
+			builder.SendFees(fee);
 
 			var tx = builder.BuildTransaction(true);
 			if (!builder.Verify(tx, out var errors))
 				throw new AggregateException(errors.Select(e => new Exception(e.ToString())));
 			Assert.Equal(2, tx.Inputs.Count);
 			Assert.Equal(1, tx.Outputs.Count);
+			Assert.Equal(Money.Coins(1.0m), fee + tx.Outputs[0].Value);
 		}
 	}
 }
--- a/NBitcoin.Tests/transaction_tests.cs
+++ b/NBitcoin.Tests/transaction_tests.cs
@@ -3412,5 +3412,28 @@ namespace NBitcoin.Tests
 		{
 			return Encoders.Hex.DecodeData(data);
 		}
+
+		[Fact]
+		public void ShouldSendAll()
+		{
+			var builder = Network
+				.CreateTransactionBuilder();
+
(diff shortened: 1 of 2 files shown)


NBitcoin/TransactionBuilder.cs
```diff
@@ -700,6 +702,24 @@ public TransactionBuilder Send(IDestination destination, Money amount)
...
 
+		/// <summary>
+		/// Send all coins added so far with no change. It will automatically estimate the fee.
+		/// </summary>
+		/// <param name="destination"></param>
+		/// <returns></returns>
+		public TransactionBuilder SendAll(IDestination destination, FeeRate rate)
+			=> SendAll(destination.ScriptPubKey, rate);
+
+		public TransactionBuilder SendAll(Script scriptPubKey, FeeRate rate)
```
knocte: I think it's better to make SendAll() not receive FeeRate param (in the same way that other Send methods don't receive it). You can call EstimateFee() from the test after SendAll(), and use the result of EstimateFee() with SendFees() after that.

joemphilips: You are right. I once thought that doesn't work. But it did. (Maybe because I was not aware of how to use `SubtractFee()` method)

NBitcoin.Tests/transaction_tests.cs
```diff
@@ -3412,5 +3412,31 @@ private byte[] ParseHex(string data)
...
+				.SendAll(dest);
+			
+			var fee = builder.EstimateFees(rate);
+			builder.SendFees(fee);
+
+			var tx = builder.BuildTransaction(true);
+			if (!builder.Verify(tx, out var errors))
+				throw new AggregateException(errors.Select(e => new Exception(e.ToString())));
+			Assert.Equal(2, tx.Inputs.Count);
+			Assert.Equal(1, tx.Outputs.Count);
```
knocte: looks great now! maybe just add one more assert to the test? checking that amount of output + estimatedFee equals amount of inputs

NBitcoin.Tests/transaction_tests.cs
```diff
@@ -3412,5 +3412,31 @@ private byte[] ParseHex(string data)
...
+
+			var rate = new FeeRate(Money.Coins(0.0004m));
+			builder
+				.AddCoins(RandomCoin(Money.Coins(0.5m), k))
+				.AddCoins(RandomCoin(Money.Coins(0.5m), k))
+				.AddKeys(k)
+				.SendAll(dest);
+			
+			var fee = builder.EstimateFees(rate);
+			builder.SendFees(fee);
```
NicolasDorier: Why no `SendEstimatedFees`?

NBitcoin.Tests/transaction_tests.cs
NicolasDorier: Yes
```csharp
Assert.Equal(Money.Coins(1.0m), fee + tx.Outputs.TxOut.Value);
```

NBitcoin.Tests/transaction_tests.cs
joemphilips: Originally just because I wasn't aware of the api. Now because `fee` is required for assertion.

PR #640, opened 2019-01-16
Nits and bump

knocte:
geewallet needs the recently introduced SendAll API,
so I include a bump here so that we can have a new
nuget package. Thanks!


NicolasDorier:
New version out

knocte:
thanks!

Issue #607, opened 2018-12-04
Dash DIP 3 testnet


pozdnyakovasp:
Hi.

I try work with Dash core testnet (13.0.0 cr6)

```
RpcClient.GetBlock(273999);  - OK

// After block 273999 i get exception
RpcClient.GetBlock(274000); next exception
```
 

>   в NBitcoin.BitcoinStream.ReadWriteBytes(Byte[]& data, Int32 offset, Int32 count)
>    в NBitcoin.BitcoinStream.ReadWriteNumber(UInt64& value, Int32 size)
>    в NBitcoin.BitcoinStream.ReadWrite(UInt32& data)
>    в NBitcoin.TxIn.ReadWrite(BitcoinStream stream)
>    в NBitcoin.BitcoinStream.ReadWrite[T](T& data)
>    в NBitcoin.BitcoinStream.ReadWriteArray[T](T[]& data)
>    в NBitcoin.BitcoinStream.ReadWriteList[TList,TItem](TList& data)
>    в NBitcoin.Transaction.ReadWrite(BitcoinStream stream)
>    в NBitcoin.BitcoinStream.ReadWrite[T](T& data)
>    в NBitcoin.BitcoinStream.ReadWriteArray[T](T[]& data)
>    в NBitcoin.BitcoinStream.ReadWriteList[TList,TItem](TList& data)
>    в NBitcoin.Block.ReadWrite(BitcoinStream stream)
>    в NBitcoin.BitcoinSerializableExtensions.ReadWrite(IBitcoinSerializable serializable, Stream stream, Boolean serializing, ConsensusFactory consensusFactory, Nullable`1 version)
>    в NBitcoin.Block.Parse(String hex, ConsensusFactory consensusFactory)
>    в NBitcoin.Block.Parse(String hex, Network network)
>    в NBitcoin.RPC.RPCClient.<GetBlockAsync>d__87.MoveNext()
> --- Конец трассировка стека из предыдущего расположения, где возникло исключение ---
>    в System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    в System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    в NBitcoin.RPC.RPCClient.<GetBlockAsync>d__90.MoveNext()
> --- Конец трассировка стека из предыдущего расположения, где возникло исключение ---
[...]

PR #645, opened 2019-01-23
Dash 0.13 Special Transactions support

BenjaminNitschke:
All the fixes for https://github.com/MetacoSA/NBitcoin/issues/607 and support to load all new Special Transaction types in Dash 0.13:
1	Provider Registration Transaction (ProRegTx),
2	Provider Update Service Transaction (ProUpServTx),
3	Provider Update Registrar Transaction (ProUpRegTx),
4	Provider Update Revocation Transaction (ProUpRevTx),
5	Coinbase Transaction (CbTx),
6	Quorum Commitment (QcTx)
Details in the source code, especially the new DashTests.cs

### DIFF
--- a/NBitcoin.Altcoins/Dash.cs
+++ b/NBitcoin.Altcoins/Dash.cs
@@ -5,10 +5,12 @@ using NBitcoin.Protocol;
 using NBitcoin.RPC;
 using System;
 using System.Collections.Generic;
+using System.Data.SqlTypes;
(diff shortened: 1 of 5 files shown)


BenjaminNitschke:
if the continous-integration stuff fails because of the manual test ConnectToDashTestnetAndCheckBlock7000, it can be removed (the [Fact] attribute is commented out to not run by default).

PR #646, opened 2019-01-28
Encrypt/Decrypt message

lontivero:
This PR is for allowing encrypt and decrypt messages with eckey.

Missing parts:
* More tests
* Test compatibility with electrum

### DIFF
--- a/NBitcoin/PubKey.cs
+++ b/NBitcoin/PubKey.cs
@@ -484,13 +484,8 @@ namespace NBitcoin
 			var encryptionKey = sharedKey.SafeSubarray(16, 16);
 			var hashingKey = sharedKey.SafeSubarray(32);
 
-			var aes = AES.Create();
-			aes.KeySize = 128;
-			aes.Key = encryptionKey;
-			aes.Mode = System.Security.Cryptography.CipherMode.CBC;
-			aes.IV = iv;
-			var encryptor = aes.CreateEncryptor();
-			var cipherText = encryptor.TransformFinalBlock(message, 0, message.Length);
+			var aes = new AesBuilder().SetKey(encryptionKey).SetIv(iv).IsUsedForEncryption(true).Build();
+			var cipherText = aes.Process(message, 0, message.Length);
 			var ephemeralPubkeyBytes = ephemeral.PubKey.ToBytes();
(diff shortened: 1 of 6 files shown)


NBitcoin/PubKey.cs
```diff
@@ -465,6 +465,33 @@ public PubKey GetSharedPubkey(Key key)
...
+			if(string.IsNullOrEmpty(message))
+				throw new ArgumentNullException(nameof(message));
+
+			var bytes = Encoding.UTF8.GetBytes(message);
+			return Encoders.Base64.EncodeData(Encrypt(bytes));
+		}
+
+		public byte[] Encrypt(byte[] message)
+		{
+			if(message is null)
```
knocte: it's the first time I see a null check using `is` instead of `==`, does this work?

NicolasDorier: Actually I read good reasons to use it and why it is better and more performant in some cases.
I think it is because it does not have to check with overload operators or stuff like this... trying to find the article.

lontivero: yes. https://stackoverflow.com/questions/53184991/c-sharp-how-to-check-for-null-value-is-null-or-null-value-can-we-use-is

This PR is ready for review @NicolasDorier @nopara73. I think more tests are not necessary and I verified it is compatible with Electrum wallet.

NicolasDorier: Actually now I am seeing article which say it is slower. (not that it really matter in this specific case though)

knocte: @NicolasDorier slower? lontivero's link says the opposite

NicolasDorier:
image
wtf, why does it do that on my machine

NicolasDorier:
Checking this no worry.

NicolasDorier:
Fixed with https://github.com/MetacoSA/NBitcoin/commit/83d0ad5bae1fff712132fe6d1273a1246eeb621d

NicolasDorier:
Will release new version as soon as tests pass.

NicolasDorier:
I need to fix https://travis-ci.org/MetacoSA/NBitcoin/jobs/485725812 I will do a bit later.
This is strange, I can't replicate locally :(

NicolasDorier:
Released `4.1.1.80`! :)

PR #654, opened 2019-02-03
Easy perf improvement for GolomRiceFilters

NicolasDorier:
Ping @lontivero some low hanging fruits for better perfs of golomb filters.
Should relief the GC a bit.

### DIFF
--- a/NBitcoin/BIP158/GolombRiceFilter.cs
+++ b/NBitcoin/BIP158/GolombRiceFilter.cs
@@ -108,13 +108,14 @@ namespace NBitcoin
 		/// <param name="key">Key used for hashing the datalements.</param>
 		/// <param name="data">Data elements to be computed in the list.</param>
 		/// <returns></returns>
-		internal static List<ulong> ConstructHashedSet(byte P, int n, uint m, byte[] key, IEnumerable<byte[]> data)
+		internal static ulong[] ConstructHashedSet(byte P, int n, uint m, byte[] key, IEnumerable<byte[]> data)
 		{
 			// N the number of items to be inserted into the set.
 			var dataArrayBytes = data as byte[][] ?? data.ToArray();
 
 			// The list of data item hashes.
-			var values = new List<ulong>();
+			var values = new ulong[dataArrayBytes.Length];
+			var valuesIndex = 0;
 			var modP = 1UL << P;
 			var modNP = ((ulong)n) * m;
 			var nphi = modNP >> 32;
@@ -128,10 +129,10 @@ namespace NBitcoin
 			{
 				var hash = SipHash(k0, k1, item);
 				var value = FastReduction(hash, nphi, nplo);
-				values.Add(value);
+				values[valuesIndex++] = value;
 			}
 
-			values.Sort();
+			Array.Sort(values);
 			return values;
 		}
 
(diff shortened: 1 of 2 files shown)


NBitcoin/BIP158/GolombRiceFilter.cs
```diff
@@ -481,7 +483,7 @@ public GolombRiceFilter Build()
 			return new GolombRiceFilter(filterData, n, _p, _m);
 		}
 
-		private static byte[] Compress(List<ulong> values, byte P)
+		private static byte[] Compress(ulong[] values, byte P)
```
knocte: here you could use IEnumerable?

NicolasDorier: meh, it is private method only used here, and a performance sensitive part. Not really worth it.

NBitcoin/BIP158/GolombRiceFilter.cs
```diff
@@ -398,7 +399,8 @@ public GolombRiceFilterBuilder AddScriptPubkey(Script scriptPubkey)
 			if (scriptPubkey == null)
 				throw new ArgumentNullException(nameof(scriptPubkey));
 
-			_values.Add(scriptPubkey.ToBytes());
+			// Unsafe is OK because Script is readonly and we do not modify the arrays inside values
+			_values.Add(scriptPubkey.ToBytes(true));
```
lontivero: Yes! this has to be good.

lontivero:
I've review the commits and they look very good. Currently operating with compact filter in NBitcoin is quite fast, and will be faster with these changes. However, cf requires a lot of bit level operations and that's where processing 200GB bit by bit can be a slow. That's something could be improved but it is not causing any problem in Wasabi, in fact I think it will never become a problem.

The last time I measured them I got the numbers below. Take into account that in Wasabi those times has to be necessarily better because we only use p2wpkh scripts what makes the filter much smaller.

```
Filters generation time: 00:00:02.9757793
Filters matching time: 00:00:00.4871420
Filters matching false time: 00:00:01.3413390
```

However, after that there were an improvement in the `SipHash` that is called millions of times (and here @NicolasDorier made it a struct :+1: ) so, those numbers have to be even better.

NBitcoin/BIP158/GolombRiceFilter.cs
```diff
@@ -155,7 +154,18 @@ public uint256 GetHeader(uint256 previousHeader)
...
+
+		/// <summary>
+		/// Checks if any of the provided elements is in the filter.
+		/// </summary>
+		/// <param name="data">Data elements to check in the filter.</param>
+		/// <param name="key">Key used for hashing the data elements.</param>
+		/// <returns>true if at least one of the elements is in the filter, otherwise false.</returns>
+		public bool MatchAny(byte[][] data, byte[] key)
+		{
+			return MatchAny(data, data.Length, key);
```
lontivero: Possible NRE in `data.Length`

NBitcoin/BIP158/GolombRiceFilter.cs
```diff
@@ -166,10 +176,32 @@ public bool Match(byte[] data, byte[] key)
...
+			{
+				return MatchAny(dataArray, dataArray.Length, key);
+			}
+			else if (data is ICollection<byte[]> dataCollection)
+			{
+				return MatchAny(dataCollection, dataCollection.Count, key);
+			}
+			else
+			{
+				return MatchAny(data, data.Count(), key);
```
lontivero: Possible NRE.

lontivero:
ACK. I review it again and don't see anything that can break. Just the two possible NRE that were not introduced in this refactoring it seems.

PR #656, opened 2019-02-09
NBitcoin.Bench project

lontivero:
This PR adds a new project for bechmarking with `BenchmarkDotNet` into NBitcoin solution. It is used for measuring `GolombRiceFilterBuilder` and `GolomRiceFilter`s in the context of the discussion in Wasabi Wallet filter creation https://github.com/zkSNACKs/WalletWasabi/issues/1134

Below you can see the running results for `GolombRiceFilter.Match` and `GolombRiceFilter.MatchAny` methods: 

```
BenchmarkDotNet=v0.11.3, OS=ubuntu 18.04
Intel Core i7-7500U CPU 2.70GHz (Kaby Lake), 1 CPU, 4 logical and 2 physical cores
.NET Core SDK=2.2.100
  [Host] : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT
  Core   : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT

Job=Core  Runtime=Core

   Method |      N |         Mean |      Error |     StdDev | Rank |
--------- |------- |-------------:|-----------:|-----------:|-----:|
    Match |   1000 |     4.954 us |  0.0226 us |  0.0200 us |    1 |
 MatchAny |   1000 |     5.187 us |  0.0477 us |  0.0422 us |    2 |
    Match |  10000 |   249.643 us |  4.8950 us |  5.4407 us |    5 |
 MatchAny |  10000 |    10.870 us |  0.0954 us |  0.0796 us |    3 |
    Match | 100000 | 3,668.975 us | 36.2116 us | 32.1006 us |    6 |
 MatchAny | 100000 |   116.348 us |  2.8049 us |  2.6237 us |    4 |

// * Legends *
  N      : Value of the 'N' parameter
  Mean   : Arithmetic mean of all measurements
  Error  : Half of 99.9% confidence interval
  StdDev : Standard deviation of all measurements
  Rank   : Relative position of current benchmark mean among all benchmarks (Arabic style)
  1 us   : 1 Microsecond (0.000001 sec)
```

This is the result of running the test against `GolombRiceFilterBuilder.Build` method:

```
BenchmarkDotNet=v0.11.3, OS=ubuntu 18.04
Intel Core i7-7500U CPU 2.70GHz (Kaby Lake), 1 CPU, 4 logical and 2 physical cores
.NET Core SDK=2.2.100
  [Host] : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT
  Core   : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT

Job=Core  Runtime=Core

 Method |      N |        Mean |       Error |      StdDev | Rank |
------- |------- |------------:|------------:|------------:|-----:|
  Build |   1000 |    175.5 us |   0.9845 us |   0.8221 us |    1 |
  Build |  20000 |  4,022.0 us |  23.9804 us |  20.0247 us |    2 |
  Build | 100000 | 24,525.1 us | 479.2576 us | 400.2018 us |    3 |

// * Legends *
  N      : Value of the 'N' parameter
  Mean   : Arithmetic mean of all measurements
  Error  : Half of 99.9% confidence interval
  StdDev : Standard deviation of all measurements
  Rank   : Relative position of current benchmark mean among all benchmarks (Arabic style)
  1 us   : 1 Microsecond (0.000001 sec)
```

It seems the `Build` times doesn't increase linearly (sucks).

### DIFF
--- a/NBitcoin.Bench/GolombRiceFilters.cs
+++ b/NBitcoin.Bench/GolombRiceFilters.cs
@@ -1,0 +1,93 @@
+using System;
+using NBitcoin;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Running;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NBitcoin.Benck
+{
+	[ClrJob(baseline: true), CoreJob]
+	[RPlotExporter, RankColumn]
+	public class GolombRiceFilterBuilding
+	{
+		private GolombRiceFilterBuilder _builder;
+		private List<byte[]> _itemsInFilter;
+
+		// We estimate a block will never have more than 100,000 txouts
+		// OTOH 20,000 outputs could be a rasonable number of txouts 
+		// for a block considering the current block sizes.
+		// Also, currently P2WPKH tx are 5% of total scripts and that
+		// gives us 1,000 txouts.
+		[Params(1_000, 20_000, 100_000)]
+		public int N;
+
+		[GlobalSetup]
+		public void Setup()
+		{
+			var random = new Random(Seed:145);
+			var keyBuffer = new byte[32];
+			random.NextBytes(keyBuffer);
+			var key = new uint256(keyBuffer);
+
+			_builder = new GolombRiceFilterBuilder()
+				.SetKey(key)
+				.SetP(20);
+
+			_itemsInFilter = new List<byte[]>();
+			for (var j = 0; j < N; j++)
+			{
+				var data = new byte[random.Next(20, 30)];
+				random.NextBytes(data);
+				_itemsInFilter.Add(data);
+			}
+			_builder.AddEntries(_itemsInFilter);
+		}
+
+		[Benchmark]
+		public GolombRiceFilter Build() => _builder.Build();
+	}
+
+	[ClrJob(baseline: true), CoreJob]
+	[RPlotExporter, RankColumn]
+	public class GolombRiceFilterQuering
+	{
+		private GolombRiceFilter _filter;
+		private byte[][] _sample;
+		private Random _random = new Random(Seed:145);
+		private byte[] _testKey;
+
+		[Params(1_000, 10_000, 100_000)]
+		public int N;
+
+		[GlobalSetup]
+		public void Setup()
+		{
+			var keyBuffer = new byte[32];
+			_random.NextBytes(keyBuffer);
+			var key = new uint256(keyBuffer);
+			_testKey = key.ToBytes().Take(16).ToArray();
+
+			var builder = new GolombRiceFilterBuilder()
+				.SetKey(key)
+				.SetP(20);
+
+			var itemsInFilter = new List<byte[]>();
+			for (var j = 0; j < N; j++)
+			{
+				var data = new byte[_random.Next(20, 30)];
+				_random.NextBytes(data);
+				itemsInFilter.Add(data);
+			}
+			builder.AddEntries(itemsInFilter);
+			_sample = itemsInFilter.OrderBy(x=>_random.Next()).Take(N / 200).ToArray();
+			_filter = builder.Build();
+		}
+
+		[Benchmark]
+		public bool Match() => _filter.Match(_sample[_sample.Count() %2 ], _testKey);
+
+		[Benchmark]
+		public bool MatchAny() => _filter.MatchAny(_sample, _testKey);
+	}
+}
--- a/NBitcoin.Bench/NBitcoin.Bench.csproj
+++ b/NBitcoin.Bench/NBitcoin.Bench.csproj
@@ -1,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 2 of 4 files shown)


lontivero:
The problem with this benchmark system is that without a profiler it is a bit useless. I mean, why is the filters building time grown faster than linearly? To answer that kind of questions we need a profiler.

lontivero:
@NicolasDorier I already tried to improve this and the two methods that take more time are `SipHasher::Write` a `BitStream::WriteBits`

image

NicolasDorier:
@lontivero well, you can run the bench through a profiler like perfview. (which is where your flame graph come from I guess?)

NicolasDorier:
For SipHasher, maybe we can get some perf improvement by aggressive inlining. https://www.dotnetperls.com/aggressiveinlining

NicolasDorier:
You can save on `BitStream` by pre-allocating the right size for the Bitstream which prevent the `Array.Resize` and size check.

NicolasDorier:
You can also save by passing to BitStream an external array. This remove the need to make copies, and save one allocation.

lontivero:
Instead of using _aggressiveinlining_ I expanded the `SipRound` method in this PR https://github.com/MetacoSA/NBitcoin/pull/553. I don't think there is a good place for adding _aggressive inlining_ and the gains are just a few microseconds in total. 

> You can save on `BitStream` by pre-allocating the right size for the Bitstream which prevent the `Array.Resize` and size check.

This is not so easy (I think it is not possible indeed) because you don't know how much space you will need. The `BitStream` is used to keep the bits resulting of the ongoing data compression process and we cannot know in advance how big will be the compressed data before compressing it. I could compress real data to get a very accurate estimation but in our case there are no buffer resizing at all and if there were any we are making the buffer to grow 4k each time.

> You can also save by passing to BitStream an external array for its internal operation. This remove the need to make copies, and save one allocation.

This is true and I will do it when I find time. Anyway, this will not improve the performance. It seems there are no more low hanging fruit.

NicolasDorier:
@lontivero how did you do to use Perfview with Benchmarkdotnet for the flamegraph?

lontivero:
@NicolasDorier I don't use Perfview because my environment is Linux. Anyway, you need to create a `ManualConfig` to let the benchmarking runner know that you want to run the project in debug. Here you have the changes that you need to do:

```
using BenchmarkDotNet.Running;
using BenchmarkDotNet.Configs;
using BenchmarkDotNet.Validators;
using System.Linq;

namespace NBitcoin.Benck
{
	public class Program
	{
		public static void Main(string[] args)
		{
			BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args, new AllowNonOptimized());
		}
	}

	public class AllowNonOptimized : ManualConfig
	{
		public AllowNonOptimized()
		{
			Add(JitOptimizationsValidator.DontFailOnError); // ALLOW NON-OPTIMIZED DLLS
			Add(DefaultConfig.Instance.GetLoggers().ToArray()); // manual config has no loggers by default
			Add(DefaultConfig.Instance.GetExporters().ToArray()); // manual config has no exporters by default
			Add(DefaultConfig.Instance.GetColumnProviders().ToArray()); // manual config has no columns by default			
		}
	}
}
``` 

I think that is enough in you case. But if you want to do it with linux here is how I did it:

```
COMPlus_PerfMapEnabled=1 dotnet ~/GitHub/NBitcoin/NBitcoin.Bench/bin/Debug/netcoreapp2.2/NBitcoin.Bench.dll &
sudo perf record -p $! -g
sudo perf script -f | ~/GitHub/FlameGraph/stackcollapse-perf.pl | ~/GitHub/FlameGraph/flamegraph.pl > flame.svg
sudo python -m SimpleHTTPServer 80 &
firefox http://127.0.0.1:80/
``` 

And there you can pick the flamegraph.

NicolasDorier:
amazing thanks!

NicolasDorier:
@lontivero I benchmark filter building on a 1MB+ block, it takes around `8ms`. I thought you said it took 7 sec on wasabi? oO

lontivero:
@NicolasDorier no, I don't remember saying that. In fact the numbers that I was talking about are in the first comment of this PR, look:

```
 Method |      N |        Mean |       Error |      StdDev | Rank |
------- |------- |------------:|------------:|------------:|-----:|
  Build |   1000 |    175.5 us |   0.9845 us |   0.8221 us |    1 |
  Build |  20000 |  4,022.0 us |  23.9804 us |  20.0247 us |    2 |
  Build | 100000 | 24,525.1 us | 479.2576 us | 400.2018 us |    3 |
```

For a block with 20,000 scriptPubKeys it takes 4 milliseconds and for a 100,000 scriptPubKeys it takes 24.5 milliseconds.

lontivero:
Oh wait, I think I made a mistake in a Wasabi issue where I said 24 seconds instead of 24 millisenconds :(

NicolasDorier:
ah lol, well no need of perf optimization in this case.

NicolasDorier:
I noticed some stuff making serialization slow though thanks to perfview and benchmarkdotnet. 

Funnily, what was taking time was some "clever" code I did in the expectation it would save time... -_-

PR #657, opened 2019-02-10
[QUESTION] Uinit test for bip84

lontivero:
I would like to add support for encoding `ExtKey` with different prefix depending the derivation scheme used. This is something that some people asked for Wasabi wallet but. The problem that I found is that `ExtKey` knows nothing about the concept of derivation scheme, what makes sense, and then it cannot know how to encode itself based on something that it doesn't know about.

I tried a couple of things but they are all ugly. I think the best could be simply overload the `ExtKey.GetWif` method to receive an extra parameter indicating what derivation scheme/prefix to use.

What do you think about this? Which is for you the best way, if any, to implement it?

Note: take a look at the new Base58Type enum elements' name and tell me if they are acceptable.

### DIFF
--- a/NBitcoin.Tests/bip32_tests.cs
+++ b/NBitcoin.Tests/bip32_tests.cs
@@ -276,6 +276,32 @@ namespace NBitcoin.Tests
 			var keyB = new ExtPubKey(keyA.ToBytes());
 			AssertEx.CollectionEquals(keyA.ToBytes(), keyB.ToBytes());
 		}
+
+		[Fact]
+		[Trait("UnitTest", "UnitTest")]
+		public void CanWorkWithSLIP32()
+		{
+			var mnemonic = new Mnemonic("abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about");
+			var extKey = mnemonic.DeriveExtKey();
+			var derivedKey = extKey.Derive(KeyPath.Parse("m/44'/0'/0'")).GetWif(Network.Main);
+			Assert.Equal("xprv9xpXFhFpqdQK3TmytPBqXtGSwS3DLjojFhTGht8gwAAii8py5X6pxeBnQ6ehJiyJ6nDjWGJfZ95WxByFXVkDxHXrqu53WCRGypk2ttuqncb", derivedKey.ToString() );
+			Assert.Equal("xpub6BosfCnifzxcFwrSzQiqu2DBVTshkCXacvNsWGYJVVhhawA7d4R5WSWGFNbi8Aw6ZRc1brxMyWMzG3DSSSSoekkudhUd9yLb6qx39T9nMdj", derivedKey.Neuter().ToString() );
+			derivedKey = extKey.Derive(KeyPath.Parse("m/44'/0'/0'/0/0")).GetWif(Network.Main);
+			Assert.Equal("1LqBGSKuX5yYUonjxT5qGfpUsXKYYWeabA", derivedKey.ScriptPubKey.GetDestinationAddress(Network.Main).ToString());
+
+			derivedKey = extKey.Derive(KeyPath.Parse("m/49'/0'/0'")).GetWif(Network.Main);
+			Assert.Equal("yprvAHwhK6RbpuS3dgCYHM5jc2ZvEKd7Bi61u9FVhYMpgMSuZS613T1xxQeKTffhrHY79hZ5PsskBjcc6C2V7DrnsMsNaGDaWev3GLRQRgV7hxF", derivedKey.ToString() );
+			Assert.Equal("ypub6Ww3ibxVfGzLrAH1PNcjyAWenMTbbAosGNB6VvmSEgytSER9azLDWCxoJwW7Ke7icmizBMXrzBx9979FfaHxHcrArf3zbeJJJUZPf663zsP", derivedKey.Neuter().ToString() );
+			derivedKey = extKey.Derive(KeyPath.Parse("m/49'/0'/0'/0/0")).GetWif(Network.Main);
+			Assert.Equal("37VucYSaXLCAsxYyAPfbSi9eh4iEcbShgf", derivedKey.ScriptPubKey.GetDestinationAddress(Network.Main).ToString());
+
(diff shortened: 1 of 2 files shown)


NicolasDorier:
I think the best way to implement it is to have the concept of derivation scheme inside NBitcoin, completely independent from `ExtPubKey`.

But I am really unsure to do this. It is a pain in the ass, ledger/trezor/electrum are all using different prefix to mean different things.

Also BIP48 is not even implemented anywhere afaik. Hardware wallet use `zpub` but use `m/49'`... This is a mess.

lontivero:
I see. Then this doesn't make a lot of sense. I will implement this in Wasabi to export the zprv/zpub as a helper function what it is enough.

lontivero:
@NicolasDorier today we are working on a feature for displaying this `zprv`/`zpub` prefixes with these extension methods. 

```c#
public static class ExtKeyDataExtensions
{
	public static string ToWif(this ExtPubKey extPubKey, Network network)
	{
		var data = extPubKey.ToBytes();
		var version = (network == Network.Main)
			? new byte[] { (0x04), (0xB2), (0x47), (0x46) }
			: new byte[] { (0x04), (0x5F), (0x1C), (0xF6) };
		
		return Encoders.Base58Check.EncodeData(version.Concat(data).ToArray());
	}

	public static string ToWif(this ExtKey extKey, Network network)
	{
		var data = extKey.ToBytes();
		var version = (network == Network.Main)
			? new byte[] { (0x04), (0xB2), (0x43), (0x0C) }
			: new byte[] { (0x04), (0x5F), (0x18), (0xBC) };
			
		return Encoders.Base58Check.EncodeData(version.Concat(data).ToArray());
	}
}
```

These extension methods would be even more simple if I could add the prefixes to the `Network` instances:
```c#
base58Prefixes[(int)Base58Type.EXT_P2WPKH_PUBLIC_KEY] = new byte[] { (0x04), (0xB2), (0x47), (0x46) }; 
base58Prefixes[(int)Base58Type.EXT_P2WPKH_SECRET_KEY] = new byte[] { (0x04), (0xB2), (0x43), (0x0C) };
``` 

So, the method could be something like this:
```c#
public static string ToWif(this ExtPubKey extPubKey, Network network)
{
	var data = extPubKey.ToBytes();
	var version = network.GetVersionBytes(Base58Type.EXT_P2WPKH_PUBLIC_KEY, true);			
	return Encoders.Base58Check.EncodeData(version.Concat(data).ToArray());
}
```

----

I am not saying this method is okay nor a good design neither. I say that many applications based NBitcoin could need to have access to this prefixes and be able to get wif strings with the `zpub`, `ypub`, `vpub`, `Ypub`, etc. This is becoming a kind of unwritten standard after Electrum, Samurai, Wasabi and other tools like [iancoleman.io](https://iancoleman.io/bip39/) and bitcoinj support them.

I understand that supporting this shit is extremely controversial and anyone can do it by themselves as we do in Wasabi, this is just to know if you agree adding the Base58Type constants and bytes prefix.

NicolasDorier:
Does prefix implies the KeyPath as well?

lontivero:
The idea behind those prefixes is that they contain two pieces of information serialized together: privatekey + derivation scheme. For example, when I give you my zpub extended public key you know that you have to scan for `m/84'/0'` (and that it is for p2wpkh) while when I give you my ypub you know you have to scan for `m/49'/0'` (and that it is for p2wpkh-nested-in-p2sh) and so on.

While it is technically ugly because we should be able to have one extended key for all kind of coins, purpose and scripts type, it is good for usability because you only need to paste your extended key in a wallet text area and it knows what derivation scheme to use for scanning the blockchain. 

Here you have a summary for these prefixes:
https://github.com/satoshilabs/slips/blob/master/slip-0132.md

NicolasDorier:
well btcpay for once, use 49' for p2wpkh

nopara73:
We ended up doing this, so you can choose whichever you want:)

image

lontivero:
> well btcpay for once, use 49' for p2wpkh.

I think the problem is clear then and I am sure we agree with these two statements:
* one extended key (xprv/xpub) should be enough for all kind of coins, purpose and scripts type and,
* the extended key along is not enough for a wallet to know how to derivate children keys (wallets need the derivation scheme too)

Also, I think the following points are not very controversial:
* users shouldn't know nor care about derivation schemes, that should be transparent for them,
* new wif strings (z/y/v/Y/prv) are being adopted because of this reason  (one _magic_ string is all what user needs)

However, the new wif extended keys with different prefixes are ruining what we agree in the first bullet, don't scale and are rigid because they don't support custom derivation schemes, among other compatibility problems.

So, what if we propose some king of structured data format for sharing (xprv/xpub + derivationScheme[])?
Here is my idea:

**Provide a unified format for containing the two pieces of information required for a wallet to be able to scan the blockchain, this is: the extended key and a list of derivation schemes in a compact format.** 

If you take a look at this proposal https:///github.com/satoshilabs/slips/blob/master/slip-0132.md, you will see that most of the derivation paths as  `44'/0'`, `49'/0'`,`84'/0'`, `44'/1'`, `49'/1'`,`84'/1'` and so on. These DPs could be compressed by sure but let's take them as they are for the sake of the example. We could have something like this:

```
xprv9DMUxX4ShgxMLku.......kiwA1S2pEU+T0
```
where the `+` character is a separator and JO is just an example for how we could encode 840.

In case we use the same extended key for different coins or script types we could just add the derivation schemes:
```
xprv9DMUxX4ShgxMLku.......kiwA1S2pEU+T0+Yy+Yz
```
This could be read as: this string contains the extended key and 3 different derivation schemes: 2 for bitcoin 44'/0', 84'/0' and one for litecoin testnet 44'/1' 

In this way a user can use the same _magic_ string with any wallet of any coin. What do you think? Am I missing something important?

nopara73:
Looks like Roger Ver was right, we are eating our own shit now :D

image

nopara73:
I'd stick with xpub for `ToWif()` and go with some functions like `ToYpub()` `ToZpub` and add some xml comments to them that "hey these are non-standard things" and be done with it.

PR #662, opened 2019-02-27
Fixed StackOverflowException when debug in Rider under linux (increas…

oripio:
Fixed StackOverflowException when debug in Rider under linux (increases max stack size)

### DIFF
--- a/NBitcoin/Crypto/Cryptsharp/SCrypt.cs
+++ b/NBitcoin/Crypto/Cryptsharp/SCrypt.cs
@@ -228,7 +228,8 @@ namespace NBitcoin.Crypto
 			Thread[] threads = new Thread[threadCount - 1];
 			for(int i = 0; i < threads.Length; i++)
 			{
-				(threads[i] = new Thread(workerThread, 8192 * 4)).Start();
+				var maxStackSize = 8192 * 4 * (Debugger.IsAttached ? 4 : 1);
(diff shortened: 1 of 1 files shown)


NBitcoin/Crypto/Cryptsharp/SCrypt.cs
```diff
@@ -228,7 +228,8 @@ public static class SCrypt
 			Thread[] threads = new Thread[threadCount - 1];
 			for(int i = 0; i < threads.Length; i++)
 			{
-				(threads[i] = new Thread(workerThread, 8192 * 4)).Start();
```
NicolasDorier: just change `4` to `8`. We are rich

oripio: Ok but 8192 * 4 != 8192 * 8. 
The PR means the current stack is 32k but when use debugger - 128k.

ForNeVeR:
I should note that the issue is not in Rider, but in .NET Core debugger. For now, one cannot debug that code at all using .NET Core debugger with Rider or VSCode on macOS or Linux.

NicolasDorier:
This code was not compiling, I just fixed it and pushed package 4.1.1.84 let me know

oripio:
Works like a charm. Thanks!

PR #670, opened 2019-03-17
Add support for Particl

spazzymoto:


NicolasDorier:
Tried to run the altcoins test? those are passing?

spazzymoto:
I must say i struggled a bit to get the regtest altcoin tests to pass, i think its something in particl core. I will chat to the core guys to see if i need some extra options passed in.

spazzymoto:
Hey @NicolasDorier,
So with Particl being a POS coin, we don't use the generate command to generate blocks and thus a lot of the tests are failing. Would it be an option for me to create new tests specific to Particl and what we currently need for the btcpayserver? Like decoding blocks and transactions etc. And that will give me some time to figure out the POS part while the Particl community can have access to Particl on the btcpayserver?

Rob

NicolasDorier:
Aside of the sync with PoW test, every tests should pass. NBitcoin is a compatibility layer above bitcoin forks. The idea is that if your project works with bitcoin, it works with all other. 

If that is not the case, adding support for your coin is useless as it will not work for people using it.
So no specific tests for Partcl, the altcoins tests (except the PoW one) must pass. Stratis is a PoS coin which made it.

spazzymoto:
Hey man,
It passes all the tests that don't require the generate rpc command, which is like 2 i admit. I'm not sure on Stratis' implementation but what from what i read are they potentially POW as well?

I don't believe its completely useless as i have got the btcpay server up and running using Particl which is mostly what our community is looking for.

If you don't see a way to merge this without passing the tests with the generate rpc commands we will just have to maintain our own forks of the repos, which i was hoping not to do.

NicolasDorier:
The generate command is not useless, without it, it is impossible to test any code using your crypto currency. Stratis added it.

PR #671, opened 2019-03-22
Make uint256 and uint160 implement IEquitable

nopara73:
This refactoring does not modify the behavior, but makes equality comparision of uint256 and uint160 faster and those classes generally more extensible and usable.

> The IEquatable(T) interface is used by generic collection objects such as Dictionary(TKey, TValue), List(T), and LinkedList(T) when testing for equality in such methods as Contains, IndexOf, LastIndexOf, and Remove.

> The IEquatable<T> implementation will require one less cast for these classes and as a result will be slightly faster than the standard object.Equals method that would be used otherwise.

https://stackoverflow.com/questions/2476793/when-to-use-iequatable-and-why

### DIFF
--- a/NBitcoin/UInt2561.cs
+++ b/NBitcoin/UInt2561.cs
@@ -7,7 +7,7 @@ using NBitcoin.Protocol;
 
 namespace NBitcoin
 {
-	public class uint256
+	public class uint256 : IEquatable<uint256>
 	{
 		public class MutableUint256 : IBitcoinSerializable
 		{
@@ -283,6 +283,8 @@ namespace NBitcoin
 			return equals;
 		}
 
+		public bool Equals(uint256 other) => this == other;
+
 		public static bool operator ==(uint256 a, uint256 b)
 		{
 			if(System.Object.ReferenceEquals(a, b))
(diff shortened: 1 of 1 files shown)


NBitcoin/UInt2561.cs
```diff
@@ -662,14 +667,19 @@ public uint160(byte[] vch)
 		public override bool Equals(object obj)
 		{
 			var item = obj as uint160;
-			if(item == null)
+			return Equals(item);
+		}
+
+		public bool Equals(uint160 other)
+		{
+			if (other == null)
```
NicolasDorier: can you do `other is null` instead, so we are sure no operator overload is used?

NBitcoin/UInt2561.cs
```diff
@@ -266,20 +266,25 @@ public uint256(byte[] vch)
 		public override bool Equals(object obj)
 		{
 			var item = obj as uint256;
-			if(item == null)
+			return Equals(item);
+		}
+
+		public bool Equals(uint256 other)
+		{
+			if (other == null)
```
NicolasDorier: can you do other is null instead, so we are sure no operator overload is used?

PR #672, opened 2019-03-22
Clone EndpointConnector

lontivero:
Fixes a missing property (EndpointConnector) in the copy constructor called for the Clone method.


NicolasDorier:
release .89

PR #680, opened 2019-04-01
Enable using different Tor identities per node.

nopara73:
Closes https://github.com/zkSNACKs/WalletWasabi/issues/1293

**UPDATE:** This PR is tested and ready.

The code is probably good, but I couldn't test it. Is there a quick way to test it or should I tie together my local NBitcoin branch with Wasabi somehow?  

If it doesn't fail, then it works as intended since the code has been brought from Wasabi where we have tests for it.  

The trick is that Tor SOCKS5 authentication is actually not authentication, but you can give it any string and if the string is different, then Tor will use a different Tor stream. (With best effort.)

### DIFF
--- a/NBitcoin.Tests/ProtocolTests.cs
+++ b/NBitcoin.Tests/ProtocolTests.cs
@@ -817,7 +817,7 @@ namespace NBitcoin.Tests
 		// You need to run Tor Browser (the test use socks port 9150, not 9050)
 
 #pragma warning disable xUnit1013 // Public method should be marked as test
-		// [Fact]
+		//[Fact]
 		public async Task TestDifferentConnectionMethods()
 #pragma warning restore xUnit1013 // Public method should be marked as test
 		{
@@ -851,10 +851,9 @@ namespace NBitcoin.Tests
 					{
 						var node = await Node.ConnectAsync(Network.Main, endpoint, new NodeConnectionParameters()
 						{
-							EndpointConnector = new DefaultEndpointConnector{ ChangeTorIdentities = changeIpIdentities },
 							TemplateBehaviors =
 							{
-								new SocksSettingsBehavior(Utils.ParseEndpoint("localhost", 9150), onlyForOnionHosts)
+								new SocksSettingsBehavior(Utils.ParseEndpoint("localhost", 9150), onlyForOnionHosts, null, changeIpIdentities)
 							},
 							ConnectCancellation = cancellationToken.Token
 						});
--- a/NBitcoin/Socks/SocksHelper.cs
+++ b/NBitcoin/Socks/SocksHelper.cs
@@ -75,36 +75,51 @@ namespace NBitcoin.Socks
 			sendBuffer[sendBuffer.Length - 1] = (byte)((port & 0x00ff));
 			return sendBuffer;
 		}
+		static Encoding NoBOMUTF8 = new UTF8Encoding(false);
+		public static Task Handshake(Socket socket, EndPoint endpoint, CancellationToken cancellationToken)
+		{
+			return Handshake(socket, endpoint, null, cancellationToken);
+		}
 
-		public static async Task Handshake(Socket socket, EndPoint endpoint, bool changeTorIdentities, CancellationToken cancellationToken)
+		public static async Task Handshake(Socket socket, EndPoint endpoint, NetworkCredential credentials, CancellationToken cancellationToken)
 		{
 			NetworkStream stream = new NetworkStream(socket, false);
-			var selectionMessage = changeTorIdentities ? SelectionMessageUsernamePassword : SelectionMessageNoAuthenticationRequired;
+			var selectionMessage = credentials is null ? SelectionMessageNoAuthenticationRequired : SelectionMessageUsernamePassword;
 			await stream.WriteAsync(selectionMessage, 0, selectionMessage.Length).WithCancellation(cancellationToken).ConfigureAwait(false);
 			await stream.FlushAsync().WithCancellation(cancellationToken).ConfigureAwait(false);
 
 			var selectionResponse = new byte[2];
+			// Note that we use WithCancellation because underlying socket operations does not support true cancellation.
+			// This mainly just abandon the operation.
 			await stream.ReadAsync(selectionResponse, 0, 2).WithCancellation(cancellationToken).ConfigureAwait(false);
 			if (selectionResponse[0] != 5)
 				throw new SocksException("Invalid version in selection reply");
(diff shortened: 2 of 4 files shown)


nopara73:
@lontivero Could you test if this code works somehow? I think you have already tied together Wasabi with a local NBitcoin branch of yours.

NBitcoin/Socks/SocksHelper.cs
```diff
@@ -66,17 +76,57 @@ internal static byte[] CreateConnectMessage(EndPoint endpoint)
...
 			await stream.FlushAsync().WithCancellation(cancellationToken).ConfigureAwait(false);
 
 			var selectionResponse = new byte[2];
 			await stream.ReadAsync(selectionResponse, 0, 2).WithCancellation(cancellationToken).ConfigureAwait(false);
 			if (selectionResponse[0] != 5)
 				throw new SocksException("Invalid version in selection reply");
-			if (selectionResponse[1] != 0)
+			if (selectionResponse[1] == 2)
+			{
+				// https://tools.ietf.org/html/rfc1929#section-2
```
nopara73: Basically testing should be just putting a breakpoint here and if it gets this far and it doesn't throw any exceptions in the rest of the `Handshake` function then the code works as intended.

lontivero:
@nopara73 I've tested with a disabled test already existing in nbitcoin and it and it works. I don't know how to update your PR. My changed UT is in this commit https://github.com/lontivero/NBitcoin/commit/bcbc29b721da8eadccf94df7c55df9a1bab7a988

nopara73:
@lontivero Here, I create a PR from your patch and merged it: https://github.com/nopara73/NBitcoin/pull/1

NBitcoin/Socks/SocksHelper.cs
```diff
@@ -66,17 +76,57 @@ internal static byte[] CreateConnectMessage(EndPoint endpoint)
...
+
+				// Create a random string 21 char string.
+				const string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+				var identity = new string(Enumerable.Repeat(chars, 21)
+				  .Select(s => s[new Random().Next(s.Length)]).ToArray());
+
+				var username = identity;
+				var password = identity;
+				var uName = Encoding.UTF8.GetBytes(username);
+				var passwd = Encoding.UTF8.GetBytes(password);
```
NicolasDorier: are you sure about this? UTF8 by default insert a BOM marker as first byte.

NBitcoin/Socks/SocksHelper.cs
```diff
@@ -66,17 +76,57 @@ internal static byte[] CreateConnectMessage(EndPoint endpoint)
...
+				// https://tools.ietf.org/html/rfc1929#section-2
+				// Once the SOCKS V5 server has started, and the client has selected the
+				// Username / Password Authentication protocol, the Username / Password
+				// subnegotiation begins.  This begins with the client producing a
+				// Username / Password request:
+
+				// Create a random string 21 char string.
+				const string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+				var identity = new string(Enumerable.Repeat(chars, 21)
+				  .Select(s => s[new Random().Next(s.Length)]).ToArray());
```
NicolasDorier: If you run this, you will get `AAAAAABBB8888888888QQQQQ` kind of password, so you need to create `new Random()` only once.

NBitcoin/Socks/SocksHelper.cs
```diff
@@ -66,17 +76,57 @@ internal static byte[] CreateConnectMessage(EndPoint endpoint)
 			return sendBuffer;
 		}
 
-		public static async Task Handshake(Socket socket, EndPoint endpoint, CancellationToken cancellationToken)
+		public static async Task Handshake(Socket socket, EndPoint endpoint, bool changeTorIdentities, CancellationToken cancellationToken)
```
NicolasDorier: Can you ask the user to pass NetworkCredentials here ?

NBitcoin/Socks/SocksHelper.cs
knocte: why not just create a new `Guid` and truncate it?

NicolasDorier:
I will update this PR myself, there is several issues with it I commented on.

NBitcoin/Socks/SocksHelper.cs
```diff
@@ -66,17 +76,57 @@ internal static byte[] CreateConnectMessage(EndPoint endpoint)
...
+			if (selectionResponse[1] == 2)
+			{
+				// https://tools.ietf.org/html/rfc1929#section-2
+				// Once the SOCKS V5 server has started, and the client has selected the
+				// Username / Password Authentication protocol, the Username / Password
+				// subnegotiation begins.  This begins with the client producing a
+				// Username / Password request:
+
+				// Create a random string 21 char string.
+				const string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
```
NicolasDorier: did not knew you could put const here

NBitcoin.Tests/ProtocolTests.cs
```diff
@@ -839,7 +841,7 @@ public async Task TestDifferentConnectionMethods()
 				// Should works for DNS names
 				"ec2-52-14-64-82.us-east-2.compute.amazonaws.com"
 				};
-			foreach (var onlyForOnionHosts in new[] { true, false })
+			foreach (var (onlyForOnionHosts, changeIpIdentities) in new[] { (true, true), (true, false), (false, true), (false, false) })
```
NicolasDorier: cool stuff, did not knew it was possible

NBitcoin/Socks/SocksHelper.cs
lontivero: BOM is inserted as first char in a text when you do `Encoding.UTF8.GetString` but here we are getting the bytes.

NicolasDorier:
I updated your PR to my taste, merging it

NicolasDorier:
Pushed .97.

NBitcoin/Socks/SocksHelper.cs
nopara73: Yes, I'm sure. Both in my `PasswdField` and `UNameField` I have this comment in the code: `// Tor accepts UTF8 encoded passwd`

I have this comment in the code: https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi/TorSocks5/Models/Fields/ByteArrayFields/PasswdField.cs#L13

PR #683, opened 2019-04-18
Allow opt-in RBF using TransactionBuilder

lontivero:
This PR is for discussing a possible implementation to opt-in signaling `RBF` in transactions built using the `TransactionBuilder`. It adds a `TransactionBuilder.OptInRBF` property (default: `False`) that when set to true sets *all* the inputs' sequence to `0xfffffffd`.

### DIFF
--- a/NBitcoin/TransactionBuilder.cs
+++ b/NBitcoin/TransactionBuilder.cs
@@ -553,6 +553,7 @@ namespace NBitcoin
 			CoinSelector = new DefaultCoinSelector(ShuffleRandom);
 			StandardTransactionPolicy = new StandardTransactionPolicy();
 			DustPrevention = true;
+			OptInRBF = false;
 			InitExtensions();
 		}
 
@@ -589,6 +590,15 @@ namespace NBitcoin
 		/// If true, it will remove any TxOut below Dust, so the transaction get correctly relayed by the network. (Default: true)
 		/// </summary>
 		public bool DustPrevention
+		{
+			get;
+			set;
+		}
+
+		/// <summary>
+		/// If true, it will signal the transaction replaceability in every input. (Default: false)
+		/// </summary>
+		public bool OptInRBF
 		{
 			get;
 			set;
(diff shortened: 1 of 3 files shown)


NBitcoin/TransactionBuilder.cs
```diff
@@ -1377,7 +1387,12 @@ bool IsEconomical(Coin c)
 				var input = ctx.Transaction.Inputs.FirstOrDefault(i => i.PrevOut == coin.Outpoint);
 				if(input == null)
 					input = ctx.Transaction.Inputs.Add(coin.Outpoint);
-				if(_LockTime != null && !ctx.NonFinalSequenceSet)
+
+				if(OptInRBF)
```
lontivero: Here is where I am not 100% sure it is correct.

NicolasDorier: It is not correct. Inverse the if and the if/else.
You are doing `if A => RBF elseif B => RBF+STUFF`

So if A is true, we are missing `STUFF`.
So you should do

`if B => RBF+STUFF elseif A =>RBF`

(Note that setting sequence to 0 effecitively put RBF on)

NicolasDorier: Or remove the if else.

`if A => RBF; if B => RBF + STUFF`

NBitcoin/TransactionBuilder.cs
```diff
@@ -1377,7 +1387,12 @@ bool IsEconomical(Coin c)
 				var input = ctx.Transaction.Inputs.FirstOrDefault(i => i.PrevOut == coin.Outpoint);
 				if(input == null)
 					input = ctx.Transaction.Inputs.Add(coin.Outpoint);
-				if(_LockTime != null && !ctx.NonFinalSequenceSet)
+
+				if(OptInRBF)
+				{
+					input.Sequence = Sequence.OptInRBF;
+				}
+				else if(_LockTime != null && !ctx.NonFinalSequenceSet)
```
NicolasDorier: remove the else here, and I think it is good.

NBitcoin/TransactionBuilder.cs
lontivero: > (Note that setting sequence to 0 effectively activate RBF)
F*ck! You're right.

lontivero:
Done. The failing test is `NBitcoin.Tests.ProtocolTests.CanGetChainsConcurrenty [FAIL]` on OSX.

PR #675, opened 2019-03-26
Support Miniscript as a separate dll.

joemphilips:
* [x] Complete implementation
* [x] Create C# Facade as an extension methods in `NBitcoin` namespace 
* [x] Create `NBitcoin.nuspec` file so that I can include `Miniscript.dll` to NBitcoin's nupkg.
* [ ] Tweak AssemblyInfo and `.fsproj`

PR #684, opened 2019-04-22
Support Miniscript  (with C#)

joemphilips:
Previous attempt is in #675
I'e been rewriting from F# to C# for following reason.

1. Mixing F# and C# in one solution complicates the build process.
1. Often the editor screwed me up. which also caused a Maintainability issue.
1. I wanted to refactor some logic. Namely, internal representation can be simplified by using classes more closer to the actual script.
1. Ease for review.


NicolasDorier:
Let me know when it is ready to review. I will probably use miniscript in NBXplorer so users can't have their own way to make arbitrary scripts.

joemphilips:
I think I can finish it in three days or so.

NicolasDorier:
I have quickly checked. I think you do way too complicated for GetHashCode/Equal implementations.
I would say you don't have to implement them at all, or if you do, you can just compare the string representation of the AST of the miniscript.

joemphilips:
@NicolasDorier Thanks for review.
Yeah those GetHashCode and Equal methods are based on decompiled F# version.
I think it doesn't hurt so we can leave it there though.

joemphilips:
Ready for review.

NicolasDorier:
Wow so F# just do this by default? cool.

knocte:
> Wow so F# just do this by default?

Yes!

joemphilips:
The error in CI seems nothing to do with my code.

joemphilips:
> The error in CI seems nothing to do with my code.

Oops sorry. It was my fault. Will fix now.

NBitcoin.Tests/Helpers/PrimitiveUtils.cs
```diff
@@ -0,0 +1,41 @@
+
```
joemphilips: This class contains a method used to to be in `transaction_test` .
Some functions (e.g. `RandomCoin` ) is quite reusable so I've moved it to here.

NBitcoin.Tests/MiniscriptTests.cs
```diff
@@ -0,0 +1,443 @@
...
+			var coins = GetRandomCoinsForAllScriptType(Money.Coins(0.5m), sc);
+			return builder.AddCoins(coins)
+				.SendFees(Money.Coins(0.001m))
+				.SendAll(new Key()); // dummy output
+		}
+
+		[Fact]
+		[Trait("UnitTest", "UnitTest")]
+
+		public void ShouldPlayWellWithTransactionBuilder_2()
```
joemphilips: See this test for how `TransactionBuilder` Supports HTLC by Miniscript

NBitcoin/BIP174/PSBTInput.cs
```diff
@@ -350,33 +350,6 @@ internal void Combine(PSBTInput other)
 
 		public bool IsFinalized() => final_script_sig != null || final_script_witness != null;
 
-		/// <summary>
```
joemphilips: Removing this method has nothing to do with Miniscript. But I noticed that it is legacy code from my version of PSBT. So I removed it.

NBitcoin/BuilderExtensions/MiniscriptBuilderExtension.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using NBitcoin.Miniscript;
+
+namespace NBitcoin.BuilderExtensions
+{
+	public class MiniscriptBuilderExtension : BuilderExtension
+	{
+		// Probably it is impossible to Combine items in generic way.
+		public override bool CanCombineScriptSig(Script scriptPubKey, Script a, Script b)
```
joemphilips: This is the part I didn't know what to do.
Combining ScriptSig in Miniscript seems impossible (since there is no straightforward way to tell which [signature|Secret] corresponds to which [pubkey|Secret hash] in the script.) It seems we need some non-trivial refactoring on TransactionBuilder side.

NBitcoin/Miniscript/Satisfy.cs
```diff
@@ -0,0 +1,674 @@
...
+			Sequence timelockS = timelock;
+			Sequence ageS = age.Value;
+			if (!ageS.IsRelativeLock)
+			{
+				errors.Add(new SatisfyError(SatisfyErrorCode.RelativeLockTimeDisabled, this));
+				return false;
+			}
+			if (ageS.LockType == SequenceLockType.Time)
+			{
+				errors.Add(new SatisfyError(SatisfyErrorCode.UnSupportedRelativeLockTimeType, this));
```
joemphilips: Currently, I decided to support only block-height based lock for time (which is `OP_CSV` )
so in DSL it `"time(0) ~ time(65535)"` is allowed.

I don't know this will be standard among other Miniscript implementation. So it might change in the future.

NBitcoin/TransactionBuilder.cs
```diff
@@ -1433,6 +1483,11 @@ bool IsEconomical(Coin c)
 					input.Sequence = 0;
 					ctx.NonFinalSequenceSet = true;
 				}
+				if (_Sequence != null)
```
joemphilips: I'm not sure this won't collide with usage of RBF.

NicolasDorier: You just need to set `ctx.NonFinalSequenceSet` to `true`.

NBitcoin/TransactionBuilder.cs
```diff
@@ -664,7 +687,21 @@ public TransactionBuilder SetLockTime(LockTime lockTime)
 			return this;
 		}
 
+		Sequence? _Sequence;
+		public TransactionBuilder SetRelativeLockTime(int lockHeight)
```
NicolasDorier: I think you can just use `SetSequence(Sequence)`.

My only problem with this is that sequence is a per input concept, not a global tx concept. 
But I guess this is still useful if a wallet has several keys depending on the same miniscript?

NBitcoin/TransactionBuilder.cs
```diff
@@ -1962,8 +2018,8 @@ private Script CreateScriptSig(TransactionSigningContext ctx, ICoin coin, Indexe
 			{
 				if(extension.CanGenerateScriptSig(scriptPubKey))
 				{
-					var scriptSig1 = extension.GenerateScriptSig(scriptPubKey, keyRepo, signer);
-					var scriptSig2 = extension.GenerateScriptSig(scriptPubKey, signer2, signer2);
+					var scriptSig1 = extension.GenerateScriptSig(scriptPubKey, keyRepo, signer, preimageRepo, txIn.TxIn.Sequence);
+					var scriptSig2 = extension.GenerateScriptSig(scriptPubKey, signer2, signer2, preimageRepo, txIn.TxIn.Sequence);
```
NicolasDorier: I don't think that `txIn.TxIn.Sequence` can ever be useful to an extension.

NBitcoin/BuilderExtensions/MiniscriptBuilderExtension.cs
NicolasDorier: Combining ScriptSig is not too important. Especially we now have PSBT.

NicolasDorier:
Unrelated to this PR, but it makes me think that a BIP should be proposed for PSBT to support adding preimages.

NBitcoin/TransactionBuilder.cs
joemphilips: It is not necessary, but I think it is useful in the case of Miniscript extension
because txbuilder fails to sign when
1. script contains OP_CSV. and
2. user forgets to set Sequence appropriately.

without this, users will notice that they have signed to the invalid tx when they try to broadcast.

joemphilips:
> Unrelated to this PR, but it makes me think that a BIP should be proposed for PSBT to support adding preimages.

Indeed, I was quite surprised that no one has written yet.

NBitcoin/TransactionBuilder.cs
joemphilips: I think the api to take `int blockHeight` or `TimeSpan period` is better than taking Sequence because we must assure tx nVersion is bigger than 1.
users are not interested into how the rule is encoded. they just want to use the relative locktime feature.

What about
`SetRelativeLockTimeTo(ICoin whichCoin, int lockHeight)`

and hold the info as `Dictionary<ICoin, Sequence>` ? In this way, we can assure that it will affect only specific input.

joemphilips: Tell me if the current implementation is not good. I decided to use `Dictionary<OutPoint, Sequence>` since Coins does not support `GetHashCode`

NBitcoin/TransactionBuilder.cs
joemphilips: (To make me clear, the user does not have to set nSequence even if OP_CSV is in the script if he tries to redeem through non-timelocked path)

NicolasDorier: But don't you have this problem in all case anyway? The transaction might still not be broadcastable depending on the Sequence, and the only way to know is to have the previous UTXOs's height.

NicolasDorier:
@joemphilips if you feel motivated to do the BIP, you can ask to people on #bitcoin-dev on IRC about it. If nobody did, do your own, publish on the mailing list.
I don't think there is lot's of way to disagree with, so as long as it is well redacted, it should be fine.

NBitcoin/TransactionBuilder.cs
joemphilips: Yes, TX might be un-broadcastable anyway, difference is that if the user signs to nSequence which is not suitable to pass `OP_CSV` (e.g. disable flag is set) that signature will be useless for redeeming through timelocked redeem condition. so the tx wont be valid forever.

NicolasDorier:
Hey, so I just created the type `IHDScriptPubKey` that you want to implement for `MiniScript`.
This allow to use PSBT methods like `GetBalance` or `SignAll`.

If it get possible to finalize any miniscript then I can move this to NBXplorer then BTCpayServer.

joemphilips:
Ah thanks for mentioning.
PSBT and OutputDescriptor support for BTCPayServer is super cool IMO ... It opens up possibility for many extensions.

I've just realized that the spec for output descriptor in bitcoin core has been much detailed and broad than last time I've read. I will update parsers and its tests in this weekend. (And finalizer afterwords)

BTW, I've been using `namespace NBitcoin.Miniscript` but this collides with `Miniscript` class in there, so I want to change the namespace name, what do you think is the best name? ( `NBitcoin.SmartContract` or `NBitcoin.MiniscriptModule` ?)

NicolasDorier:
@joemphilips I did an object model to build a OutputDescriptor, but it does not support parsing.

Maybe `NBitcoin.Scripting` for both MiniScript and OutputDescriptor is good.

NicolasDorier:
lol this PR is massive, can you break it down into 2 PR, like one with only the output descriptor stuff and the other one with miniscript?

joemphilips:
OK. I should have created separate PR in the first place ... Miniscript and OutputDescriptor is quite different beast.

joemphilips:
Closing in favor of #694, #695 , #696 .

PR #686, opened 2019-04-24
RBF is not disabled by default anymore

knocte:
But it's not enabled by default either: the developer is forced to
make a decision if he wants to avoid the compile-time warnings.

This way we avoid the potential problems that could have been caused
by simply switching the default to be RFB-enabled, as discussed
originally here: https://github.com/MetacoSA/NBitcoin/pull/355


NicolasDorier:
Wait I don't understand why there is both `OptInRBF` and `RBF`. What is the rational @lontivero ?

If `RBF` is set to true we have RBF, else we don't.... 

https://github.com/MetacoSA/NBitcoin/pull/683

knocte:
Nicolas: Lucas didn't introduce two flags. After Lucas' PR, there was only `OptinRBF`. However, after this PR, calling it with the prefix "Optin" wouldn't make much sense, so I renamed it to "RBF". However, I left the old property with an [Obsolete] attribute in order not to break API.

NicolasDorier:
Ah ok.
Well I think we can rename fine in this case the change is 5 days ago we will not break too much...

NicolasDorier:
I don't really like bothering users with a question they don't care or are unlikely to understand. (what the hell is RBF??)

knocte:
> Well I think we can rename fine in this case the change is 5 days ago we will not break too much...

Ok, fixed.

> (what the hell is RBF??)

I can rename the parameter to `replaceByFeeEnabled`?

NicolasDorier:
No it should be named RBF, but most users don't care about it and asking them a question if they don't care about it is not a good solution I think.

I am mixed... ping @lontivero what do  you think?

knocte:
> asking them a question if they don't care about it is not a good solution I think.

Well, I can use XML documentation for that API, where we recommend to enable it unless they need to care about backwards compatibility.

NicolasDorier:
That make sense.

knocte:
> That make sense.

Done, and rebased (and fixed build).

NBitcoin/TransactionBuilder.cs
```diff
@@ -602,17 +607,18 @@ public bool DustPrevention
 		{
 			get;
 			set;
-		}
+		}
 
+		[Obsolete("Use Network.CreateTransactionBuilder(RBF) or ConsensusFactory.CreateTransactionBuilder(RBF) instead")]
```
NicolasDorier: remove the obsolete, if the user change here, it is not big deal.

knocte: fixed (and CI pipelines are green now)

NBitcoin/TransactionBuilder.cs
```diff
@@ -555,17 +555,22 @@ internal BuilderGroup CurrentGroup
 				return _CurrentGroup;
 			}
 		}
-		[Obsolete("Use Network.CreateTransactionBuilder() or ConsensusFactory.CreateTransactionBuilder() instead")]
-		public TransactionBuilder()
```
NicolasDorier: do not remove this for now

NBitcoin/ConsensusFactory.cs
```diff
@@ -120,13 +120,32 @@ public virtual TxOut CreateTxOut()
...
+		protected virtual TransactionBuilder CreateTransactionBuilderCore(bool RBF)
+		{
+			return new TransactionBuilder(RBF);
+		}
+
+		[Obsolete("Use CreateTransactionBuilder(bool RBF)")]
 		protected virtual TransactionBuilder CreateTransactionBuilderCore()
 		{
 #pragma warning disable CS0618 // Type or member is obsolete
 			return new TransactionBuilder();
```
NicolasDorier: Call `CreateTransactionBuilderCore(false)`

NicolasDorier: Call CreateTransactionBuilderCore(false)

knocte: fixed

nopara73:
Concept ACK.

> Well I think we can rename fine in this case the change is 5 days ago we will not break too much...

Agree, renaming is fine, probably nobody using it yet.

knocte:
> nobody using it yet.

Then let's merge it asap and bump?

lontivero:
> No it should be named RBF, but most users don't care about it and asking them a question if they don't care about it is not a good solution I think.
>
> I am mixed... ping @lontivero what do you think?

It is okay for me but let me be honest here: the name RBF sucks bad because the semantic is unclear, for example: what does it means IsRBF? Does it means that the tx is replaceable? It is confusing.  Bitcoin core calls it `OptInRBF` (what also sucks) but its semantic is at least a bit more clear. I would call it `ExplicitlySignalReplaceability` or something like that in a no library component. 
I think the high level wording are:
* `replaceable transactions` to those _signaling_ (iimplicitly/explicitly) replaceability,
* `replacement transactions` to those that spend the same inputs that and unconfirmed tx signaling replaceability.

So, in the library it is okay to call it `RBF` but in higher level software should be called with a more meaningful name.

NicolasDorier:
@knocte have you fixed all the warnings?

knocte:
> @knocte have you fixed all the warnings?

Oh well... I started doing this, and found out that Transaction.Sign() calls CreateTransactionBuilder()... Which means that we should add overloads on Sign() (or Transaction ctor) with RBF flag too.

I went for the latter, and oh my, the rabbit hole is very deep. I need help.

knocte:
(I pushed my warnings-fixage WIP work as a 2nd commit)

NicolasDorier:
@knocte I think that you just have to call with `RBF=false` on existing stuff.

knocte:
> @knocte I think that you just have to call with RBF=false on existing stuff.

Sorry for late reply, but I never understood the statement above. If I mark RBF=false on existing stuff, it would defeat the point of this PR.

Anyway I've been thinking about this lately, and as you Nicolas seem to always recommend TransactionBuilder API over Transaction API, would you accept marking `Transaction` type as `Obsolete`? This way I can live with it being RBF=false by default while TransactionBuilder has RBF enabled by default.

PR #687, opened 2019-04-25
Fix GetBlockchainInfoAsync() for Bitcoin 0.13

ultimaweapon:
Some coins still based on Bitcoin 0.13 (e.g. Zcoin). In Bitcoin 0.13, the fields `initialblockdownload`, `size_on_disk` and `since` in `bip9_softforks` is absent.


NicolasDorier:
hey, this is a breaking change so I won't accept it.
However, you can do a test whether the field exists, and if not, set a default value.

ultimaweapon:
Fixed.

PR #691, opened 2019-05-08
Make uint256 and uint160 IComparable

joemphilips:
I need this for using uint256 as a key for F#'s `Map` type.

### DIFF
--- a/NBitcoin/UInt2561.cs
+++ b/NBitcoin/UInt2561.cs
@@ -7,7 +7,7 @@ using NBitcoin.Protocol;
 
 namespace NBitcoin
 {
-	public class uint256 : IEquatable<uint256>
+	public class uint256 : IEquatable<uint256>, IComparable<uint256>, IComparable
 	{
 		public class MutableUint256 : IBitcoinSerializable
 		{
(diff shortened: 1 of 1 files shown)


nopara73:
This can be also useful to eventually be able to compare and order `OutPoint`s with each other based on @kristovatlas's [BIP69.](https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki) I'm not sure we have to go that far anytime soon, maybe if someone needs it, but it's a good stepping stone.

NBitcoin/UInt2561.cs
```diff
@@ -477,8 +477,25 @@ public override int GetHashCode()
 			}
 			return hash;
 		}
+
+		public int CompareTo(object other)
+		{
+			var o = other as uint256;
+			if (o == null)
+				return -1;
+			return CompareTo(o);
```
nopara73: No need to do the casting before null check.

```cs
if(other is null)
	return -1;
var o = other as uint256;
return CompareTo(o);
```

NBitcoin/UInt2561.cs
```diff
@@ -819,5 +836,20 @@ public override int GetHashCode()
 			}
 			return hash;
 		}
+		public int CompareTo(object other)
+		{
+			var o = other as uint160;
+			if (o == null)
+				return -1;
+			return CompareTo(o);
```
nopara73: Same here.

NBitcoin/UInt2561.cs
```diff
@@ -477,8 +477,25 @@ public override int GetHashCode()
...
+
+		public int CompareTo(object other)
+		{
+			if (other is null)
+				return -1;
+			var o = other as uint256;
+			return CompareTo(o);
+		}
+
+		public int CompareTo(uint256 other)
```
NicolasDorier: You can treat the `other is null` case here

joemphilips:
closing since #697 has done the same.

PR #694, opened 2019-05-22
Prepare Parser for scripting

joemphilips:
Original PR is #684 

This creates `NBitcoin.Scripting.Parser` namespace, and puts Monadic Parser combinator for internal usage.
It does not affect the library's public AP,I nor it has tests. Tests will be done in following PRs.

The API almost completely follows those of [Sprache](https://github.com/sprache/Sprache), the biggest difference is that it is generalized against its input type. (Sprache only allows string as an input, but this one will take arbitrary `IEnumerable<T>` for input.) This is crucial for Miniscript decompilation.

This might be a good candidate for using `Span<T>` as an input type to improve the speed. But I didn't want to do an early optimization so I didn't.

### DIFF
--- a/NBitcoin/Scripting/Parser/IInput.cs
+++ b/NBitcoin/Scripting/Parser/IInput.cs
@@ -3,8 +3,8 @@ using System.Collections.Generic;
 
 namespace NBitcoin.Scripting.Parser
 {
-    public interface IInput<out T>
-    {
+	internal interface IInput<out T>
+	{
 		IInput<T> Advance();
 
 		T GetCurrent();
--- a/NBitcoin/Scripting/Parser/IInput.cs
+++ b/NBitcoin/Scripting/Parser/IInput.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 
 namespace NBitcoin.Scripting.Parser
 {
-	internal interface IInput<out T>
+	internal interface IInput<out T> : IEnumerable<T>
 	{
 		IInput<T> Advance();
 
--- a/NBitcoin/Scripting/Parser/IInput.cs
+++ b/NBitcoin/Scripting/Parser/IInput.cs
@@ -9,8 +9,6 @@ namespace NBitcoin.Scripting.Parser
 
 		T GetCurrent();
 
-		T GetNext();
-
 		bool AtEnd { get; }
 		int Position { get; }
 		IDictionary<object, object> Memos { get; }
--- a/NBitcoin/Scripting/Parser/IInput.cs
+++ b/NBitcoin/Scripting/Parser/IInput.cs
@@ -1,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+
+namespace NBitcoin.Scripting.Parser
+{
+    public interface IInput<out T>
+    {
+		IInput<T> Advance();
+
+		T GetCurrent();
+
+		T GetNext();
+
+		bool AtEnd { get; }
+		int Position { get; }
+		IDictionary<object, object> Memos { get; }
+	}
+}
--- a/NBitcoin/Scripting/Parser/Parse.Char.cs
+++ b/NBitcoin/Scripting/Parser/Parse.Char.cs
@@ -106,7 +106,7 @@ namespace NBitcoin.Scripting.Parser
 		{
 			return from nothing in Return<char, string>("")
 					   // dummy so that CultureInfo.CurrentCulture is evaluated later
-				   from dot in String((ci ?? CultureInfo.CurrentCulture).NumberFormat.NumberDecimalSeparator).Text()
+				   from dot in String((ci ?? CultureInfo.InvariantCulture).NumberFormat.NumberDecimalSeparator).Text()
 				   from fraction in Number
 				   select dot + fraction;
 		}
@@ -117,15 +117,9 @@ namespace NBitcoin.Scripting.Parser
 		}
 
 		/// <summary>
-		/// Parse a decimal number using the current culture's separator character.
-		/// </summary>
-		public static readonly Parser<char, string> Decimal = DecimalWithLeadingDigits().XOr(DecimalWithoutLeadingDigits());
-
-		/// <summary>
 		/// Parse a decimal number with separator '.'.
 		/// </summary>
-		public static readonly Parser<char, string> DecimalInvariant = DecimalWithLeadingDigits(CultureInfo.InvariantCulture)
-																	 .XOr(DecimalWithoutLeadingDigits(CultureInfo.InvariantCulture));
-
+		public static readonly Parser<char, string> Decimal = DecimalWithLeadingDigits()
+																	 .XOr(DecimalWithoutLeadingDigits());
 	}
 }
--- a/NBitcoin/Scripting/Parser/Parse.Char.cs
+++ b/NBitcoin/Scripting/Parser/Parse.Char.cs
@@ -1,0 +1,131 @@
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+
+namespace NBitcoin.Scripting.Parser
+{
+	internal static partial class Parse
+	{
+		private const string ValidHex = "0123456789abcdef";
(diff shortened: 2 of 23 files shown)


NicolasDorier:
Awesome, I suggest putting all those classes internal?

NBitcoin/Scripting/Parser/IInput.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+
+namespace NBitcoin.Scripting.Parser
+{
+    public interface IInput<out T>
```
NicolasDorier: This one should be internal ?

NicolasDorier: Also, I would suggest to inherit  from `IEnumerator<out T>` and remove `Advance/GetCurrent/GetNext`.
Then just make methods returning `IInput<out T>` before implement `IEnumerable<T>` and return `IInput<T>` .

The nice thing is that people can use `foreach` then.

joemphilips: Nice. I wonder why this didn't come up to my mind...

NBitcoin/Scripting/Parser/ParseException.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+
+namespace NBitcoin.Scripting.Parser
+{
+	/// <summary>
+	/// Represents an error that occurs during parsing.
+	/// </summary>
+	public class ParseException : Exception
```
NicolasDorier: Rename to `ParsingException`, inherits from `FormatException`

NBitcoin/Scripting/Parser/IInput.cs
joemphilips: I've been trying to inherit from IEnumerator<out T>. But it is not working well :(
It is because that Input for Parser combinator must be immutable. (`Advance()` will create another instance each time it proceeds)
IEnumerator is only for mutable object.
I think inheriting from `IEnumerable` and delegate GetEnuemrator call to underlying Source object is enough.

knocte:
> IEnumerator is only for mutable object.

Not sure I agree. You could create a specific Enumerator class for this type which is mutable without making your type mutable.

joemphilips:
> Not sure I agree. You could create a specific Enumerator class for this type which is mutable without making your type mutable.

 You mean make my class inherit from `IEnumerable<T>` instead of `IEnumerator<T>` right? That's  what I have done here.

NicolasDorier:
@joemphilips if my suggestion is not really used and does not make the code more visible, then you can just roll it back.

NBitcoin/Scripting/Parser/Parse.Char.cs
```diff
@@ -0,0 +1,131 @@
...
+		/// <summary>
+		/// Parse a number.
+		/// </summary>
+		public static readonly Parser<char, string> Number = Numeric.AtLeastOnce().Text();
+
+		static Parser<char, string> DecimalWithoutLeadingDigits(CultureInfo ci = null)
+		{
+			return from nothing in Return<char, string>("")
+					   // dummy so that CultureInfo.CurrentCulture is evaluated later
+				   from dot in String((ci ?? CultureInfo.CurrentCulture).NumberFormat.NumberDecimalSeparator).Text()
```
NicolasDorier: oh no, never use `CurrentCulture`!

NicolasDorier:
Oh actually it is used in the test, seems better.

NBitcoin/Scripting/Parser/Parse.Char.cs
joemphilips: I've drop the support for culture-aware decimal.
now it uses InvariantCulture only.

NicolasDorier:
Tests failing rebasing on top of master should fix it

joemphilips:
Rebased but not fixed.

It seems that some parts of codes are using obsolete `PubKey.GetAddress` 

```
BIP38\BitcoinConfirmationCode.cs(125,58): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinEncryptedSecret.cs(39,49): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinEncryptedSecret.cs(94,49): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinEncryptedSecret.cs(202,27): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinPassphraseCode.cs(236,27): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinConfirmationCode.cs(125,58): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinEncryptedSecret.cs(39,49): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinEncryptedSecret.cs(94,49): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinEncryptedSecret.cs(202,27): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' [C:\projects\nbitcoin\NBitcoin\NBitcoin.csproj]
BIP38\BitcoinPassphraseCode.cs(236,27): warning CS0618: 'PubKey.GetAddress(Network)' is obsolete: 'Use GetAddress(ScriptPubKeyType.Legacy, network) instead' 
```

NicolasDorier:
Yes you can fix... but so strange, on master the build is passing!

NBitcoin/Scripting/Parser/StringInput.cs
```diff
@@ -31,12 +32,12 @@ public IInput<char> Advance()
 
 		public IEnumerator<char> GetEnumerator()
 		{
-			return this.Source.GetEnumerator();
+			return this.Source.AsEnumerable().GetEnumerator();
```
NicolasDorier: ?? should not be needed

joemphilips: I thought so too, but build fails when `netstandard1.3`

joemphilips: Weird ... with `AsEnumerable()` , netstandard1.3 build succeeds in local. But not in CI.

joemphilips: I have to go a sleep. Will think about this tomorrow. (I may just give up supporting IEnumerable<T> for IInput and rollback to the original.)

joemphilips: Fixed.

NicolasDorier:
What should I review and merge in priority after @joemphilips ?

joemphilips:
I want you to take a look for #696 since the spec is well formalized than #695 , and it has some changes to an existing objects.

But honestly, I myself am not in hurry so please take your time. @NicolasDorier

PR #695, opened 2019-05-22
Support miniscript

joemphilips:
Depends on #694 Previous attempt is in #684 

This is a somewhat primitive version of Miniscript, which does not support taking the `xpub` as its internal key representation. So it is closer to an [Ivy](https://github.com/ivy-lang/ivy-bitcoin) rather than an output descriptor.

Integration with the output descriptor is not impossible, but I think we should wait for bitcoin core to support it.


joemphilips:
Closing since the spec for Miniscript has changed.
a new attempt is in #764

PR #696, opened 2019-05-22
Support output descriptor

joemphilips:
depends on #694 , previous attempt is #684 

Currently it is independent from #695 , and it follows bitcoin core's spec.


joemphilips:
Actually I've just realized that many Inconvenience that I'm facing right now might be fixed by using OutputDescriptor. I will  revive this PR and add RPC methods too from now on.

joemphilips:
There is a new descriptor type `sortedmulti` in the bitcoin core so I'm trying to update the descriptor now.

NicolasDorier:
ping @lontivero do you plan using output descriptor this in the near term? If so, can you review that?

lontivero:
I am going to review it during the weekend. I think Wasabi could use output descriptors soon. We could display the OD for exporting the wallet and imported in Core or export public/private keys in a better way.

NBitcoin.Tests/OutputDescriptorTests.cs
```diff
@@ -0,0 +1,348 @@
...
+				for (int t = 0; t < 2; ++t)
+				{
+					bool isHardend = (flags & HARDENED) != 0;
+					var keyProvider = isHardend ? keysPriv : keysPub;
+
+					var scriptProvider = new FlatSigningRepository();
+					Assert.True((t != 0 ? parsePriv : parsePub).TryExpand((uint)i, keyProvider.GetPrivateKey, scriptProvider, out var spks));
+					Assert.Equal(spks.Count, expectedScript.Length);
+
+					// --- cache ---
```
joemphilips, line 248: If you want to move your hand while reviewing, it may be good idea to port the caching part from bitcoin-core. the test is in [Line 148](https://github.com/bitcoin/bitcoin/blob/6a2ba626852bd5bd8531dfa335c78fa87e4f6282/src/test/descriptor_tests.cpp#L148)-215 . Which should fit in this line.
@lontivero

lontivero:
This work is not a monkey porting directly from bitcoin core but a implementation from scratch, i mean, it is clearly based on the bitcoin core but not copied. Classes, methods, identifiers, code structures are different and that makes the "comparison" hard to do. Most of the changes make sense to me btw, `HD` instead of `Bip32` and other similar are preferable to me too.

Some parts are completely different, the parser is the more obvious case. Given it is hard to grasp to me (because there are concepts that are unknown to me) i decided to insert modifications in the core tests and in this pr's test in order to see if they both behave in the same way. The bitcoin core tests pass green even if i insert empty spaces before, in between or after parenthesis, brackets. The tests in this pr fail. 

There are a few design decisions that I don't understand and that could be simply the result of  coding paradigms but i would like to understand. For example, `GetPubkey` method in Core is overridden by each `Descriptor` class while in this PR is implemented with a `switch` statement comparing types. What is the reason @joemphilips?

I will continue reviewing and understanding the code when I find more time. This is a great work @NicolasDorier, just give me a few more days. Btw, I enjoy the functional style but i agree that the parser was coded in a way that c# does not fit well.

-----------

@joemphilips in Core the cache is there like a think that all `KeyProviders` should be aware of but it is obviously only for `HDKeyProviders` and i need to get more familiar with your design in order to be able to do it.

joemphilips:
> There are a few design decisions that I don't understand and that could be simply the result of coding paradigms but i would like to understand. For example, GetPubkey method in Core is overridden by each Descriptor class while in this PR is implemented with a switch statement comparing types. What is the reason @joemphilips?

Short answer: It is a functional way.

Long answer: 1. I want implementation of the OutputDescriptor to be "closed" inside this library. i.e. user's should not create the new type by overriding `OutputDescriptor`, So I did not want to use virtual or abstract methods.
2. I think it is easier to understand when the logics are grouped in this way. If we use abstract methods we might forget to override in subtypes. In this way it is easy to see that methods are implemented for every possible types.

> i decided to insert modifications in the core tests and in this pr's test in order to see if they both behave in the same way. The bitcoin core tests pass green even if i insert empty spaces before, in between or after parenthesis, brackets. The tests in this pr fail.

This is good catch. When I started this work, bitcoin-core OD was very strict about parsing and it did not permit any white space. I will update the parser then.

Btw, I found out that OD does not help me for the problem I was facing right now, so there is no hurry in my side. Take your time for reviewing.

NBitcoin/Scripting/OutputDescriptorParser.cs
```diff
@@ -0,0 +1,303 @@
...
+				.Or(PCombo(repo))
+				.Or(PInner(repo))
+				.Or(PWSH(repo))
+				.Or(PSH(repo)).End();
+
+		internal static bool TryParseOD(string str, out OutputDescriptor? result, bool requireCheckSum = false, ISigningRepository? repo = null)
+			=> TryParseOD(str, out _, out result,requireCheckSum, repo);
+		private static bool TryParseOD(string str, out string? whyFailure, out OutputDescriptor? result, bool requireCheckSum = false, ISigningRepository? repo = null)
+		{
+			str = str.Replace(" ", "");
```
lontivero, line 250: NRE here if the caller pass `null`. Given it is not possible to prevent callers to pass null I think it is a good idea to check for null on the public interface.

NBitcoin/Scripting/PubKeyProvider.cs
```diff
@@ -0,0 +1,311 @@
...
+			{
+				if (extkey == null)
+					throw new ArgumentNullException(nameof(extkey));
+
+				if (path == null)
+					throw new ArgumentNullException(nameof(path));
+
+				Extkey = extkey;
+				Path = path;
+				Derive = derive;
```
lontivero, line 71: Here it could be convenient to check the `derive` value is a valid `DeriveType`. A wrong value here could result in misbehaviors and derivation of wrong keys. 

```c#
			if (!Enum.IsDefined(typeof(DeriveType), value))
			{
				throw new ArgumentException("Invalid value for DeriveType", nameof()derive);
			}
```

lontivero (commented):
Some comments for your consideration.

NicolasDorier:
> I want implementation of the OutputDescriptor to be "closed" inside this library. i.e. user's should not create the new type by overriding OutputDescriptor, So I did not want to use virtual or abstract methods.

When you want to do this, you can just declare the abstract/virtual methods internal.

joemphilips:
> When you want to do this, you can just declare the abstract/virtual methods internal.

But doesn't it make it impossible for user to use the method?

NicolasDorier:
@joemphilips yes, one thing often used also is have a method "BlahBlah" public and another one "BlahBlahCore" internal/protected.
With BlahBlah which don't have to do much... That's ugly if you have too many methods that need that, but if it is a one or two, I think that;s fine.

NicolasDorier:
@lontivero you plan to still review some stuff or should I start review as well?

lontivero:
I will not review it more.

NBitcoin.Tests/util_tests.cs
```diff
@@ -183,7 +183,7 @@ public void CanParseRootedKeyPath()
 			Assert.Equal("7b09d780", result.MasterFingerprint.ToString());
 			Assert.True(RootedKeyPath.TryParse("7b09d780/", out result));
 			Assert.Equal("7b09d780", result.MasterFingerprint.ToString());
-			Assert.Equal("7b09d780/", result.ToString());
+			Assert.Equal("7b09d780", result.ToString());
```
NicolasDorier, line 186: we should not change that.

NicolasDorier:
@joemphilips I think I will merge that soon. One question: About `KeyId` and `WitScriptId`. Do you really need it in say `FlatSigningRepository`?
As time pass, I get more and more convinced that KeyId/ScriptId/WitScriptId are classes that should never have existed in the first place, so I am curious why you need it.

joemphilips:
I can try using `uint160`/`uint250` instead if you want, I've just used it because I like using different types for those cases. (It makes hard to misuse the API.)
Let me check if that works or not.

joemphilips:
After testing a bit, I think it is possible to use `uint160/uint256` for `ISigningRepository` but It requires substantial refactoring.

e.g. I want to use the api like `pk.WitHash.ScriptPubKey.Hash.ScriptPubKey` or `PayToPubkeyHashTemplate.ExtractScriptPubKeyParameters()` which uses `KeyId` or `ScriptId`so I first have to add a new methods/properties to `PubKey` , `ScriptTemplate` , that will be huge change on the existing code. 

I suppose it is out of the scope for this PR, if you really want to make `KeyId/ScriptId` obsolete, first we’d better make those types obsolete in existing code.

(which I don’t think is necessary though btw, having wrapper types like `WitScriptId` and `ScriptId` is a nice way to specify the intention.)

NBitcoin.Tests/util_tests.cs
joemphilips: Fixed in https://github.com/MetacoSA/NBitcoin/pull/696/commits/2d521d4817a3b0621988c1113311aada38d0546f

NicolasDorier:
@joemphilips don't refactor to use uint160/uint256 instead I just want to know, why you need those Ids classes rather than just ScriptPubKey

joemphilips:
1. It is done in that way in original implementation
2. It is a bit smaller
3. It is more easy to understand the intention (e.g. for `TryGetPubKey(KeyId keyId, out PubKey pubkey)` , if the first argument is Script, reader might get confused whether that is ScriptPubKey for the p2pk, or ScriptPubKey for the p2pkh)

joemphilips:
Would you consider merging this? I have a plan to use it.

NicolasDorier:
Released a new version.

PR #697, opened 2019-05-24
Implement IComparable for uint256

ultimaweapon:
This will make it able to put `uint256` as a key in sorted collection (e.g. `SortedDictionary`) without custom implementation for `IComparer`.

### DIFF
--- a/NBitcoin/UInt2561.cs
+++ b/NBitcoin/UInt2561.cs
@@ -1,13 +1,11 @@
 ﻿
 using System;
 using System.Linq;
-using System.IO;
 using NBitcoin.DataEncoders;
-using NBitcoin.Protocol;
 
 namespace NBitcoin
 {
-	public class uint256 : IEquatable<uint256>
+	public class uint256 : IComparable<uint256>, IEquatable<uint256>, IComparable
 	{
 		public class MutableUint256 : IBitcoinSerializable
 		{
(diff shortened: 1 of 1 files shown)


NBitcoin/UInt2561.cs
```diff
@@ -288,6 +286,26 @@ public bool Equals(uint256 other)
...
+			{
+				return 1;
+			}
+
+			return Comparison(this, other);
+		}
+
+		public int CompareTo(object obj)
+		{
+			if (obj != null && obj.GetType() != GetType())
```
NicolasDorier: Please, do this 1 line:
```csharp
return obj is uint256 v ? CompareTo(v) : throw new ArgumentException($"Object is not an instance of uint256", nameof(obj));
```

NicolasDorier:
can you do for `uint160` as well?

NBitcoin/UInt2561.cs
ultimaweapon: The problems is it will violate documentation  https://docs.microsoft.com/en-us/dotnet/api/system.icomparable.compareto?view=netframework-4.8#System_IComparable_CompareTo_System_Object_

ultimaweapon:
Sure. Will do it on Monday.

NBitcoin/UInt2561.cs
NicolasDorier: how does it violate it?

ultimaweapon: Here is how it say:

```
The parameter, obj, must be the same type as the class or value type that implements this interface; otherwise, an ArgumentException is thrown.
```

NicolasDorier: Yes, and my code is checking this.

NicolasDorier: This handle the null case:
```csharp
return obj is uint256 v ? CompareTo(v) : 
       obj is null ? CompareTo(null as uint256) : throw new ArgumentException($"Object is not an instance of uint256", nameof(obj));
```

ultimaweapon: The problems is if `obj` is a derived type it will not throw. I have another idea. The idea is making `uint256` class as sealed. But this is a breaking change if someone already derived from it.

NicolasDorier: `uint256` can safely be sealed.

NicolasDorier: We can seal `uint256`, or just don't care about derived type. This is such a corner case...

ultimaweapon:
Updated.

PR #698, opened 2019-05-29
Change node filters to use a concurrent list instead of a dictionary

mikedennis:
This is a prerequisite for the work I'm doing in #692

ConcurrentDictionary does not maintain insertion order so it's usage in NodeFiltersCollection does not allow for ordering of Filters. 

ThreadSafeList uses a list instead of ConcurrentDictionary and the locking of the collection is done manually. Currently I've only replaced the usage in NodeFiltersCollection but I have run all the tests using ThreadSafeList isntead of ThreadSafeCollection in all occurrences and all the same tests pass. So we may be able to replace other instances as well.

ThreadSafeList will allow insertion of duplicates whereas ThreadSafeCollection will not so that is a key difference to be aware of beyond the ordering difference.

### DIFF
--- a/NBitcoin/Utils/ThreadSafeList.cs
+++ b/NBitcoin/Utils/ThreadSafeList.cs
@@ -96,16 +96,16 @@ namespace NBitcoin
 
 		public IEnumerator<T> GetEnumerator()
 		{
-			IEnumerator<T> enumerator = null;
-
-			lock (_lock)
+			IEnumerator<T> enumerator = _EnumeratorList?.GetEnumerator();
+			if (enumerator == null)
 			{
-				if (_EnumeratorList == null)
-					_EnumeratorList = _Behaviors.ToList();
-
-				enumerator = _EnumeratorList?.GetEnumerator();
+				lock (_lock)
+				{
+					var behaviorsList = _Behaviors.ToList();
+					_EnumeratorList = behaviorsList;
+					enumerator = behaviorsList.GetEnumerator();
+				}
 			}
-
 			return enumerator;
 		}
 
--- a/NBitcoin/Utils/ThreadSafeList.cs
+++ b/NBitcoin/Utils/ThreadSafeList.cs
@@ -96,14 +96,17 @@ namespace NBitcoin
 
 		public IEnumerator<T> GetEnumerator()
 		{
-			if (_EnumeratorList == null)
+			IEnumerator<T> enumerator = null;
+
+			lock (_lock)
 			{
-				lock (_lock)
-				{
+				if (_EnumeratorList == null)
 					_EnumeratorList = _Behaviors.ToList();
-				}
+
+				enumerator = _EnumeratorList?.GetEnumerator();
 			}
-			return _EnumeratorList?.GetEnumerator();
+
+			return enumerator;
 		}
 
 		#endregion
--- a/NBitcoin/Utils/ThreadSafeList.cs
+++ b/NBitcoin/Utils/ThreadSafeList.cs
@@ -1,0 +1,109 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NBitcoin
+{ 
+	public class ThreadSafeList<T> : IEnumerable<T>
+	{
+		private List<T> _Behaviors;
+		private object _lock = new object();
+
+		public ThreadSafeList()
+		{
+			lock (_lock)
+				_Behaviors = new List<T>();
+		}
+
+		/// <summary>
+		/// Add an item to the collection
+		/// </summary>
+		/// <param name="item"></param>
+		/// <returns>When disposed, the item is removed</returns>
+		public IDisposable Add(T item)
+		{
+			if (item == null)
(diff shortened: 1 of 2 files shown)


NBitcoin/Utils/ThreadSafeList.cs
```diff
@@ -0,0 +1,109 @@
...
+			{
+				Remove(b);
+			}
+		}
+
+		#region IEnumerable<T> Members
+
+		public IEnumerator<T> GetEnumerator()
+		{
+			return _Behaviors.ToList().GetEnumerator();
```
NicolasDorier: to be thread safe you need to lock Behavior, make a copy of the list, and returning the enumerator of this copy.

NBitcoin/Utils/ThreadSafeList.cs
```diff
@@ -0,0 +1,114 @@
...
+		#region IEnumerable<T> Members
+
+		public IEnumerator<T> GetEnumerator()
+		{
+			List<T> list = null;
+			lock (_lock)
+			{
+				list = _Behaviors.ToList();
+			}
+			return list?.GetEnumerator();
```
NicolasDorier: But then, it is a performance issue because we are duplicating a list everytimes a message come. :(

mikedennis: I think we were also doing that before (this is from ThreadSafeCollection):
```
		public IEnumerator<T> GetEnumerator()
		{
			return _Behaviors.Select(k => k.Key).GetEnumerator();
		}
```

mikedennis: Since we are doing all this copying anyways... wondering whether we might be able to use some of the new Immutable collections like [ImmutableList<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablelist-1?view=netcore-2.2), been dying to find a reason to try them out 😄

NicolasDorier: `_Behaviors.Select(k => k.Key).GetEnumerator()` is not a copy.

NicolasDorier:
An idea, is that you make an array that is a copy of `_Behaviors`. Everytimes you modify `_Behaviors`, you set the array to null.

mikedennis:
Updated to reduce list copying for enumerator as suggested. Reran the tests and they still pass.

NBitcoin/Utils/ThreadSafeList.cs
```diff
@@ -92,12 +96,14 @@ public T FindOrCreate<U>() where U : T, new()
...
+			if (_EnumeratorList == null)
 			{
-				list = _Behaviors.ToList();
+				lock (_lock)
+				{
+					_EnumeratorList = _Behaviors.ToList();
+				}
 			}
-			return list?.GetEnumerator();
+			return _EnumeratorList?.GetEnumerator();
```
NicolasDorier: Please capture `enumeratorList` in a lock variable so we are sure GetEnumerator never returns null.

NBitcoin/Utils/ThreadSafeList.cs
```diff
@@ -96,14 +96,17 @@ public T FindOrCreate<U>() where U : T, new()
...
-				{
+				if (_EnumeratorList == null)
 					_EnumeratorList = _Behaviors.ToList();
-				}
+
+				enumerator = _EnumeratorList?.GetEnumerator();
 			}
-			return _EnumeratorList?.GetEnumerator();
+
+			return enumerator;
```
NicolasDorier:
```csharp
IEnumerator<T> enumerator = _EnumeratorList?.GetEnumerator();
if (enumerator == null)
{
   lock (_lock)
   {
       _EnumeratorList = _Behaviors.ToList();
       enumerator =  _EnumeratorList.GetEnumerator();
   }
  return enumerator;
}
```

So you lock only if really needed.

NicolasDorier:
I am tempted to revert this. The build is very unstable since I merged this. `CanMaintainChainWithSlimChainBehavior`  is failing here and there, I have no idea why.

NicolasDorier:
Tried to refactor a bit, no luck.

mikedennis:
Sure no worries - go ahead and revert and I'll do some more testing with it with that test.

NicolasDorier:
I change back to `ThreadSafeCollection` for now.

NicolasDorier:
I don't understand the issue.

zeptin:
@mikedennis is `SynchronizedCollection<T>` unsuitable for what you were trying to do here?

mikedennis:
@zeptin I believe that's only .NET framework. Found the source code though, looks like it would be easy to port https://github.com/microsoft/referencesource/blob/3b1eaf5203992df69de44c783a3eda37d3d4cd10/System.ServiceModel/System/ServiceModel/SynchronizedCollection.cs

zeptin:
@NicolasDorier I am unable to get that test to fail locally while using the `ThreadSafeList`. Can we give that modification another try?

NicolasDorier:
yes, make a pr

PR #701, opened 2019-06-07
Make Script.GetDestination() support P2PK script

ultimaweapon:
This will make `Script.GetDestinationAddress()` success with P2PK script. The reason I need this because I need to get address from the early mining reward, which is P2PK instead of P2PKH (e.g. https://bchain.info/BTC/tx/0e3e2357e806b6cdb1f70b54c3a3a17b6714ee1f0e68bebb44a74b1efd512098).

### DIFF
--- a/NBitcoin/Script.cs
+++ b/NBitcoin/Script.cs
@@ -1,5 +1,4 @@
 ﻿using System.Runtime.InteropServices;
-using NBitcoin.Crypto;
 using NBitcoin.DataEncoders;
 using System;
 using System.Collections.Generic;
@@ -128,7 +127,7 @@ namespace NBitcoin
 		/// them to be valid. (but old blocks may not comply with) Currently just P2SH,
 		/// but in the future other flags may be added, such as a soft-fork to enforce
 		/// strict DER encoding.
-		/// 
+		///
 		/// Failing one of these tests may trigger a DoS ban - see CheckInputs() for
 		/// details.
 		/// </summary>
@@ -754,7 +753,7 @@ namespace NBitcoin
 		}
 
 		/// <summary>
-		/// Extract P2SH/P2PH/P2WSH/P2WPKH id from scriptPubKey
+		/// Extract P2PK/P2SH/P2PH/P2WSH/P2WPKH id from scriptPubKey
 		/// </summary>
 		/// <param name="network"></param>
 		/// <returns></returns>
@@ -767,7 +766,12 @@ namespace NBitcoin
 			if(scriptHashParams != null)
 				return scriptHashParams;
 			var wit = PayToWitTemplate.Instance.ExtractScriptPubKeyParameters(this);
-			return wit;
+			if(wit != null)
+				return wit;
+			var pubKeyParams = PayToPubkeyTemplate.Instance.ExtractScriptPubKeyParameters(this);
+			if(pubKeyParams != null)
+				return pubKeyParams.Hash;
+			return null;
 		}
 
 		/// <summary>
--- a/NBitcoin/Script.cs
+++ b/NBitcoin/Script.cs
@@ -408,6 +408,11 @@ namespace NBitcoin
 			return new Script(data, true, true);
 		}
 
+		public static Script FromHex(string hex)
+		{
+			return FromBytesUnsafe(Encoders.Hex.DecodeData(hex));
+		}
+
 		public Script(byte[] data)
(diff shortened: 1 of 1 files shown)


ultimaweapon:
Can you guys tell me which tests that fail on Travis? The logging is so hard to find which on is fail. Or at least give me how to locate failed tests.

NicolasDorier:
Search for `xUnit`.
Anyway NACK for this PR, because the block explorer is wrong a P2PK does not have address, only P2PKH are.

ultimaweapon:
I take a look at how `decoderawtransaction` works for P2PK and found out it create P2PKH from public key and use it to get the address. Do you mean the address it show in `decoderawtransaction` is not a valid address?

NicolasDorier:
Yes, this is a bug.
If you send to this address it will not give you a P2PK scriptPubKey.

NicolasDorier:
Hopefully I removed the bitcoind logs. (https://github.com/MetacoSA/NBitcoin/commit/19347c1116595a4c31c918b1b797e89a0a8f75beb)

NicolasDorier:
I just improved test feedback! https://travis-ci.org/MetacoSA/NBitcoin/jobs/542814000

ultimaweapon:
The commits is inaccessible for me. That does not look right for me since the address it show for P2PK output in `decoderawtransaction` will be effectively the same address as if that output is P2PKH since it get `KeyId` from the public key in the same way. That mean if somebody spend to that address the owner should be able to spend that output.

NicolasDorier:
> That mean if somebody spend to that address the owner should be able to spend that output.

By the same logic, `1 <pubkey> 1 MULTISIG` should also have the same address?

NicolasDorier:
Also, `script.GetDestinationAddress().ScriptPubKey == script` is a strong invariant that lot's of people depends on. Breaking this would break their code.

ultimaweapon:
I will take a look more on Monday. Thanks for information.

ultimaweapon:
I don't know about multi-sig but this PR will handle only P2PK script. Okay I agree with your last comment.

NicolasDorier:
My point is that if you consider:

> That mean if somebody spend to that address the owner should be able to spend that output.

To be a good argument, then it means that any scriptPubKeys which satisfy the same condition should show a bitcoin address. However, this is impossible to make code which can detect that condition automatically for any arbitrary scriptPubKey.

ultimaweapon:
What do you mean by that? Current `GetDestination` support only P2SH, P2PKH and P2W so when the script is multi-sign it should return null. With this PR is will also support P2PK but not for multi-sig. But the problems is it might unexpected by user that when they call `GetDestinationAddress` for P2PK script but they got P2PKH address (and P2PKH script from `ScriptPubKey` of that address).

NicolasDorier:
Yes, that's the problem. My point is that if you add support for P2PK why not adding support for `1 A 1 MULTISIG` script as well?

ultimaweapon:
I will take a look more for solutions when I have time. Thanks for discussion.

NicolasDorier:
@ultimaweapon no no don't do it, you can't find solution. There is an infinity number of script that can be signed by the same pubkey. This was just to explain you that your argument for adding P2PK support can't hold.

ultimaweapon:
What about adding a new property to tell if script is P2PK and let the caller do the rest?

NicolasDorier:
You can already do this

```csharp
PubKey key = NBitcoin.PayToPubkeyTemplate.Instance.ExtractScriptPubKeyParameters(scriptPubKey);
if (key != null) // This is PK
{ }
```

But seriously, don't spend time on this P2PK is not used anymore by anybody.

ultimaweapon:
Yes I can do that but would it be better to have `IsPayToPubKey` due to it already have `IsPayToScriptHash` and `IsWitness`? It may not useful if using NBitcoin just for creating transactions but it useful when someone want to parse legacy blocks.

NicolasDorier:
@ultimaweapon I think this is a good idea. 
However, what about this instead:
* Make `IsPayToScriptHash` and `IsWitness` obsolete
* Add new `Script.IsScriptType(ScriptType)` where `ScriptType` is an enum with `Witness`, `P2PKH`, `P2SH`, `P2PK`, `P2WPKH`, `P2WSH`, `MultiSig`?

So it is easy in the future to add new types.

ultimaweapon:
Look good to me. I will do it. Should done within this week.

PR #703, opened 2019-06-13
Remove Andreas Schildbach's Testnet DNS Seed

igorgue:
Remove DNS seed that's empty and owner claims he does not run it anymore.

Source: "Indeed, I'm not running it any more." [0]

[0] https://twitter.com/schildbach/status/1138793353749946369

### DIFF
--- a/NBitcoin/Network.cs
+++ b/NBitcoin/Network.cs
@@ -2201,7 +2201,9 @@ namespace NBitcoin
 #if !NOSOCKET
 			vFixedSeeds.Clear();
 			vSeeds.Clear();
-			vSeeds.Add(new DNSSeedData("bitcoin.petertodd.org", "testnet-seed.bitcoin.petertodd.org"));
+			vSeeds.Add(new DNSSeedData("bitcoin.jonasschnelli.ch", "testnet-seed.bitcoin.jonasschnelli.ch"));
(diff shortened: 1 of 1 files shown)


igorgue:
Not sure why that test failed on travis only. But anyways, how can I modify the DNS seed list without changing the source code?

Would it be possible to expose that? I broke my brain trying to do this.

Help @NicolasDorier ? :)

lontivero:
I think your change is okay and the test failed because it is not deterministic.

igorgue:
@lontivero out of curiosity I was checking Bitcoin's code and the list is this one:

```cpp
        // nodes with support for servicebits filtering should be at the top
        vSeeds.emplace_back("testnet-seed.bitcoin.jonasschnelli.ch");
        vSeeds.emplace_back("seed.tbtc.petertodd.org");
        vSeeds.emplace_back("seed.testnet.bitcoin.sprovoost.nl");
        vSeeds.emplace_back("testnet-seed.bluematt.me"); // Just a static list of stable node(s), only supports x9
``` [0]

Maybe NBitcoin should have the same one?

[0] https://github.com/bitcoin/bitcoin/blob/f5a623eb66c81d9d7b11206d574430af0127546d/src/chainparams.cpp#L232-L236

So the list is a little bit different...

NicolasDorier:
We should have the same as latest bitcoin release.

PR #707, opened 2019-06-21
Fix for RPC getrawtransaction to avoid null exception

sipsorcery:
For derivatives that don't support segwit there is no `hash` field returned in the `getrawtransaction` JSON response.


nopara73:
# Syntax Nitpicking

You read the hash value twice. You can avoid it for example like this: 

```cs
var hashValue = json["hash"];
Hash = hashValue is null ? null : uint256.Parse(hashValue),
```

# Concept Discussion

Should that even be null though? As I see `Parse` is used and not `TryParse` to intentionally throw the exception. In what circumstance did you encounter this exception? `getrawtransaction` not returning hash should not happen I think.

sipsorcery:
I replicated the null check approach from the surrounding code. Probably too small a change to justify breaking the style.

The field is null in Dash.

The `getrawtransaction` help text for bitcoin-core:

```
 "hash" : "id",        (string) The transaction hash (differs from txid for witness transactions)
```

Dash doesn't support segwit and the help text for the same command does not list `hash` as a returned field.

NicolasDorier:
You are both dinosaurs, one liner of modern C#

```csharp
Hash = json["hash"] is JToken token ? uint256.Parse(token.Value<string>()) : null,
```

PR #710, opened 2019-06-24
Updated test framework downloads + added new node builder option

sipsorcery:
Updated node software downloads for:
 - Litecoin, 
 - BCash, 
 - Liquid,
 - Dash. Left the Dash update commented out. Dash have recently removed mining functionality from their official binaries to avoid false positives from Anti-Virus software. This means regtest mode is now useless since it's not possible to generate blocks.

Added extra setting to the test framework node builder so that nodes can optionally be run in a dedicated console window with log messages.

### DIFF
--- a/NBitcoin.TestFramework/WellknownNodeDownloadData.cs
+++ b/NBitcoin.TestFramework/WellknownNodeDownloadData.cs
@@ -193,6 +193,33 @@ namespace NBitcoin.Tests
 					Hash = "fe1a593ffb10fec817157903a49d8965c49594dda4021fb76cf7d341e5300e17"
 				}
 			};
+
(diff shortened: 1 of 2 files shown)


NBitcoin.TestFramework/WellknownNodeDownloadData.cs
```diff
@@ -338,7 +392,36 @@ public class DashNodeDownloadData
 					Executable = "dashcore-0.13.0/bin/dashd",
 					Hash = "6f97f502732e5b63a431d0edb5a9d14e95ff8afb8e7eb94463566a75e7589a70"
 				}
-			};
+			};
+
+			// Note that Dash have DISABLED mining in their offical Windows and Mac binaries as per
```
NicolasDorier: **facepalm**

PR #711, opened 2019-06-26
Fix err on parse Money value from response

reximus:


NicolasDorier:
I am surprised, in which case what existed before does not work?

reximus:
in case json response:
`{"result":0,"error":null,"id":1}`
will get ex: Cannot parse string to money

PR #721, opened 2019-08-06
Add support for Energi

babgev:


NicolasDorier:
CI not passing

babgev:
> CI not passing

But the error it shows is "error NETSDK1045: The current .NET SDK does not support targeting .NET Core 3.0.  Either target .NET Core 2.1 or lower, or use a version of the .NET SDK that supports .NET Core 3.0.". I didn't change the SDK version in my branch.

nopara73:
Something may be off with CI globally. Our CI stopped passing yesterday with similar strangeness, too. Of course it can be unrelated.

PR #722, opened 2019-08-07
Updated the regtest Liquid asset ID

sipsorcery:
Since Liquid has switched back to Elements as the code repository the chain and asset ID's have changed for regtest.

See readme on https://github.com/Blockstream/liquid.

### DIFF
--- a/NBitcoin.Altcoins/Liquid.cs
+++ b/NBitcoin.Altcoins/Liquid.cs
@@ -12,7 +12,7 @@ namespace NBitcoin.Altcoins
 		static Liquid()
 		{
 			ElementsParams<Liquid>.PeggedAssetId = new uint256("6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d");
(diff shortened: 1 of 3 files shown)


NicolasDorier:
@sipsorcery so we need to update liquid version, because if you do only this change, it will become impossible to dev on top of TestFramework for liquid.

sipsorcery:
That PR is coming as well :).

NicolasDorier:
Well bundle it in the same PR then so I can test.

NicolasDorier:
(and rebase, I fixed the test)

NicolasDorier:
@sipsorcery something still wrong?

sipsorcery:
@NicolasDorier no I think the PR is good to go except for the failing travis test which I can't reproduce. The integration tests I previously ran against the liquid 3.14.1.23 daemon are now passing against the elements 0.17.0.1 daemon.

NicolasDorier:
oh they made an official windows version awesome!

PR #725, opened 2019-08-12
Add testcase for ecdsa serilization

joemphilips:
In Lightning Network, signature serialization is done by non-DER format, that is
 R value(32 bytes) + S Value(32 bytes)

I've been doing this serialization by creating wrapper class for ECDSASignature, and it seems to work fine. But only when the first byte of an R is 0xff, it fails to deserialize properly. (found by Property based testing.)
So I have reproduced the problem in this PR for begging someone's help.

(I've been trying to fix it by myself but BigInteger's `ToByteArray` method is too cryptic :()

I will really appreciate if someone tells me why this test is failing and if this is an intended behavior or not.

### DIFF
--- a/NBitcoin.Tests/ECDSATests.cs
+++ b/NBitcoin.Tests/ECDSATests.cs
@@ -1,0 +1,26 @@
+using Xunit;
+using NBitcoin;
+using NBitcoin.Tests.Helpers;
+
+namespace NBitcoin.Tests
+{
+	public class ECDSATests
+	{
+		[Fact]
+		[Trait("UnitTest", "UnitTest")]
+		public void ECDSASerializationTests()
+		{
+			var rBytes = new byte[] { 255, 243, 22, 18, 30, 119, 145, 145, 181, 246, 229, 69,
+			127, 129, 220, 20, 146, 135, 213, 55, 116, 193, 77,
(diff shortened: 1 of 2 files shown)


lontivero:
@joemphilips if the BigInteger.ToArray() is 33 byte length that means the number is negative and then it needs an extra byte for the sign. You can try to create the instances specifying the sign explicitely as follow:

```c#
var r = new BouncyCastle.Math.BigInteger(1, rBytes);
```

If that is not what you want, remember you can control how to serialize bitintegers with `ToByteArray` (with or without sign byte) and  `ToByteArrayUnsigned` (without sign byte)

lontivero:
Oh sorry, i answered without reading carefully but in any case the answer is still valid, the byte array starts with 0xff because it is negative. Try using the constructor that allows you to pass the sign.

joemphilips:
> Try using the constructor that allows you to pass the sign.

Thank you. @lontivero  This solved the problem perfectly. I still don't understand why it happens only in the case of 0xff, but I guess this is a problem on my side that I don't understand how BigInteger works pretty well. I should RTFM.

NicolasDorier:
the byte serialisation of biginteger is easy to shoot yourself in the foot :(

PR #730, opened 2019-08-19
Fix formatting (NBitcoin.TestFramework project)

yahiheb:


NBitcoin.TestFramework/NodeBuilder.cs
```diff
@@ -19,677 +19,677 @@
...
-
-		private void CleanFolder()
-		{
-			try
-			{
-				Directory.Delete(_Folder, true);
-			}
-			catch(DirectoryNotFoundException) { }
-		}
+{
```
NicolasDorier: should be tabs

yahiheb: It was supposed to be tabs. Anyway I fixed that.

PR #734, opened 2019-08-20
Fix formatting (NBitcoin.Altcoins)

yahiheb:


nopara73:
ReYahihebSharper

NicolasDorier:
this one is tricky for me to review. (can't just check it is only whitespaces) and tests does not pass.

yahiheb:
@nopara73 lol but I really like to see the code very clean :p.
@NicolasDorier I agree.
Should I break it into smaller PRs (6 or 7 files per PR) but they might still be big?
Or to avoid the review you can do it yourself. If you are using Visual Studio 2019, use `Alt + Enter` (Quick Actions)  on a line containing for ex `if(` and choose `Fix formatting` and then `Fix all occurrences in: Solution`.

NicolasDorier:
Closing this one for inactivity. I don't think we have to format, we can do as we go. However, if we do, you can break in smaller pieces.

PR #738, opened 2019-09-05
Fix typo on obsolete call

igorgue:
Also, would like to ask why is this obsolete, I know you can always do:

```csharp
Network.Parse<BitcoinAddress>(addr.ToString(), null)
```

But to me also `BitcoinAddress.Create(str)` makes sense as well.

### DIFF
--- a/NBitcoin/BitcoinAddress.cs
+++ b/NBitcoin/BitcoinAddress.cs
@@ -106,7 +106,7 @@ namespace NBitcoin
 		/// <param name="str">The string to parse</param>
 		/// <returns>A BitcoinAddress or BitcoinScriptAddress</returns>
 		/// <exception cref="System.FormatException">Invalid format</exception>
-		[Obsolete("Use BitcoinCreate(string, Network) instead")]
+		[Obsolete("Use BitcoinAddress.Create(string, Network) instead")]
 		public static BitcoinAddress Create(string str)
 		{
 			if (str == null)


igorgue:
A great example of this use is serializing addresses to JSON, e.g.:

```csharp
using System;

using NBitcoin;

using Newtonsoft.Json;

namespace Utilities.JsonConverters
{
    public class BitcoinAddressConverter : JsonConverter
    {
        public override bool CanConvert(Type objectType)
        {
            return objectType == typeof(BitcoinAddress);
        }

        /// <inheritdoc />
        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
        {
            return Network.Parse((string)reader.Value, null);
        }

        /// <inheritdoc />
        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
        {
            writer.WriteValue(((BitcoinAddress)value).ToString());
        }
    }
}
```

NicolasDorier:
@igorgue it is obsolete because it will lead to bug as you can't assume shitcoins are using different prefixes. For example, monacoin has the same prefix for addresses than litecoin, so `Network.Parse( xxx, null)` will returns a `BitcoinAddress` with `Network` set to litecoin, when you expect monacoin or vis versa.

Note that NBitcoin already has a bunch of json converters in `NBitcoin.JsonConverters`. You can register it to a `JsonSerializerSettings` with `NBitcoin.JsonConverters.Serializer.RegisterFrontConverters(jsonSerializerSettings, network)`.

igorgue:
Fucking shitcoins! Yeah makes sense now, I only work with Bitcoin though I'll try the settings cause at that point I should know the Network the user is in and will make things faster.

nopara73:
This was probably the most painful NBitcoin upgrade for years (which tells you how stable this library is) to make sure that network is always available at address parsing.

NicolasDorier:
@igorgue even if you work only on Bitcoin you need to properly do it, so you don't get bug on Testnet address being mistaken for Regtest or vis versa.

PR #741, opened 2019-09-11
Improve comparison and equality

knocte:
/cc @joemphilips

### DIFF
--- a/NBitcoin/PubKey.cs
+++ b/NBitcoin/PubKey.cs
@@ -29,7 +29,7 @@ namespace NBitcoin
 		/// </summary>
 		SegwitP2SH
 	}
-	public class PubKey : IBitcoinSerializable, IDestination, IComparable<PubKey>, IEquatable<PubKey>
+	public class PubKey : IBitcoinSerializable, IDestination, IComparable<PubKey>, IEquatable<PubKey>, IComparable
 	{
 		/// <summary>
 		/// Create a new Public key from string
@@ -89,7 +89,7 @@ namespace NBitcoin
 		}
 
 
-		public int CompareTo(PubKey other) => BytesComparer.Instance.Compare(this.ToBytes(), other.ToBytes());
+		public int CompareTo(PubKey other) => other == null ? 1 : BytesComparer.Instance.Compare(this.ToBytes(), other.ToBytes());
 
 		public PubKey Compress()
 		{
@@ -552,6 +552,11 @@ namespace NBitcoin
 		public BitcoinWitPubKeyAddress GetSegwitAddress(Network network)
 		{
 			return new BitcoinWitPubKeyAddress(WitHash, network);
+		}
+
+		public int CompareTo(object obj)
+		{
+			return this.CompareTo(obj as PubKey);
 		}
 
 		#endregion
--- a/NBitcoin/Transaction.cs
+++ b/NBitcoin/Transaction.cs
@@ -9,7 +9,7 @@ using System.Linq;
 
 namespace NBitcoin
 {
-	public class OutPoint : IBitcoinSerializable
+	public class OutPoint : IBitcoinSerializable, IComparable, IComparable<OutPoint>, IEquatable<OutPoint>
 	{
 		public bool IsNull
 		{
@@ -180,13 +180,6 @@ namespace NBitcoin
 		{
 			return !(a == b);
 		}
-		public override bool Equals(object obj)
-		{
-			OutPoint item = obj as OutPoint;
-			if (object.ReferenceEquals(null, item))
-				return false;
-			return item == this;
-		}
 
 		public override int GetHashCode()
 		{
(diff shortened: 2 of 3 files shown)


nopara73:
@lontivero would you mind to review it? Smells a bit dangerous.

NBitcoin/UInt2561.cs
```diff
@@ -1,11 +1,13 @@
 ﻿
```
joemphilips: For uint256 and uint160, this branch and master have both implemented IEquatable and IComparable,
both seems to work fine. But lets just have minimum change here. So would you remove this file from this PR? @knocte 

It should be no functional difference.

knocte: done

NBitcoin/PubKey.cs
```diff
@@ -29,7 +29,7 @@ public enum ScriptPubKeyType
 		/// </summary>
 		SegwitP2SH
 	}
-	public class PubKey : IBitcoinSerializable, IDestination, IComparable<PubKey>, IEquatable<PubKey>
+	public class PubKey : IBitcoinSerializable, IDestination, IComparable<PubKey>, IEquatable<PubKey>, IComparable
```
nopara73: `IComparable` makes no sense for `PubKey`.

NBitcoin/Transaction.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace NBitcoin
 {
-	public class OutPoint : IBitcoinSerializable
+	public class OutPoint : IBitcoinSerializable, IComparable, IComparable<OutPoint>, IEquatable<OutPoint>
```
nopara73: `IComparable` makes no sense for `OutPoint` outpoint, unless you want to implement @kristovatla's BIP69: https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki

NBitcoin/Transaction.cs
```diff
@@ -200,10 +193,32 @@ public override string ToString()
...
+			if (other is null)
+				return false;
+			return other == this;
+		}
+
 	}
 
 
-	public class TxIn : IBitcoinSerializable
+	public class TxIn : IBitcoinSerializable, IEquatable<TxIn>
```
nopara73: `IComparable` makes no sense for `TxIn` outpoint, unless you want to implement @kristovatla's BIP69: https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki

NBitcoin/Transaction.cs
joemphilips: I implemented these interface just for use of convenience on my code. (To use in F# collection types) And the actual logic does not really matter. So yes, I think modifying these methods to follow BIP69 is the right way to go.

NBitcoin/PubKey.cs
joemphilips: I think it makes sense to follow [bip67](https://github.com/bitcoin/bips/blob/master/bip-0067.mediawiki)

joemphilips:
@knocte Would you mind adding test vectors from bip67 and bip69 ?
Unless someone have an objection to make these objects comparable in the first place.

knocte:
> @knocte Would you mind adding test vectors from bip67 and bip69 ?

Well, if comparison is something that helps implementation of some BIPs, I would say this change has value in and of itself. Now, if someone were to implement those BIPs in particular and finds issues, he can always write tests that cover those issues, right? What I mean is, can we just mark the Compare() methods with a //TODO or a //FIXME comment saying "add tests"? It's not that I'm lazy, it's that I don't want to get deep in yak-shaving, you know. My main motivation is that DotNetLightning stops using a fork of NBitcoin, so that consumers of DotNetLightning which were already consumers of upstream NBitcoin don't find assembly-loading conflicts.

NBitcoin/PubKey.cs
nopara73: ACK.

lontivero:
Concept NACK. 
It assumes that `OutPoint` and `TxIn` are comparable and that there is only one official/correct way to do it, I mean, it assumes they should be always compared in the way that bip67 and bip69 specifies, what makes sense only in one very specific use case (the one @joemphilips needs for his project).

NBitcoin/Transaction.cs
NicolasDorier: NACK on this one, `TxIn` can be subclassed by shitcoins so you can't assume you know equality.

joemphilips:
@knocte  You can close this now. I made a wrapper class on my side.

FYI, I don't really agree with @lontivero  with this one. I think having default ordering for PubKey TxOut, and OutPoint is important for anonymity (I agree with Nicolas about TxIn though) . Especially when not every devs are aware of the existence of those BIPs. And developers can define custom comparator anytime when they need other way of sorting.

But anyway, it is not a big deal for me. someone can create another PR if he thinks this is necessary.

PR #742, opened 2019-09-11
Key: allow specifying forceLowR for Sign*() methods

knocte:
/cc @joemphilips


nopara73:
Would you mind to provide more context?

joemphilips:
I use this in unit test for my lightning network daemon.
BOLT assumes RFC6979 signer for its test vectors.

joemphilips:
To share little more context, This and #741 are from my own fork of NBitcoin.
My plan was to make a PR after finishing all my work, but me and @knocte thought things get easy if it could be merged into master now (or at least to have a feedback ).

NBitcoin/Key.cs
```diff
@@ -100,21 +100,21 @@ public ECDSASignature Sign(uint256 hash)
 			return _ECKey.Sign(hash, true);
 		}
 
-		public string SignMessage(String message)
+		public string SignMessage(String message, bool forceLowR = true)
```
lontivero: The new optional parameter is not used.

knocte: fixed

knocte:
> Just an small change is required.

Done.

@NicolasDorier can you review/merge?

NicolasDorier:
@knocte instead of having an optional parameter, can you instead make an overload?

Adding optional parameter is a breaking change that ask libraries which depends on NBitcoin to be rebuilt, while adding an overload will make the new NBitcoin library just work without recompiling.

PR #745, opened 2019-09-22
CoinSelector infinite loop

lontivero:
This PR is for demo a corner case where an `ICoinSelector` makes NBitcoin to enter in an infinite loop. If well the problem is not in NBitcoin library, I share this here because I would like to understand better which are the invariants/contract that  a  custom CoinSelector should respect in order to avoid situations like this. 

For context: https://github.com/zkSNACKs/WalletWasabi/issues/2304

### DIFF
--- a/NBitcoin.Tests/transaction_tests.cs
+++ b/NBitcoin.Tests/transaction_tests.cs
@@ -237,6 +237,63 @@ namespace NBitcoin.Tests
 			});
 		}
 
+
+		class BrokenCoinSelector : ICoinSelector
+		{
+			private Dictionary<OutPoint, Coin> CoinsByOutpoint { get; }
+
+			public BrokenCoinSelector(IEnumerable<Coin> coins)
+			{
+				CoinsByOutpoint = coins.ToDictionary(x=>x.Outpoint);
+			}
+
+			public IEnumerable<ICoin> Select(IEnumerable<ICoin> coins, IMoney target)
+			{
+				var totalOutAmount = (Money)target;
+				var unspentCoins = coins.Select(c => CoinsByOutpoint[c.Outpoint]).ToArray();
+				var coinsToSpend = new HashSet<Coin>();
+
+				foreach (var coin in unspentCoins.OrderByDescending(x => x.Amount))
+				{
+					coinsToSpend.Add(coin);
+					if (coinsToSpend.Select(x => x.Amount).Sum() >= totalOutAmount)
+					{
+						// Add if we can find address reuse.
+						foreach (var c in unspentCoins
+							.Except(coinsToSpend) // So we're choosing from the non selected coins.
+							.Where(x => coinsToSpend.Any(y => y.ScriptPubKey == x.ScriptPubKey)))// Where the selected coins contains the same script.
+						{
+							coinsToSpend.Add(c);
+						}
+
+						break;
+					}
+				}
+				return coinsToSpend;
+			}
+		}
+
+		[Fact]
+		[Trait("UnitTest", "UnitTest")]
+		public void RunForever()
+		{
+			var network = Network.TestNet;
+			var scriptPubKey = new Key().ScriptPubKey;
+			var amount = Money.Coins(0.01m);
+			var inputs = new[]{
+				(new Coin(RandOutpoint(), new TxOut(amount, scriptPubKey ))),
+				(new Coin(RandOutpoint(), new TxOut(amount, scriptPubKey))),
+				(new Coin(RandOutpoint(), new TxOut(amount, new Key().ScriptPubKey ))),  // this is different script.
+				(new Coin(RandOutpoint(), new TxOut(amount, scriptPubKey)))};
+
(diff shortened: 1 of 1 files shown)


lontivero:
I was debugging this test and this is what I get:

Let imagine we have 3 coins: `(scriptPubKey: e5fab3..., amount: 1.0)`, `(scriptPubKey: e5fab3..., amount: 0.5)` and `(scriptPubKey: bac79d..., amount: 0.1)`. Note that there are two coins with the same scriptPubKey.
We want to send 1.4btc to a given address using out `FancifulCoinSelector` logic to always spend coins with the same scriptPubKey. Our `FancifulCoinSelector` class is going to select them all (1.51btc) but NBitcoin realizes that the resulting change (0.11btc) is bigger than one of the coins and then it calls the `FancifulCoinSelector` algorithm again with a corrected `target` with the hope it selects less coins but our coins selection algo insists with the same set of coins again and again so, it never ends.

It seems NBitcoin doesn't support any arbitrary coin selection strategy, instead it expects some kind of _optimous_ selection.  If that is the case I think coin selectors like our `FancifulCoinSelector` are not possible. Could you @NicolasDorier confirm this, please?

yahiheb:
`(scriptPubKey: bac79d..., amount: 0.1)` --> `(scriptPubKey: bac79d..., amount: 0.01)`

NicolasDorier:
Why did you closed that?

> NBitcoin realizes that the resulting change (0.11btc) is bigger than one of the coins and then it calls the FancifulCoinSelector algorithm again with a corrected target with the hope it selects less coins but our coins selection algo insists with the same set of coins again and again so, it never ends.

I don't remember coding any of that.

nopara73:
He closed it because we refactored Wasabi so NBitcoin won't call it forever anymore. But I agree, this should be fixed in NBitcoin too.

NicolasDorier:
Checking what is going on.

NicolasDorier:
Ok so the problem imaging you have three coins 1.0, 2.0, 3.0 for a total of 6 BTC.

If you want to send 5.5, I will call the coin selector, to target 5.5 BTC.
If I detect you selected 6.0 coin, then there is 0.5 of change back. 

Because this change amount is more than dust, I decide to create a new transaction with a change output of 0.5. (this might update the fees because the transaction is now bigger)

So I retry the same process with one change output. But this time, I am asking a target of 6 BTC to the coin selector. Normally, this should be an exact match with the target this time.... except if it is not the case because of the coin selector!

------------

So here is what happen with your repro. You have 4 coins of 0.01 (total 0.04) and send 0.015 somewhere.

The first pass you select 3 coins (0.03) to target 0.015 BTC (0 BTC of change + 0.015 to destination), so I set the change amount to 0.015 and make a second pass.

The second pass you select 4 coins (0.04) to target 0.03 BTC (0.015 BTC of change + 0.015 to destination), so I set the change amount to 0.01 and make a third pass.

The third pass you select 4 coins (0.04) to target 0.025 BTC (0.01 BTC of change + 0.015 to destination), so I set the change amount to 0.015 BTC

The fourth path you select 4 coins (0.04) to target 0.03  BTC (0.015 of change + 0.015 to destination), so I set the change to 0.1 BTC.... and then now back to the second pass step.

All in all, I think this is a bug of NBitcoin, I will think how to solve this, thanks for the repro!

NicolasDorier:
Ping @nopara73 and @lontivero please wait a moment, I publish new version of NBitcoin soon with a fix. The problem have nothing to do with address reuse, so your bug might happen again, even if you fixed the specific issue in Wasabi.

NicolasDorier:
For information, I fixed it by making sure that I don't ask for the CoinSelector to get the coins again.

NicolasDorier:
Fixed in 4.2.12 @lontivero @nopara73

nopara73:
@NicolasDorier Did you do something with your logging? We don't even use it yet it complains after NBitcoin update:

image

nopara73:
So you updated to .NET Core 3 and that conflicts with a dependency that's being dragged in with some large .net or testing sdks. The recommended solution throws warnings. I guess the only way to cleanly update NBitcoin without the warnings is to also update to .NET Core 3.0

NicolasDorier:
@nopara73 I did but I did not expected conflict, I will revert my change.

NicolasDorier:
I referenced the library in 3.0 because it is supported by .net standard 2.0, so .net core 2.1 by extension.

https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/3.0.0

NicolasDorier:
But that was stupid, reverting this.

NicolasDorier:
published 4.2.13.

My motivation was that 3.0.0 has no dependency while 1.0.0 has a bunch of them, see https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/1.0.0

nopara73:
Thanks, we're updating 4 things in parallel and we're in an update hell.. so this is very good for us.

PR #753, opened 2019-10-18
Add ConsensusFactory.CreateTxIn()

ultimaweapon:
Seem like it already use `TryCreateNew` to create `TxIn` but there is no `CreateTxIn` like the other so the derived class need to override `TryCreateNew` to be able to create a derived `TxIn`.


ultimaweapon:
Seem like my Git change file's EOL. Working on it now.

ultimaweapon:
Ready for review. Not sure why AppVeyor failed. The error is not related to my changes.

PR #756, opened 2019-10-20
Implement GolombRiceFilter Equality

nopara73:
@lontivero I used the `Data` byte array for uniqueness. That's sufficient right?

@NicolasDorier I based it on this, so it's battle tested:

https://github.com/zkSNACKs/WalletWasabi/blob/eaaef0c763200e01da0cf7de39ed74862704ffc6/WalletWasabi/Bases/ByteArraySerializableBase.cs#L55-L96

### DIFF
--- a/NBitcoin/BIP158/GolombRiceFilter.cs
+++ b/NBitcoin/BIP158/GolombRiceFilter.cs
@@ -322,9 +322,9 @@ namespace NBitcoin
 
 		public static bool operator ==(GolombRiceFilter x, GolombRiceFilter y)
 		{
-			if(x is null)
+			if (x is null)
 			{
-				if(y is null)
+				if (y is null)
 				{
 					return true;
 				}
@@ -333,7 +333,7 @@ namespace NBitcoin
 					return false;
 				}
 			}
-			else if(y is null)
+			else if (y is null)
 			{
 				return false;
 			}
--- a/NBitcoin/BIP158/GolombRiceFilter.cs
+++ b/NBitcoin/BIP158/GolombRiceFilter.cs
@@ -314,10 +314,37 @@ namespace NBitcoin
 				hash ^= Data[i];
 			}
 
+			hash ^= P.GetHashCode();
+			hash ^= M.GetHashCode();
+
 			return hash;
 		}
 
-		public static bool operator ==(GolombRiceFilter x, GolombRiceFilter y) => Utils.ArrayEqual(y?.Data, x?.Data);
+		public static bool operator ==(GolombRiceFilter x, GolombRiceFilter y)
+		{
+			if(x is null)
+			{
+				if(y is null)
+				{
+					return true;
+				}
+				else
+				{
+					return false;
+				}
+			}
+			else if(y is null)
+			{
+				return false;
+			}
+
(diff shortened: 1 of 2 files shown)


NBitcoin/BIP158/GolombRiceFilter.cs
```diff
@@ -290,6 +290,38 @@ private static ulong SipHash(ulong k0, ulong k1, byte[] data)
 			hasher.Write(data);
 			return hasher.Finalize();
 		}
+
+		#region EqualityAndComparison
+
+		public override bool Equals(object obj) => obj is GolombRiceFilter tx && this == tx;
```
lontivero: The variable name for filter shouldn't be `tx` but `filter` or any other better name.

nopara73: fixed

nopara73:
@lontivero Which parameters exactly?

lontivero:
The parameters are `P` and `M`. Also, before comparing `Data` you could compare `N` (number of elements in the filter). So, if `P`, `M` and `N` are equal in both filters then it makes sense to compare the `Data` array.

nopara73:
Done.

NicolasDorier:
Why equality makes sense here? Why do you need it?

nopara73:
I'm implementing something new (that honestly may never get into production) and I have equality implemented in that class, where that class has a `GolombRiceFilter` so right now I'm just ToStringing and compare equality like that.

NicolasDorier:
I think it is counterintuitive to have an overridden equality for this class. 
This is not a value object unlike say outpoint. So I'd like to understand if there is a real use case for this or it is just a "nice to have" for some corner case reason.

nopara73:
What do you mean? It is a value class. It maps to a filter.

NicolasDorier:
Typically a value object map to small set of data and could potentially be a `struct` instead of a class. `Money` or `Outpoint` maps well to this concept.

Filter, not so much. What usecase do you need it, I can't see any reason you would need that.

nopara73:
I'm writing a test and I'm asserting the serialized filter and the deserialized filter is the same.

> Filter, not so much.

Sure it could. It's not a builder or something. Maybe what confuses you is that you think about "filter" as a verb? While it's used as a noun in this context.

NicolasDorier:
> I'm writing a test and I'm asserting the serialized filter and the deserialized filter is the same.

Is this for the only reason? Because if that was the only reason, there would be no reason to not do the same for virtually any type, like `Transaction`.

The equality in a type should be limited to what can't have any possible doubt we are comparing values and not reference. `ScriptPubKey`, `Outpoint`, `uint256` or `KeyPath` are obvious, you very often use uniqueness based on their value only and don't care about the reference. But for any other type, it is far from being obvious.

Another idea could be to have a `Filter.GetHash()` whose value is obviously comparing values.
But are you sure you want this? If it is only for a unit test, this seems pretty pointless.

If it is for a protocol, I am quite confident comparing filter equality is the wrong approach and there is a better one, so that's why I want to know why you need it.

nopara73:
I see. I just wanted this, because comparing strings looked stupid. `.GetHash` is a good idea, I'll do that when I continue working on that branch.

PR #759, opened 2019-10-24
Liquid updates 

Kukks:
Updated Liquid Support

* Set correct hrps for liquid networks
* Add Blech32 encoder for elements segwit blinded addresses
* Allow alt networks to have more bech32 encoders configured than btc
* Added additional RPC commands for elements 
* Update WellKnownNode versions for elements and bitcoin
* Add additional serialization methods to BitcoinStream
* Adjust altcoin tests  for elements differences
* Add support for dynamic federation mode blocks in elements
* Adjust Elements regtest genesis hash


NBitcoin.Altcoins/Elements/ElementsBlock.cs
```diff
@@ -75,24 +83,26 @@ public override void ReadWrite(BitcoinStream stream)
...
+				isDynamic = version < 0;
+				nVersion = ~DYNAFED_HF_MASK & version;
 
 			}
 			else
 			{
-				nVersion = nVersion;
-				if (DynaFedParams == null || (DynaFedParams.Current == null && DynaFedParams.Proposed == null))
+				version = nVersion;
+				if (!DynaFedParams?.IsNull??true)
```
NicolasDorier: this is unreadable `DynaFedParams is true`

NicolasDorier:
does not build

NBitcoin.Tests/AltcoinTests.cs
```diff
@@ -73,7 +77,7 @@ public void HasCorrectGenesisBlock()
 				var rpc = builder.CreateNode().CreateRPCClient();
 				builder.StartAll();
 				var genesis = rpc.GetBlock(0);
-				if (builder.Network == Altcoins.Liquid.Instance.Regtest)
+				if (new[] {Liquid.Instance.Mainnet, Liquid.Instance.Testnet, Liquid.Instance.Regtest,}.Contains(rpc.Network))
```
NicolasDorier: or just use `rpc.Network.NetworkSet.CryptoCode`

NBitcoin.Tests/AltcoinTests.cs
```diff
@@ -186,14 +227,22 @@ public void CanParseAddress()
...
-		public void DoesRPCCapabilitiesWellAdvertised()
+		public async Task DoesRPCCapabilitiesWellAdvertised()
 		{
 			using (var builder = NodeBuilderEx.Create())
 			{
 				var node = builder.CreateNode();
 				builder.StartAll();
-				node.Generate(builder.Network.Consensus.CoinbaseMaturity + 1);
 				var rpc = node.CreateRPCClient();
+				if (new[] {Liquid.Instance.Mainnet, Liquid.Instance.Testnet, Liquid.Instance.Regtest}.Contains(node.Network))
```
NicolasDorier: nop, don't do this. Because all tests depends on `GenerateAsync`, if `DoesRPCCapabilitiesWellAdvertised` is well implemented, it should use `GenerateToAddress` under the hood.

NBitcoin.Tests/AltcoinTests.cs
```diff
@@ -252,8 +312,16 @@ private void CheckCapabilities(RPCClient rpc, string command, bool supported)
 		[Fact]
 		public void CanSyncWithPoW()
 		{
+
 			using (var builder = NodeBuilderEx.Create())
 			{
+				if (new[] {Liquid.Instance.Mainnet, Liquid.Instance.Testnet, Liquid.Instance.Regtest,}.Contains(
```
NicolasDorier: Please use a method for this copy/paste.

NBitcoin.Tests/AltcoinTests.cs
```diff
@@ -267,17 +335,29 @@ public void CanSyncWithPoW()
...
 			using (var builder = NodeBuilderEx.Create())
 			{
 				var node = builder.CreateNode();
 				builder.StartAll();
 				node.Generate(builder.Network.Consensus.CoinbaseMaturity);
 				var rpc = node.CreateRPCClient();
-				Assert.Equal(Money.Zero, rpc.GetBalance());
-				node.Generate(1);
-				Assert.NotEqual(Money.Zero, rpc.GetBalance());
+				if (new[] {Liquid.Instance.Mainnet, Liquid.Instance.Testnet, Liquid.Instance.Regtest,}.Contains(
```
NicolasDorier: The `GetBalanceAsync` should still pass and implies LBTC, even if it is liquid.

NBitcoin/Network.cs
```diff
@@ -76,7 +76,8 @@ public enum Base58Type
 	public enum Bech32Type
 	{
 		WITNESS_PUBKEY_ADDRESS,
-		WITNESS_SCRIPT_ADDRESS
+		WITNESS_SCRIPT_ADDRESS,
+		BLINDED_ADDRESS
```
NicolasDorier: no shitcoin code in NBitcoin

Kukks: I put this one here because there is already `Base58Type.BLINDED_ADDRESS`

NBitcoin.Tests/AltcoinTests.cs
Kukks: GetBalance rpc call send a different json structure than bitcoind

NBitcoin.Tests/AltcoinTests.cs
Kukks: haha yes

NBitcoin/Network.cs
NicolasDorier: fuck ok

NBitcoin.Altcoins/Liquid.cs
```diff
@@ -12,7 +14,12 @@ class LiquidRegtest { }
 		static Liquid()
 		{
 			ElementsParams<Liquid>.PeggedAssetId = new uint256("6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d");
+			ElementsParams<Liquid>.SignedBlocks = true;
+			ElementsParams<Liquid>.BlockHeightInHeader = true;
```
NicolasDorier: why are those? if they are always true, why even put them?

NBitcoin.TestFramework/WellknownNodeDownloadData.cs
```diff
@@ -732,20 +759,20 @@ public class ElementsNodeDownloadData
...
-					Executable = "elements-0.17.0/bin/elementsd.exe",
-					Hash = "e1dd04716d23e214b697ee33f2a77c803ae7e93ed93f2db68462a8c278361a24"
+					Executable = "elements-0.18.1.1/bin/elementsd.exe",
+					Hash = "f6ca18e3f4fe4fb4aadb70b7b447466afb68eec11d9cd5e2cac613414a28a1a5"
 				},
 				RegtestFolderName = "elementsregtest",
 				Chain = "elementsregtest",
-				AdditionalRegtestConfig = "initialfreecoins=210000000000000\nvalidatepegin=0",
-				UseSectionInConfigFile = true
+				AdditionalRegtestConfig = "initialfreecoins=0\nvalidatepegin=0\n\ncon_dyna_deploy_start=99999999999999999",
```
NicolasDorier: why changing this?

Kukks: initialfreecoins sets the default balance of bitcoin. For the tests, we want this 0

NBitcoin.Altcoins/Liquid.cs
Kukks: depends on your custom chain config for elements.

NBitcoin.TestFramework/WellknownNodeDownloadData.cs
NicolasDorier: Well, if we mine 101 blocks we should be able to spend something.
But I don't think liquid blocks have reward, so that's why I used freecoins.

Kukks: fair, let me switch and adjust

Kukks: done

Kukks:
The CI is failing weirdly, i think it has an old C# compiler version

build compiles & tests pass locally

NicolasDorier:
```
Elements/ElementsBlock.cs(114,6): error CS0311: The type 'NBitcoin.WitScript' cannot be used as type parameter 'T' in the generic type or method 'BitcoinStream.ReadWrite<T>(T)'. There is no implicit reference conversion from 'NBitcoin.WitScript' to 'NBitcoin.IBitcoinSerializable'. [/home/travis/build/MetacoSA/NBitcoin/NBitcoin.Altcoins/NBitcoin.Altcoins.csproj]
```

Kukks:
Weird, locally it compiles and executes. I wonder what is up with that

NBitcoin/Transaction.cs
```diff
@@ -1208,6 +1208,24 @@ public override string ToString()
 			return ToScript().ToString();
 		}
 
+		public void ReadWrite(BitcoinStream stream)
```
NicolasDorier: I want WitScript to be immutable

NBitcoin.Tests/AltcoinTests.cs
```diff
@@ -208,7 +246,18 @@ public void DoesRPCCapabilitiesWellAdvertised()
...
-					// If this fail, rpc support segwit bug you said it does not
+					if (IsElements(node.Network))
+					{
+//						if (address is BitcoinBlindedAddress blindedAddress)
+//						{
+//							address =blindedAddress.UnblindedAddress;
+//						}
+						//the above works too but might as well test this elements rpc call too
+						var addrInfo = await rpc.GetAddressInfoAsync(address.ToString(), rpc.Network);
+						address = addrInfo.Unconfidential;
```
NicolasDorier: 1. Why is this needed?  The `address.ScriptPubKey` of the confidential should be the same than the non confidential (check with a test)
2. It should be possible to get the unconfidentila address without a call to RPC, if so can you test that `addrInfo.Unconfidential` match what NBItcoin is returning?

NBitcoin.Tests/AltcoinTests.cs
```diff
@@ -267,17 +321,28 @@ public void CanSyncWithPoW()
...
+				{
+					Assert.Contains((await rpc.GetBalancesAsync()),
+						pair => pair.Value == Money.FromUnit(2100000, MoneyUnit.BTC));
+					node.Generate(1);
+					Assert.Contains((await rpc.GetBalancesAsync()),
+						pair => pair.Value == Money.FromUnit(2100000, MoneyUnit.BTC));
+				}
+				else
+				{
+					Assert.Equal(Money.Zero, await rpc.GetBalanceAsync());
```
NicolasDorier: Please hack `RPCClient.GetBalance` for elements so if/else is not needed (even if you need to do some small change to NBitcoin's rpcclient method)

Too many tests in NBXplorer depends on `GetBalance` and I don't want to pollute code everywhere with if/else specific to Element.

NBitcoin.Tests/AltcoinTests.cs
Kukks: You're right, removed this

NBitcoin.Tests/AltcoinTests.cs
Kukks: the initialfreecoins params on elements  still requires this test condition to be different as the starting balance is its value instead. 

Only this test uses GetBalance so I would rather not pollute further up code

NicolasDorier: @Kukks the problem is that tests in nbxplorer depends on getbalance doing what it is expected to do.

Kukks: Ok let me see what I can do, I will avoid tainting the virgin RPCClient code with blockstream lizard code

NBitcoin/BitcoinStream.cs
```diff
@@ -265,6 +265,25 @@ public void ReadWrite<T>(ref List<T> list) where T : IBitcoinSerializable, new()
 			}
 		}
 
+		public  void ReadWriteListBytes<TList>(ref TList data)
+			where TList : List<byte[]>, new()
```
NicolasDorier: Why making it generic? there is no reason to.

NBitcoin.Tests/NodeBuilderEx.cs
```diff
@@ -47,7 +47,7 @@ public static NodeBuilder Create([CallerMemberName] string caller = null)
 			//var builder = NodeBuilder.Create(NodeDownloadData.Dystem.v1_0_9_9, Altcoins.Dystem.Instance.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcoinplus.v2_7_0, Altcoins.AltNetworkSets.Bitcoinplus.Regtest, caller);
 
-			var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_18_0, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
+//			var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_18_1, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
```
NicolasDorier: uncomment this.

NBitcoin.Tests/NodeBuilderEx.cs
```diff
@@ -56,6 +56,8 @@ public static NodeBuilder Create([CallerMemberName] string caller = null)
 			//var builder = NodeBuilder.Create(NodeDownloadData.Stratis.v3_0_0, Altcoins.AltNetworkSets.Stratis.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.ZCoin.v0_13_8_3, Altcoins.ZCoin.Instance.Regtest, caller);
 
+
+			var builder = NodeBuilder.Create(NodeDownloadData.Elements.v0_18_1_1, Altcoins.AltNetworkSets.Liquid.Regtest, caller);
```
NicolasDorier: comment that

NBitcoin.Tests/AltcoinTests.cs
NicolasDorier: meh let it go if there is no other test in nbxplorer impacted.

NBitcoin.Altcoins/Liquid.cs
NicolasDorier: there is no custom chain. I prefer you remove it.

NBitcoin.Altcoins/Elements/ElementsTransaction.cs
```diff
@@ -603,9 +606,20 @@ protected override ConfidentialAsset GetAssetCore()
...
+		}
+
+		public ElementsTransaction(ElementsConsensusFactory<TNetwork> consensusFactory)
+		{
+			_consensusFactory = consensusFactory;
+		}
 		public override ConsensusFactory GetConsensusFactory()
 		{
-			return ElementsConsensusFactory<TNetwork>.Instance;
+			return _consensusFactory;
```
NicolasDorier: I worked hard so we don't have to do this.
Why can't we do what was done before?

Kukks: done

NBitcoin/BitcoinStream.cs
Kukks: done

PR #771, opened 2019-11-18
Clean Uint256

nopara73:
Done a bit of cleanup to get more familiar with the code. Feel free to close it if you don't find it useful.


NicolasDorier:
there is a build error.

nopara73:
Since I did not do anything remotely dangerous yet I still managed to somehow ruin the build, and considering that there's no functionality improvement here, maybe the best would be just to not touch it. (Also my suspicion of why I started investigating this class turned out to be false, so I kinda lost interest there:)

dangershony:
ha I think I can count on two hands already the attempts to refactor this class 😄

PR #774, opened 2019-11-19
Try using threadsafe list for filters again

zeptin:
This change was made in #698 and seemed to introduce stability problems.

However, repeated local runs of the test `CanMaintainChainWithSlimChainBehavior` have not elicited a similar result: they all pass.

### DIFF
--- a/NBitcoin/Protocol/Filters/NodeFiltersCollection.cs
+++ b/NBitcoin/Protocol/Filters/NodeFiltersCollection.cs
@@ -7,7 +7,7 @@ using System.Threading.Tasks;
 
 namespace NBitcoin.Protocol.Filters
 {
-	public class NodeFiltersCollection : ThreadSafeCollection<INodeFilter>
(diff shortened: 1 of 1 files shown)


NicolasDorier:
@zeptin well Travis do not agree https://travis-ci.org/MetacoSA/NBitcoin/jobs/613979111?utm_medium=notification&utm_source=github_status

mikedennis:
Interesting. I think the test that is failing is the linux netcoreapp3.0

```dotnet test --no-build -v n -c Release -f netcoreapp3.0 ./NBitcoin.Tests/NBitcoin.Tests.csproj```

Interestingly enough running the linux netstandard2.0 one passes. Seems like there might be subtle issue there with the framework.

NicolasDorier:
@mikedennis I don't think this is deterministic.

NicolasDorier:
I got this crash earlier even without this PR.

PR #776, opened 2019-11-22
Fix special case with empty filters

lontivero:
This fixes the filters matching for the special case where the filter is empty. Querying an empty filter must result in false result but it was returning true.


NicolasDorier:
can you remove the formating change?

nopara73:
IMO the formatting is good. A better view of the PR: https://github.com/MetacoSA/NBitcoin/pull/776/files?w=1

nopara73:
Also it's strange, I don't see any whitespace change.

lontivero:
oh fuck! I forced the push because git added the windows ending CRLF to each line. Sorry

lontivero:
Are you going to modify this PR? Or should I add your commit and push it on top of the forced commit?

PR #777, opened 2019-11-24
Update Bitcoin Core node version

lontivero:
This PR updates the Bitcoin Core node version in order to be able to start implementing the new `getfilter` RPC call for Wasabi usage. This is a WIP

### DIFF
--- a/NBitcoin.Tests/NodeBuilderEx.cs
+++ b/NBitcoin.Tests/NodeBuilderEx.cs
@@ -47,7 +47,8 @@ namespace NBitcoin.Tests
 			//var builder = NodeBuilder.Create(NodeDownloadData.Dystem.v1_0_9_9, Altcoins.Dystem.Instance.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcoinplus.v2_7_0, Altcoins.AltNetworkSets.Bitcoinplus.Regtest, caller);
 
-			var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_18_1, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
+			var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_19_0_1, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
+			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_18_1, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Liquid.v3_14_1_21, Altcoins.AltNetworkSets.Liquid.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcore.v0_15_2, Altcoins.Bitcore.Instance.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Gincoin.v1_1_0_0, Altcoins.Gincoin.Instance.Regtest, caller);
--- a/NBitcoin.Tests/NodeBuilderEx.cs
+++ b/NBitcoin.Tests/NodeBuilderEx.cs
@@ -48,7 +48,6 @@ namespace NBitcoin.Tests
 			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcoinplus.v2_7_0, Altcoins.AltNetworkSets.Bitcoinplus.Regtest, caller);
 
 			var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_19_0_1, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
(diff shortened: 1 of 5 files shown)


nopara73:
> in order to be able to start implementing the new getfilter RPC call for Wasabi usage. 

How do you imagine that?

lontivero:
Something like

```c#
	public CompactFilter GetFilter(uint256 blockHash)
	{
		var resp = SendCommand("getfilter", blockHash, "basic");
		return ParseCompactFilter(resp);
	}

	public async Task<CompactFilter> GetFilterAsync(uint256 blockHash)
	{
		var resp = await SendCommandAsync("getfilter", blockHash, "basic").ConfigureAwait(false);
		return ParseCompactFilter(resp);
	}

	public class CompactFilter
	{
		public GolombRiceFilter Filter { get; }
		public uint256 Header { get; }
	}
```

nopara73:
I mean how do you imagine using it in Wasabi? Our filters are 200MB, these filters are 4GB.  

Btw, I had an idea. I think it can come really handy for Bitcoin Core integration. Our last large problem is porting the backend filter building into the client, but with this we don't need to. Since we already only iterate through the last 100 filters, except at new wallet generation, we could make a great use for this. We don't need to build out filters on the client side, but it's ok to let Core build their filters, and we could create 2 IIndexProviders, where the new one would be using RPC directly.

nopara73:
Regarding this PR. Tests fail.

lontivero:
Yes, I know. A lot of RPC methods changed and have different parameter or return different results now. I am on that right now.

lontivero:
I will continue tomorrow. I don't know why `CanTestMempoolAccept` fails. Also, I have to see why the `AbortScanTxoutSetAsync` method fails now, it seems it requires some extra parameter maybe.

lontivero:
RPC is updated now and all the RPC tests pass but these other two tests fail:

* NBitcoin.Tests.ProtocolTests.CanExchangeFastPingPong
* NBitcoin.Tests.ProtocolTests.CanGetMerkleRoot

I don't know why. @NicolasDorier any advice would be great.

lontivero:
**It is ready for review**. The test failing has nothing to do with this work and passed all the previous commits, i mean, it is a random error.

knocte:
> so if the user uses older Core, then it'll fail

how about checking the version of Core via RPC and fail first if older version before trying to do anything

NicolasDorier:
hey sorry I am back.

> There are backwards incompatible changes, so if the user uses older Core, then it'll fail, but I guess there's nothing to do about that, Bitcoin Core just cannot keep not breaking the user space.

This.

The `RPCClient` should try to silently fix breaking changes from core as much as possible, I don't see your PR doing this.

NicolasDorier:
@knocte 

> how about checking the version of Core via RPC and fail first if older version before trying to do anything

For some stuff it is not really needed, you can just observe the shape of the json.
For other stuff, this is what the `ScanCapabilities` in the RPCClient call is about.

lontivero:
@NicolasDorier let me know if this version is better.

NBitcoin/RPC/RPCClient.cs
```diff
@@ -1389,9 +1402,21 @@ public MempoolAcceptResult TestMempoolAccept(Transaction transaction, bool allow
 			return TestMempoolAcceptAsync(transaction, allowHighFees).GetAwaiter().GetResult();
 		}
 
-		public async Task<MempoolAcceptResult> TestMempoolAcceptAsync(Transaction transaction, bool allowHighFees = false)
+        public async Task<MempoolAcceptResult> TestMempoolAcceptAsync(Transaction transaction, bool allowHighFees = false)
```
NicolasDorier: should be tab

NBitcoin.Tests/NodeBuilderEx.cs
```diff
@@ -47,7 +47,8 @@ public static NodeBuilder Create([CallerMemberName] string caller = null)
 			//var builder = NodeBuilder.Create(NodeDownloadData.Dystem.v1_0_9_9, Altcoins.Dystem.Instance.Regtest, caller);
 			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcoinplus.v2_7_0, Altcoins.AltNetworkSets.Bitcoinplus.Regtest, caller);
 
-			var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_18_1, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
+			var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_19_0_1, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
+			//var builder = NodeBuilder.Create(NodeDownloadData.Bitcoin.v0_18_1, Altcoins.AltNetworkSets.Bitcoin.Regtest, caller);
```
NicolasDorier: remove this line

NicolasDorier:
This is better, can you fix the two smalls nits I pointed out and rebase? I should have fixed the unit tests, sorry for that. Will make a release once merged this.

NicolasDorier:
Sorry, fixed the build again... can you rebase?

PR #780, opened 2019-11-30
Add convenience functions.

nopara73:
In case someone else wanting to learn would got stuck on this.

https://github.com/ProgrammingBlockchain/ProgrammingBlockchain/issues/134

### DIFF
--- a/NBitcoin/Transaction.cs
+++ b/NBitcoin/Transaction.cs
@@ -1622,15 +1622,15 @@ namespace NBitcoin
 		/// <summary>
 		/// Sign a specific coin with the given secret
 		/// </summary>
-		/// <param name="keys">Private key</param>
-		/// <param name="coins">Coins to sign</param>
-		public void Sign(BitcoinSecret key, params ICoin[] coins)
-			=> Sign(key, coins as IEnumerable<ICoin>);
+		/// <param name="key">Private key</param>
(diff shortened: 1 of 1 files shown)


NBitcoin/Transaction.cs
```diff
@@ -1619,6 +1619,22 @@ public int GetVirtualSize()
 			return (weight + WITNESS_SCALE_FACTOR - 1) / WITNESS_SCALE_FACTOR;
 		}
 
+		/// <summary>
+		/// Sign a specific coin with the given secret
+		/// </summary>
+		/// <param name="keys">Private key</param>
+		/// <param name="coins">Coins to sign</param>
+		public void Sign(BitcoinSecret key, params ICoin[] coins)
```
knocte: be careful, using params may even bring more confusion because then the parameter could be omitted completely; see https://github.com/MetacoSA/NBitcoin/pull/269 for more info

NicolasDorier: I think @knocte is right, should not be `params`.

nopara73: I'll remove it. The other overload addresses the reader's confusion, too.

nopara73: Actually removed and added another overload that may make sense.

knocte:
LGTM

PR #782, opened 2019-12-04
Adding DogeCash Support

Liquid369:
Testnet/Regtest are down at the moment and will be adjusted once they are there will be a subsequent PR
This is just starting for DogeCash integration so that we can continue to push into other services.
https://www.dogec.io/


Liquid369:
Will reopen another PR with README.md updates and NodeDownloadData

PR #786, opened 2019-12-09
Tests: Add failing WitScript serialization test

ysangkok:


NicolasDorier:
thanks for reporting this. This is very strange, looking into it.

NicolasDorier:
This was a huge bug. It was serializing the script as string instead as a hex string.

I just fixed it, but this basically mean this is a breaking change. Anybody who stored WitScript as json in database will have problem now... Hopefully this should be rare.

ysangkok:
@NicolasDorier  Thanks for fixing it. What versioning scheme does NBitcoin follow? Seems odd that this was only an increment in the subminor release version. I guess there is no promise of any ABI/API stability at all?

NicolasDorier:
@ysangkok I follow semver as much as I can. But when there is something that is almost unused which I break, I don't bump for this.

PR #787, opened 2019-12-13
Support provhdr on Stratis altcoin

zeptin:
Intended to supersede #692


NicolasDorier:
rebase?

NBitcoin/RPC/RestClient.cs
```diff
@@ -76,7 +76,7 @@ public RestClient(Uri address, Network network)
 		/// <param name="blockId">The block identifier.</param>
 		/// <returns>Given a block hash (id) returns the requested block object.</returns>
 		/// <exception cref="System.ArgumentNullException">blockId cannot be null.</exception>
-		public async Task<Block> GetBlockAsync(uint256 blockId)
+		public async Task<Block> GetBlockAsync(uint256 blockId, int verbosity = 0)
```
NicolasDorier: should be enum

zeptin: Both `RestClient` and `RPCClient` inherit from `IBlockRepository`, which defines a method `Task<Block> GetBlockAsync(uint256 blockId, int verbosity);`. So I'm not sure what the best approach would be, unless perhaps the enum was renamed and moved to be used by both implementations.

zeptin:
I've got this branch into a mess, reopening as a new PR

PR #789, opened 2019-12-19
Implement uptime RPC

nopara73:


NicolasDorier:
The test fail, probably because it is time sensitive. I think you can just simplify the test and make sure that calling the method does not crash.

nopara73:
Done. Indeed it could be zero if the query is so fast that it queries Bitcoin Core's uptime before a full second has been elapsed since start.

nopara73:
This time the appveyor failure is unrelated.

PR #798, opened 2019-12-31
Add Terracoin Support

TheSin-:
This is the initial request to add Terracoin.  I made this fork a while ago and tested it but got side tracked, I just merged in all the upstream changes but haven't tested it since.  I wanted to make the request before I forgot, I'll be testing this week but looking at the PR I can't see things breaking.


dangershony:
Nice to see TRC still going hard 👍

NicolasDorier:
@TheSin-  you tried running the tests for Altcoins with `var builder = NodeBuilder.Create(NodeDownloadData.Terracoin.v0_12_2, Altcoins.Terracoin.Instance.Regtest, caller);` ?

TheSin-:
I did but again it was like a. year ago I haven't seen the merge, it's on my list incase some calls changed, but it used to work, spent lots of time making sure of it.

TheSin-:
looks like tests require dotnetcore3.1 now, so might take me a bit to figure out how to get the on my Mac.  but hopefully before the day is up I'll be running tests again.

TheSin-:
I have one issue with tests, I need to be able to set a MIN Protocol on the clients, Terracoin bans obsolete protocols to force updates, so I can't have clients with 70012 connecting which is what the current tests do.  But I can only find a place to set a MAXP2PVersion not MIN.

Anyone have any hints on this issue?

NicolasDorier:
retry with this commit https://github.com/MetacoSA/NBitcoin/commit/c4fd43efdfaea62be346c645e328352e764c68c2

TheSin-:
That's that is much better, I don't have to stop it by hand now.

Now I just have 3 fails

NBitcoin.Tests.AltcoinTests.ElementsAddressSerializationTest which I do not think will fix as I think it's a macOS issue since it's even before Terracoin gets installed.

NBitcoin.Tests.AltcoinTests.CanSyncWithPoW and NBitcoin.Tests.AltcoinTests.CanSyncWithoutPoW still fail but with errors I can debug and work one, so thank you!!

Once those 2 work I'll post again

NicolasDorier:
`ElementsAddressSerializationTest ` is not important.
`CanSyncWithoutPoW` is the most important

TheSin-:
Just wanted to update, I'm stuck on

```
  X NBitcoin.Tests.AltcoinTests.CanSyncWithoutPoW [905ms]
  Error Message:
   System.FormatException : Superfluous witness record
```

Looking through the code I'm not sure how any AuxPow works at all, as there is no chain or version swaps happening, I'm trying to reimplement what I can but I"m having a have time figuring out how I can access the network consensus params inside of BlockHeader so I can get the AuxPow values I remember like ChainId, StrictChainId, etc etc

TheSin-:
Okay after lots of time I think the test might be the issue.

```
Assert.Equal(b.WithOptions(TransactionOptions.Witness).Header.GetType(), chain.GetBlock(50).Header.GetType());
```

This assumes Segwit which I have set to false, and there is no check.  If I check this line to

```
Assert.Equal(b.WithOptions(TransactionOptions.None).Header.GetType(), chain.GetBlock(50).Header.GetType());
```

Then it passes all tests.

NicolasDorier:
There is no reason the test does not pass even if you don't use segwit. It is only comparing types of headers. What is error?

TheSin-:
> There is no reason the test does not pass even if you don't use segwit. It is only comparing types of headers, the problem is probably deeper. What is error?

```
  X NBitcoin.Tests.AltcoinTests.CanSyncWithoutPoW [905ms]
  Error Message:
   System.FormatException : Superfluous witness record
```

If I make that change to from TransactionOptions.Witness to TransactionOptions.None in the test then it passes.

TheSin-:
Thank you.

TheSin-:
One more question, out client does version checking, I need to be able to modify _NUserAgent in VersionPayload.cs

I'd like to make it something like 

```
_NUserAgent = "/NBitcoin " + version.Major + "." + version.Minor + "." + version.Build + ":0.12.2.4/";
```

there can only be one colon and I need the TerracoinCore version after it.  So it'll look something like

```
/NBitcoin 5.0.14:0.12.2.4/
```

NicolasDorier:
There is no easy way to control that at the altcoin level, this is up to the user of the Node class to set that up. I advise you you remove this requirement, else you will go through lot's of pain to make NBitcoin works for you.

TheSin-:
hmmm, well this is a pickle, as I have no choice if I remove that there is a bad version that makes our network very unstable, so we need this control.  I'll have to think and figure something out it seems.  Thanks again for your help.

NicolasDorier:
You should probably just allow any version if the node is whitelisted.

PR #799, opened 2020-01-01
Make OutPoint Equatable

lontivero:
Simply implements IEquatablefor OutPoint https://github.com/zkSNACKs/WalletWasabi/pull/2909

### DIFF
--- a/NBitcoin/Transaction.cs
+++ b/NBitcoin/Transaction.cs
@@ -10,7 +10,7 @@ using System.Linq;
 namespace NBitcoin
 {
 #nullable enable
-	public class OutPoint : IBitcoinSerializable
+	public class OutPoint : IBitcoinSerializable, IEquatable<OutPoint>
 	{
 		public bool IsNull
 		{
@@ -180,12 +180,15 @@ namespace NBitcoin
 		{
 			return !(a == b);
 		}
+
 		public override bool Equals(object? obj)
 		{
-			var item = obj as OutPoint;
-			if (object.ReferenceEquals(null, item))
-				return false;
-			return item == this;
+			return Equals(obj as OutPoint);
+		}
+
+		public bool Equals(OutPoint? other)
+		{
+			return other == this;
 		}
 
 		public override int GetHashCode()


nopara73:
Can you implement all the related overloads? Like here: https://github.com/zkSNACKs/WalletWasabi/blob/f5b08391eecda420980174aeae8062c4b9584d6b/WalletWasabi/Blockchain/TransactionOutputs/TxoRef.cs

lontivero:
They are already there, I mean, `==`, `!=`, `<`, `>`, `Equals(object)` and `GetHashCode`. Am I missing something?

nopara73:
No, I didn't know.

NBitcoin/Transaction.cs
```diff
@@ -180,12 +180,15 @@ void SetNull()
...
 			return !(a == b);
 		}
+
 		public override bool Equals(object? obj)
 		{
-			var item = obj as OutPoint;
-			if (object.ReferenceEquals(null, item))
-				return false;
-			return item == this;
+			return Equals(obj as OutPoint);
```
nopara73:
```suggestion
			return obj is OutPoint op && Equals(op);
```

NBitcoin/Transaction.cs
```diff
@@ -180,12 +180,15 @@ void SetNull()
...
+		}
+
+		public bool Equals(OutPoint? other)
```
nopara73: Is the `?` is some C#8 nullable thingy?

lontivero: Yes, `Equals` must accept `null`.

NBitcoin/Transaction.cs
lontivero: It is unnecessary because in case it `obj` is not an `OutPoint` instance, after casting it is `null` and will be evaluated as false by the `==` operator.

nopara73: Indeed. I didn't know that, I thought it throws.

NBitcoin/Transaction.cs
nopara73: Does your reply imply the answer to my question is yes?

lontivero: Yes.

NicolasDorier:
Just to understand, why is it useful?

NBitcoin/Transaction.cs
nopara73: Btw this could be a good opportunity for @yahiheb to improve our IEquitable implementations based on this information.

nopara73:
IEquitable is used by .NET code under the hood, like `Contains` or `IndexOf`.

lontivero:
It is an endless discussion because there are no classes that will not eventually be stored in collections and queried by `Contains`, `IndexOf`, etc. It makes really sense with value types.

NicolasDorier:
@lontivero `Cointains` already works without equatable as Equals is on the Object type.

lontivero:
I know, that exactly my point.

PR #803, opened 2020-01-16
Support Stratis provhdr

zeptin:
Also tweaked some altcoin tests to allow them to pass more consistently on multiple networks.


NBitcoin/RPC/RPCClient.cs
```diff
@@ -1284,9 +1284,9 @@ public async Task<Block> GetBlockAsync(uint256 blockId)
 		/// </summary>
 		/// <param name="blockId"></param>
 		/// <returns></returns>
-		public Block GetBlock(uint256 blockId)
+		public Block GetBlock(uint256 blockId, int verbosity = 0)
```
knocte: should this be an int-based enum instead?

zeptin: https://github.com/MetacoSA/NBitcoin/pull/787#discussion_r366527035

NicolasDorier: 1. `IBlockRepository` will probably go away, this has not been very useful interface for this library
2. Adding optional parameter is a breaking change which requires rebuilding libraries depending on the version not using the parameter.

NBitcoin/Protocol/PayloadAttribute.cs
```diff
@@ -33,6 +33,15 @@ static PayloadAttribute()
 			}
 		}
 
+		public static void Add(string name, Type type)
+		{
+			if (!_NameToType.ContainsKey(name))
```
NicolasDorier: if you don't lock this, you'll end up with concurrency issue very soon.

NBitcoin/RPC/RestClient.cs
```diff
@@ -78,21 +77,24 @@ public RestClient(Uri address, Network network)
 		/// <exception cref="System.ArgumentNullException">blockId cannot be null.</exception>
 		public async Task<Block> GetBlockAsync(uint256 blockId)
 		{
-			if (blockId == null)
-				throw new ArgumentNullException(nameof(blockId));
-
-			var result = await SendRequestAsync("block", RestResponseFormat.Bin, blockId.ToString()).ConfigureAwait(false);
```
NicolasDorier: Keep this one as it was, there might be shitcoins not supporting the verbosity parameter, we don't want to break them.

NBitcoin.Tests/NodeBuilderEx.cs
```diff
@@ -1,59 +1,77 @@
...
 using System.Runtime.CompilerServices;
 
 namespace NBitcoin.Tests
 {
 	public class NodeBuilderEx
 	{
-		public static NodeBuilder Create([CallerMemberName] string caller = null)
+		public static NodeDownloadData GetNodeDownloadData()
 		{
-			//var builder = NodeBuilder.Create(NodeDownloadData.Litecoin.v0_17_1, Altcoins.Litecoin.Instance.Regtest, caller);
```
NicolasDorier: please don't edit that

zeptin: The original method still exists below. It merely calls the two new methods that split out the network set and node data into two reusable methods (e.g. note that some of the altcoin tests need to know which network they are running against, for some reason)

NicolasDorier:
It does not build on appveyor.

NBitcoin.Tests/NodeBuilderEx.cs
NicolasDorier: There is no point in doing that, now we need to modify 2 distinct places when 99% of the time we just want to uncomment 1 line.

zeptin: Can you suggest an improved way of determining which network set is in effect in the tests? It is necessary to inhibit certain tests from running when they are not running against the intended network (e.g. Liquid). If this is not done, at least one spurious test failure results across all other networks, which feels sloppy.

zeptin:
> It does not build on appveyor.

It seems like this would only happen if `NOSOCKET` is set on appveyor, which would render any of the tests involving the protocol inoperative. That would appear to defeat the point of having the tests?

NicolasDorier:
then surround the test with NOSOCKET

PR #811, opened 2020-02-19
Change access modifiers to allow inheritance

FrozDark:
Change access modifiers to allow inheritance for forked altcoins of the coin like BSV instead of creating whole code again


NicolasDorier:
Just copy/paste in BCash and make your modifications there. Those classes never change.
Note that BSV won't work unless you fork NBitcoin, there is a limit on the p2p message size.

PR #812, opened 2020-02-21
WIP: Liquid Assets Transaction Builder support

Kukks:
Move AssetMoney and AssetCoin from NBX to NBitcoin and support liquid assets in TransactionBuilder


Kukks:
Quite certain all this code is obsolete and needs a different approach.

PR #813, opened 2020-02-25
Speed up uint256

nopara73:
This is the same pattern I've applied twice when we've achieved significant speedup to Wasabi's loading time https://github.com/zkSNACKs/WalletWasabi/pull/3132

There are 2 intuition behind:

- recognize uint256 is immutable
- recognize uint256 equality comparisons must be fast when it's false, as most heavy usecases are searching through these lists

It's important that it's immutable, because then I can calculate and save the hashcode right when I create the object.

So I can give back the hashcode when I do GetHashCode, which speeds up set operations.

But also I can check for hashcode equality and return before I would check for equality, so I find the falses much faster.

### DIFF
--- a/NBitcoin/UInt2561.cs
+++ b/NBitcoin/UInt2561.cs
@@ -121,6 +121,8 @@ namespace NBitcoin
 		internal readonly UInt32 pn6;
 		internal readonly UInt32 pn7;
 
+		public int HashCode { get; }
+
 		public byte GetByte(int index)
 		{
 			var uintIndex = index / sizeof(uint);
@@ -175,6 +177,8 @@ namespace NBitcoin
 			pn5 = 0;
 			pn6 = 0;
 			pn7 = 0;
+
+			HashCode = CalculateHashCode();
 		}
 
 		public uint256(byte[] vch, bool lendian = true) : this(vch, 0, vch.Length, lendian)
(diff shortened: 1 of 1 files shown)


NBitcoin/UInt2561.cs
```diff
@@ -299,21 +307,34 @@ public int CompareTo(object obj)
...
+					equals &= a.pn5 == b.pn5;
+					equals &= a.pn6 == b.pn6;
+					equals &= a.pn7 == b.pn7;
+					return equals;
+				}
+				else
+				{
+					return false;
+				}			
+			}
```
yahiheb: Small nit: can be written this way:
```
else if (a.HashCode == b.HashCode)
{
	bool equals = true;
	equals &= a.pn0 == b.pn0;
	equals &= a.pn1 == b.pn1;
	equals &= a.pn2 == b.pn2;
	equals &= a.pn3 == b.pn3;
	equals &= a.pn4 == b.pn4;
	equals &= a.pn5 == b.pn5;
	equals &= a.pn6 == b.pn6;
	equals &= a.pn7 == b.pn7;
	return equals;
}
else
{
	return false;
}
```

NicolasDorier:
I don't like this, this make uint256 bigger, and calculate hashcode for each new instance of uint256 even if not used.

Maybe instead just make uint256 not time constant, this may fix the issue.

NicolasDorier:
Are you also sure the performance issues come from this?You tried with vs profiler? I find it hard to believe.

nopara73:
I did not close this wtf.

dangershony:
FYI this improvement for gethashcode was made by @aprogiena

https://github.com/block-core/blockcore/blob/master/src/NBitcoin/UInt256.cs#L507

PR #815, opened 2020-02-27
Speed up uint256 equality (again)

nopara73:
Same as https://github.com/MetacoSA/NBitcoin/pull/813. @NicolasDorier I'll address your points later.

### DIFF
--- a/NBitcoin/UInt2561.cs
+++ b/NBitcoin/UInt2561.cs
@@ -121,6 +121,8 @@ namespace NBitcoin
 		internal readonly UInt32 pn6;
 		internal readonly UInt32 pn7;
 
+		public int HashCode { get; }
+
 		public byte GetByte(int index)
 		{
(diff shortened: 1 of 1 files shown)


nopara73:
@NicolasDorier Replying to your question https://github.com/MetacoSA/NBitcoin/pull/813#issuecomment-591236435

I measured this, but unfortunately my largest wallet load is instant already, so I cannot get reliable data. The VS debugger has no large hotspots anymore either, but the hotspots are well distributed among many calls from the process tx.

This was however the largest bottleneck, but I eliminated it the same way that I solved it here for a class that relies on uint256 comparision and hashcode performance.

Reindexing the wallet still takes time, but this is not the bottleneck for that according to my measurements.

So to sum up, this does not help for our current situation, but will revisit it as it becomes a bottleneck again, as this class tends to be the bottleneck often.

PR #821, opened 2020-03-04
Moving out of BC to Secp256k1 ported in C#

NicolasDorier:
I plan to move out of BC to Secp256k1.

The main motivation is that it get easier for us to properly implement new feature of bitcoin like Schnorr signature by monkey coding what is built on top of Secp256k1.

NBitcoin will not expose externally the Secp256k1 port, but will instead try to rely on it as much as possible for things not natively supported by the standard library.
This is only available for environments with Span support (that is .NET Core 3.1 and 2.1).

New features won't have a port in BouncyCastle, we will keep using BC for environments without Span support.

* Less memory overhead. (The inner crypto of secp256k1 is made entirely with stack alloc and structs)
* Faster in some cases (sig verification is faster than BC)
* Side channel resistance. BC has been done without side channel resistance in mind. While we can't be sure that SECP256K1 ported in C# has side channel resistance, it is still an improvement over BC.
* Easy to port C/C++ code written by Core (We can just monkey port tests and code from C/C++ into C#)

As explained on twitter, I don't want to depends on bindings to Secp256k1.
The reason are:
1. When you start depending on native libraries, you need to ship all of them in your nuget package for all architectures you explicitly support.
2. You need to build for all those architecture in a reproductible build way so people can audit the native binaries.
3. Some environment may not even have binding support (Tiny CLR or WASM environments)

It is also inacceptable to put the burden of building secp256k1 on the users of NBitcoin. Not only this will break lot's of code, but make the use of this library considerably harder for shops without C++ programmers.

Crypto does not change overnight, but architecture support does.
As such, it is easier to maintain a port of secp256k1 in C# than to maintain repeatable builds.

I am porting all the tests from Secp256k1.

TODO:

* [x] Porting our schnorr stuff to secp256k1
* [x] Removing deps of BouncyCastle on ECDSASignature
* [ ] Porting the tests for ECDH support

---

Note on breaking change:

The is a single breaking change concerning blinding and maybe schnorr signing. @lontivero @nopara73 

The code before was requiring uint256 messages to be big endian, while ECDSA code does not, so to keep consistency, I inverted endianness.
You will need to be careful with that, because it means that if the coordinator is not same versio of NBitcoin with wasabi client, you'll get problems verifying schnorr sigs.

---

Note on native bindings to secp256k1:

I will consider native bindings when somebody have done the work of producing repeatable build on at least AMD64, ARM32, ARM64, windows and mac. Will then try to first use the native binaries and fallback to managed code if not available.


nopara73:
That was fast. Would you mind addressing this https://github.com/zkSNACKs/WalletWasabi/issues/3196#issuecomment-595904357 comment?

NicolasDorier:
that's 2 week I work full time on that :D

PR #824, opened 2020-03-04
Add gitattributes file

### LINKED ISSUE #823: LineEnding issues
Seriously, what is wrong with this project?

setting git config to `autocrlf true` - as it is recommended by github results in all lines considered to be changed.

so i set `autocrlf false` and in my editor lines are ending with CRLF, but `git diff` and also github too show the line endings as changed. with the addition of an `^M` (which is a `CR`) suggesting that your files indeed use `LF` as line endings.

The `.editorconfig` specifies a line ending of `CRLF`.

a `.gitattributes` could be added to solve this mystery.

I'd be glad for some clarification.
Thanks

yahiheb:
Fixes #823

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,63 @@
+###############################################################################
+# Set default behavior to automatically normalize line endings.
+###############################################################################
+* text=auto
+
+###############################################################################
+# Set default behavior for command prompt diff.
+#
+# This is need for earlier builds of msysgit that does not have it on by
+# default for csharp files.
+# Note: This is only used by command line
+###############################################################################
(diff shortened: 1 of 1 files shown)


knocte:
Why does Travis fail?

nopara73:
Not because of this.

knocte:
Would be great to get CI stable. Maybe converge to GithubActionsCI fixes it as a bonus?

knocte:
(GithubActionsCI also supports macOS, which IIRC is not supported by neither AppVeyor nor Travis in the free tier.)

PR #825, opened 2020-03-05
Update Terracoin to 0.12.2.5 which removed the version restriction

TheSin-:
Update Terracoin to 0.12.2.5 which removed the version restriction


TheSin-:
OMG I'm so sorry I forgot to update the hash :\

PR #826, opened 2020-03-06
Make RPCCLient.CreateWebRequest overrideable

Kukks:


NicolasDorier:
Should not use inheritance but composition.

Kukks:
But I just need to replace the http request message builder with my own?

PR #827, opened 2020-03-07
Make UnblindedSignature constructor public

nopara73:
image


NicolasDorier:
you can't, secp stuff are internal.
I don't want to expose internal crypto structure, why do you need it?

NicolasDorier:
Ah I guess you need it because you serialize that on the wire? In that case, give me a sec I will make some parsing methods.

NicolasDorier:
I pushed on 5.0.17

NicolasDorier:
Need to serialize on 64 bytes, 32 bytes for c followed by 32 bytes for s.

PR #832, opened 2020-03-15
Remove Schnorr binding stuff

lontivero:
This PR removes the types and tests about schnorr binding signatures.


knocte:
The commit message says the what, but not the why.

nopara73:
I guess the reasoning is that the latest change broke backwards compatibility, so we decided to move a compatible code to the Wasabi repo instead, and since nobody uses it other than us it's better to remove this non-essential part of the library? (I don't agree with this reasoning 100% though, but I don't mind removing it much either.)

NicolasDorier:
@knocte schnorr blinding is not useful to bitcoin in general, just specifically to Wasabi wallet, so I encouraged them to take on their own repo.

PR #833, opened 2020-03-16
added thousands separator to MoneyUnit for readability

soraphis:


NBitcoin/Money.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System.Globalization;
+using System.Globalization;
```
lontivero: This is not necessary.

nopara73: What is this btw? I see this glitch here and there.

knocte: it's the BOM being changed accidentally by text editors

soraphis: I don't get why this project i so broken in regards to line endings

Issue #773, opened 2019-11-18
ECDSASignature.ToDER() makes invalid DER from some r, s values


DanGould:
For example, given the raw 64 byte `r|s` signature `E1AC191AA300507C6EB94A0F77717E6523FA523CE06F19904C970F4F1B1167AA6FC193416BF35B38EFA47EBDA9BB93681206BC7B6FFF49BC045B2D7491EFD394`
such that bytes 0-31 are `r` and 32-63 are `s`

```csharp
var realSig = "E1AC191AA300507C6EB94A0F77717E6523FA523CE06F19904C970F4F1B1167AA6FC193416BF35B38EFA47EBDA9BB93681206BC7B6FFF49BC045B2D7491EFD394";
var r = new NBitcoin.BouncyCastle.Math.BigInteger(StringToByteArray(realSig.Substring(0, 64)));
var s = new NBitcoin.BouncyCastle.Math.BigInteger(StringToByteArray(realSig.Substring(64)));
var ecdsa = new ECDSASignature(r, s).MakeCanonical();
Assert.True(ECDSASignature.IsValidDER(ecdsa)); // false
var der = new ECDSASignature(ecdsa.ToDER());
Assert.True(ECDSASIgnature.IsValidDER(der)); // false

// ...
// The above relies on the following helper:
public static byte[] StringToByteArray(string hex)
{
	return Enumerable.Range(0, hex.Length)
		 .Where(x => x % 2 == 0)
		 .Select(x => Convert.ToByte(hex.Substring(x, 2), 16))
		 .ToArray();
}
```

Because both `r` and `s`'s top bytes `0xE1` and `0x6F` begin with 1, for them to be valid DER a `0x00` byte must be appended to the front. They pass the `IsValidDER` tests if imported as follows:

```csharp
var r = new NBitcoin.BouncyCastle.Math.BigInteger(StringToByteArray("00" + realSig.Substring(0, 64)));
var s = new NBitcoin.BouncyCastle.Math.BigInteger(StringToByteArray("00" + realSig.Substring(64)));
```

but isn't the whole point of `ToDER()` or `MakeCanonical()` to right these wrongs and make a valid DER? Happy to PR to add this check if my understanding of these methods is in line with NBitcoin's intention for them.

zeptin:
Unless I am missing something, that example does not compile.

`MakeCanonical` only affects the S value, per the method's comment.

DanGould:
You're missing
```csharp
[...]

PR #836, opened 2020-03-25
Test ECDSASignature.ToDER is Bip66 compatible

DanGould:
demonstrates #773 is resolved. ECDSASignature.ToDER() encodes only positive exponents.

-- 

To get the tests to appear and build on VS Mac, I had to switch around the order of TargetFramework Dependencies:
```diff
-         <TargetFrameworks>net461;netcoreapp3.1</TargetFrameworks>
+         <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
```

If this doesn't cause any other issue, I will create a separate pull request to resolve this


DanGould:
I think this is failing because the test environment is using BouncyCastle ECDSASignature still, which should fail

NicolasDorier:
@DanGould can you just remove this test for netframework? That's fine, we don't need to fix it.
Then I can merge this.

NicolasDorier:
Also don't use separate file, put it somewhere else where there is tests on sigs

NicolasDorier:
Also you can switch both to fix on mac in this PR.

NicolasDorier:
You added the tests in a `!HAS_SPAN` section, should be the reverse.

DanGould:
thanks for your patience

PR #837, opened 2020-03-27
Calculate dust threshold for native segwit

lontivero:
This PR takes into account the fact that the scriptSig for segwit is smaller than for other like p2pkh.

Related https://github.com/zkSNACKs/WalletWasabi/issues/3217

### DIFF
--- a/NBitcoin/Transaction.cs
+++ b/NBitcoin/Transaction.cs
@@ -647,8 +647,16 @@ namespace NBitcoin
 		{
 			if (minRelayTxFee == null)
 				throw new ArgumentNullException("minRelayTxFee");
-			int nSize = this.GetSerializedSize() + 148;
-			return 3 * minRelayTxFee.GetFee(nSize);
+
+			// OutPoint (32 + 4) + script_size (1) + sequence (4)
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Have you actually tried that a fee with exactly the treshold is accepted by bitcoin core just in case?

lontivero:
No. I will do it and let you know here.

lontivero:
I verified that I can broadcast a tx with a segwit script and value of 294 satoshis: https://testnet.smartbit.com.au/tx/e65e3afc1369d10bb4aa877c76f540d883f4d54b23f5d41cb5c208bef0074eb0
The network does not accept a segwit output of 293 (what is the expected behavior).

nopara73:
Great work @lontivero! It wasn't a nasty bug after all, but better make sure about these edge cases when they are about sending money as that's the number one most important functionality of a bitcoin software, coinjoin and everything else is just sugar.

NicolasDorier:
I pushed new version btw.

nopara73:
Thanks, but no worry, we cannot update as we hold off with it till the release because of the crypto stuff.

PR #843, opened 2020-04-14
Fix PSBTInput.UpdateFrom method

lontivero:
Honestly I am not 100% sure about this change but it looks wrong to me at first glance. I can be wrong.


NicolasDorier:
Makign new version now.

PR #852, opened 2020-05-18
Improve performance of uint256

NicolasDorier:
Before
|       Method |     Mean |    Error |   StdDev |
|------------- |---------:|---------:|---------:|
|         Read | 27.78 ns | 0.422 ns | 0.352 ns |
| WriteToArray | 89.02 ns | 1.452 ns | 1.358 ns |
|  WriteToSpan | 20.86 ns | 0.404 ns | 0.338 ns |

After
|       Method |      Mean |     Error |    StdDev |
|------------- |----------:|----------:|----------:|
|         Read |  9.020 ns | 0.1607 ns | 0.1424 ns |
| WriteToArray | 14.387 ns | 0.1589 ns | 0.1487 ns |
|  WriteToSpan | 14.172 ns | 0.1521 ns | 0.1348 ns |

Note that I only do on endian machine, as the logic is too tricky and untestable for bigendian machine. @lontivero @nopara73 you wanted that for a while I think.

@dangershony  I think somebody was also bumping his head on this at one point in stratis?

### DIFF
--- a/NBitcoin.Bench/UInt256Bench.cs
+++ b/NBitcoin.Bench/UInt256Bench.cs
@@ -1,0 +1,36 @@
+﻿using BenchmarkDotNet.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NBitcoin.Bench
+{
+	public class UInt256Bench
+	{
+		uint256 Value;
+		byte[] Bytes;
+		byte[] EmptyArray;
+		[GlobalSetup]
+		public void Setup()
+		{
+			Value = RandomUtils.GetUInt256();
+			Bytes = Value.ToBytes();
+			EmptyArray = new byte[32];
+		}
+		[Benchmark]
+		public void Read()
+		{
+			new uint256(Bytes);
+		}
+		[Benchmark]
+		public void WriteToArray()
+		{
+			Value.ToBytes(EmptyArray);
+		}
+		[Benchmark]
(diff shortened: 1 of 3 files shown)


lontivero:
Wow. This is really good. Don't forget to share the benchmark project too. It is hard to believe this can be improve further but we also make a lot of comparisons too.

NicolasDorier:
Sorry, forgot to add the bench file, just did. Test failing on travis, but this is unrelated to this.

NicolasDorier:
@lontivero to be honest, I am surprised it is so slow. As a comparison, parsing a bech32 takes 5 ns, and base58 10ns... I can't believe those are not doing more stuff...

NicolasDorier:
I also think we can make things faster by moving from uint to ulong. (probably twice faster)
But I let it to the motivated person :D

dangershony:
Hi @NicolasDorier  this is great.
At the moment for blockcore I used uint256 improvements made by @mithrilman in his [mithrilshards](https://github.com/MithrilMan/MithrilShards/tree/master/MithrilShards.P2P.Benchmark/Benchmarks/UInt256) project.

See [benchmarks](https://github.com/block-core/blockcore/issues/87) here.

The idea is using 4 ulongs instead of 8 uints

NicolasDorier:
I think you can win a bit more by using Span cast like I did.

NicolasDorier:
moving to ulong

PR #854, opened 2020-05-20
Use ulong inside uint256 to improve perf

NicolasDorier:
Continuation from https://github.com/MetacoSA/NBitcoin/pull/852

Use ulong instead of uint.

Before

|       Method |      Mean |     Error |    StdDev |
|------------- |----------:|----------:|----------:|
|         Read |  9.020 ns | 0.1607 ns | 0.1424 ns |
| WriteToArray | 14.387 ns | 0.1589 ns | 0.1487 ns |
|  WriteToSpan | 14.172 ns | 0.1521 ns | 0.1348 ns |

After 
|       Method |      Mean |     Error |    StdDev |
|------------- |----------:|----------:|----------:|
|         Read |  7.195 ns | 0.1243 ns | 0.1102 ns |
| WriteToArray | 13.537 ns | 0.2288 ns | 0.2140 ns |
|  WriteToSpan | 13.271 ns | 0.1050 ns | 0.0820 ns |

### DIFF
--- a/NBitcoin/Crypto/Hashes.cs
+++ b/NBitcoin/Crypto/Hashes.cs
@@ -615,13 +615,13 @@ namespace NBitcoin.Crypto
 				switch (position)
 				{
 					case 0:
-						return (ulong)val.pn0 + (ulong)((ulong)val.pn1 << 32);
+						return (ulong)val.pn0;
 					case 1:
-						return (ulong)val.pn2 + (ulong)((ulong)val.pn3 << 32);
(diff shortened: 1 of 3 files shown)


nopara73:
Pinging @dangershony as you may be interested.

NicolasDorier:
I don't reallly understand why, but this is still slower than bech32 parsing...

dangershony:
Ah this is really great, I think @MithrilMan is probably also interested too.

PR #856, opened 2020-05-20
Refactor TransactionBuilder

NicolasDorier:
Refactoring transaction builder. In some corner case situation, the transaction builder can pay less than the relay fee. Fixing this bug is not easy, so I need to refactor step by step this spaghetti.

### DIFF
--- a/NBitcoin/Utils.cs
+++ b/NBitcoin/Utils.cs
@@ -739,20 +739,24 @@ namespace NBitcoin
 				arr[fromIndex] = to;
 			}
 		}
-		public static void Shuffle<T>(List<T> arr, Random rand)
+		public static void Shuffle<T>(List<T> arr, int start, Random rand)
 		{
 			rand = rand ?? new Random();
-			for (int i = 0; i < arr.Count; i++)
+			for (int i = start; i < arr.Count; i++)
 			{
-				var fromIndex = rand.Next(arr.Count);
+				var fromIndex = rand.Next(start, arr.Count - start);
 				var from = arr[fromIndex];
 
-				var toIndex = rand.Next(arr.Count);
+				var toIndex = rand.Next(start, arr.Count - start);
 				var to = arr[toIndex];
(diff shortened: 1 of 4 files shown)


NBitcoin/Utils.cs
```diff
@@ -739,21 +739,25 @@ public static void Shuffle<T>(T[] arr, Random rand)
...
 			rand = rand ?? new Random();
-			for (int i = 0; i < arr.Count; i++)
+			for (int i = start; i < arr.Count; i++)
 			{
-				var fromIndex = rand.Next(arr.Count);
+				var fromIndex = rand.Next(start, arr.Count - start);
 				var from = arr[fromIndex];
 
-				var toIndex = rand.Next(arr.Count);
+				var toIndex = rand.Next(start, arr.Count - start);
```
lontivero, line 750: This works okay but you could use some syntax sugar, something from:
 
```c#
var segment = arr[start..]; 

var fromIdx = rand.TakeOne(segment); 
var toIdx = rand.TakeOne(segment); 
(segment[fromIdx], segment[toIdx]) = (segment[toIdx], segment[fromIdx]);
```

NicolasDorier: I must admit that I have hard time to understand what it does :p

dangershony: ha I agree I am not sure I would call this sugar `arr[start..]`

NicolasDorier:
The old TransactionBuilder had a slight change to generate a transaction with fee lower than the feerate if the coinselector was chosing a different set of coins than it did when `SendEstimatedFee` was called.

This was a very rare bug, actually found it via spurious tests running the builder thousands of time as the CI was sometimes failing the test.

The refactoring took me 3 fucking days. This is the trickiest code of NBitcoin. But now it is not only more readable, with the bug fixed, but also it will generate transactions with way closer to the expected fee rate!

PR #859, opened 2020-05-25
Add Qtum to Altcoins

dabura667:


dabura667:
@qtum-neil Maybe take a look too?

PR #860, opened 2020-05-25
Add ECDSASignature.ToCompact

joemphilips:
For compact byte format serialization.


NBitcoin/Crypto/ECDSASignature.cs
```diff
@@ -243,6 +243,8 @@ public ECDSASignature(BigInteger[] rs)
 			_R = rs[0];
 			_S = rs[1];
 		}
+
+		[Obsolete("This member is not available on .netstandard2.1 and will soon be removed")]
```
NicolasDorier, line 247: this member IS available in .netstandard2.1

joemphilips:
I think test failure has nothing to do with my change.

PR #869, opened 2020-06-10
Fix couple of typos in ArgumentException messages

knocte:


NicolasDorier:
If they do that, it means either that they are bad dev, or that NBitcoin need better exception type and they failed to tell us. Both is their fault. :p

PR #870, opened 2020-06-12
Add TryCombine function for ECPubKey

joemphilips:
* This is a port of `secp256k1_ec_pubkey_combine` from
  original secp256k1 implementation.
* It also have one unit test. In the original implementation,
  there is also a test named `run_eckey_edge_case_test`.
  Ideally this test should be ported too. But since the test is
  not only for combining pubkey, I've omitted from this commit.

This is necessary for deriving remote pubkey in LN.

### DIFF
--- a/NBitcoin/Secp256k1/ECPubKey.cs
+++ b/NBitcoin/Secp256k1/ECPubKey.cs
@@ -312,10 +312,11 @@ namespace NBitcoin.Secp256k1
 		{
 			if (pubkeys == null)
 				throw new ArgumentNullException(nameof(pubkeys));
-			combinedPubKey = null;
+
 			var count = pubkeys.Count();
 			if (count == 0)
 			{
+				combinedPubKey = null;
 				return false;
 			}
 			var ptj = GEJ.Infinity;
@@ -323,8 +324,7 @@ namespace NBitcoin.Secp256k1
 			{
 				ptj = ptj.Add(in p.Q);
 			}
-			var pt = ptj.ToGroupElement();
-			combinedPubKey = new ECPubKey(new GE(pt.x.Normalize(), pt.y.Normalize()), ctx);
+			combinedPubKey = new ECPubKey(ptj.ToGroupElement(), ctx);
 			return true;
 		}
 
--- a/NBitcoin/Secp256k1/ECPubKey.cs
+++ b/NBitcoin/Secp256k1/ECPubKey.cs
@@ -308,17 +308,12 @@ namespace NBitcoin.Secp256k1
 		/// <summary>
 		/// The original function name is `secp256k1_ec_pubkey_combine`
 		/// </summary>
-		public static bool TryCombine(Context ctx, ECPubKey[] pubkeys, out ECPubKey? combinedPubKey)
+		public static bool TryCombine(Context ctx, IEnumerable<ECPubKey> pubkeys, out ECPubKey? combinedPubKey)
 		{
 			if (pubkeys == null)
 				throw new ArgumentNullException(nameof(pubkeys));
 
 			combinedPubKey = null;
-			var n = pubkeys.Count();
-			if (n == 0)
-			{
-				return false;
-			}
 			var qj = GEJ.Infinity;
 			foreach (var p in pubkeys)
 			{
--- a/NBitcoin/Secp256k1/ECPubKey.cs
+++ b/NBitcoin/Secp256k1/ECPubKey.cs
@@ -1,6 +1,7 @@
 ﻿#if HAS_SPAN
 #nullable enable
 using System;
+using System.Linq;
 using System.Collections.Generic;
 using System.Text;
 
@@ -301,6 +302,29 @@ namespace NBitcoin.Secp256k1
 				return false;
 			}
 			key = pt.ToGroupElement();
+			return true;
+		}
+
+		/// <summary>
+		/// The original function name is `secp256k1_ec_pubkey_combine`
+		/// </summary>
+		public static bool TryCombine(Context ctx, IEnumerable<ECPubKey> pubkeys, out ECPubKey? combinedPubKey)
+		{
+			if (pubkeys == null)
+				throw new ArgumentNullException(nameof(pubkeys));
+			combinedPubKey = null;
+			var count = pubkeys.Count();
+			if (count == 0)
+			{
+				return false;
+			}
+			var ptj = GEJ.Infinity;
+			foreach (var p in pubkeys)
+			{
+				ptj = ptj.Add(in p.Q);
+			}
+			var pt = ptj.ToGroupElement();
+			combinedPubKey = new ECPubKey(new GE(pt.x.Normalize(), pt.y.Normalize()), ctx);
 			return true;
 		}
 
--- a/NBitcoin/Secp256k1/ECPubKey.cs
+++ b/NBitcoin/Secp256k1/ECPubKey.cs
@@ -308,23 +308,26 @@ namespace NBitcoin.Secp256k1
 		/// <summary>
 		/// The original function name is `secp256k1_ec_pubkey_combine`
 		/// </summary>
-		public static bool TryCombine(Context ctx, IEnumerable<ECPubKey> pubkeys, out ECPubKey? combinedPubKey)
+		public static bool TryCombine(Context ctx, ECPubKey[] pubkeys, out ECPubKey? combinedPubKey)
 		{
 			if (pubkeys == null)
 				throw new ArgumentNullException(nameof(pubkeys));
 
-			var count = pubkeys.Count();
-			if (count == 0)
+			combinedPubKey = null;
+			var n = pubkeys.Count();
+			if (n == 0)
 			{
-				combinedPubKey = null;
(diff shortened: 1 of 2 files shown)


NBitcoin/Secp256k1/ECPubKey.cs
```diff
@@ -304,6 +305,29 @@ private bool secp256k1_eckey_pubkey_tweak_add(ECMultContext ctx, ref GE key, in
...
+			{
+				return false;
+			}
+			var ptj = GEJ.Infinity;
+			foreach (var p in pubkeys)
+			{
+				ptj = ptj.Add(in p.Q);
+			}
+			var pt = ptj.ToGroupElement();
+			combinedPubKey = new ECPubKey(new GE(pt.x.Normalize(), pt.y.Normalize()), ctx);
```
lontivero, line 327: What's the difference with this code?
```c#
combinedPubKey = new ECPubKey(pt.ToGroupElement(), ctx);
```

NBitcoin/Secp256k1/ECPubKey.cs
```diff
@@ -304,6 +305,29 @@ private bool secp256k1_eckey_pubkey_tweak_add(ECMultContext ctx, ref GE key, in
...
 		}
 
+		/// <summary>
+		/// The original function name is `secp256k1_ec_pubkey_combine`
+		/// </summary>
+		public static bool TryCombine(Context ctx, IEnumerable<ECPubKey> pubkeys, out ECPubKey? combinedPubKey)
+		{
+			if (pubkeys == null)
+				throw new ArgumentNullException(nameof(pubkeys));
+			combinedPubKey = null;
```
lontivero, line 315: I would move this immediately before the `return false` statement.

NBitcoin/Secp256k1/ECPubKey.cs
joemphilips: I wasn't aware that ECPubKey constructor was running `NormalizeVariable`, so your concern is right, running Normalize() here is unnecessary. I will change it to what you have suggested.

NicolasDorier: @joemphilips folllow what secp256k1 is doing. If they do it, do it as well.

NBitcoin/Secp256k1/ECPubKey.cs
```diff
@@ -304,6 +305,29 @@ private bool secp256k1_eckey_pubkey_tweak_add(ECMultContext ctx, ref GE key, in
...
 
+		/// <summary>
+		/// The original function name is `secp256k1_ec_pubkey_combine`
+		/// </summary>
+		public static bool TryCombine(Context ctx, IEnumerable<ECPubKey> pubkeys, out ECPubKey? combinedPubKey)
+		{
+			if (pubkeys == null)
+				throw new ArgumentNullException(nameof(pubkeys));
+
+			var count = pubkeys.Count();
```
NicolasDorier, line 316: It is an `IEnumerable`, I would prefer you do not cause it to be evaluated twice.

NBitcoin/Secp256k1/ECPubKey.cs
```diff
@@ -304,6 +305,29 @@ private bool secp256k1_eckey_pubkey_tweak_add(ECMultContext ctx, ref GE key, in
 			return true;
 		}
 
+		/// <summary>
+		/// The original function name is `secp256k1_ec_pubkey_combine`
+		/// </summary>
+		public static bool TryCombine(Context ctx, IEnumerable<ECPubKey> pubkeys, out ECPubKey? combinedPubKey)
```
NicolasDorier, line 311: This implementation is not following line by line 

```c
int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *pubnonce, const secp256k1_pubkey * const *pubnonces, size_t n) {
    size_t i;
    secp256k1_gej Qj;
    secp256k1_ge Q;

    ARG_CHECK(pubnonce != NULL);
    memset(pubnonce, 0, sizeof(*pubnonce));
    ARG_CHECK(n >= 1);
    ARG_CHECK(pubnonces != NULL);

    secp256k1_gej_set_infinity(&Qj);

    for (i = 0; i < n; i++) {
        secp256k1_pubkey_load(ctx, &Q, pubnonces[i]);
        secp256k1_gej_add_ge(&Qj, &Qj, &Q);
    }
    if (secp256k1_gej_is_infinity(&Qj)) {
        return 0;
    }
    secp256k1_ge_set_gej(&Q, &Qj);
    secp256k1_pubkey_save(pubnonce, &Q);
    return 1;
}
```

Please use array instead of IEnumerable. (so we can get `n` from it, without having to evaluate the enumerable)

NBitcoin/Secp256k1/ECPubKey.cs
```diff
@@ -304,6 +305,29 @@ private bool secp256k1_eckey_pubkey_tweak_add(ECMultContext ctx, ref GE key, in
...
+			{
+				combinedPubKey = null;
+				return false;
+			}
+			var ptj = GEJ.Infinity;
+			foreach (var p in pubkeys)
+			{
+				ptj = ptj.Add(in p.Q);
+			}
+			combinedPubKey = new ECPubKey(ptj.ToGroupElement(), ctx);
```
NicolasDorier, line 327: Should be jacobian.
` secp256k1_ge_set_gej(&Q, &Qj);`

joemphilips: I don't understand, I thought what these two lines do is
```c
secp256k1_ge_set_gej(&Q, &Qj);
secp256k1_pubkey_save(pubnonce, &Q);
```
1. Convert jacobian coordinate (Qj) into non jacobian-coordinate representation Q and
2. initialize pubkey (pubnonce) with that Q.
 
so it seems current code is fine.

NBitcoin/Secp256k1/ECPubKey.cs
joemphilips: Ok 
> Use variable name as well.

You mean same variable name as much as possible right?

NBitcoin/Secp256k1/ECPubKey.cs
joemphilips: it is passing generated ge to `secp256k1_pubkey_save` , I suppose that is equivalent to ECPubKey constructor. So `new ECPubKey(qj.ToGroupElement())` is equivalent.

joemphilips:
Updated.

NBitcoin/Secp256k1/ECPubKey.cs
```diff
@@ -304,6 +305,32 @@ private bool secp256k1_eckey_pubkey_tweak_add(ECMultContext ctx, ref GE key, in
...
+		/// <summary>
+		/// The original function name is `secp256k1_ec_pubkey_combine`
+		/// </summary>
+		public static bool TryCombine(Context ctx, ECPubKey[] pubkeys, out ECPubKey? combinedPubKey)
+		{
+			if (pubkeys == null)
+				throw new ArgumentNullException(nameof(pubkeys));
+
+			combinedPubKey = null;
+			var n = pubkeys.Count();
```
NicolasDorier, line 317: You must use `.Length` because `.Count()` is actually walking over the whole collection to count stuff.

NBitcoin/Secp256k1/ECPubKey.cs
```diff
@@ -304,6 +305,32 @@ private bool secp256k1_eckey_pubkey_tweak_add(ECMultContext ctx, ref GE key, in
...
+			if (n == 0)
+			{
+				return false;
```
NicolasDorier, line 320: I would actually prefer if you don't count at all, like the C code. Less code paths (and you could even keep IEnumerable)

NBitcoin/Secp256k1/ECPubKey.cs
joemphilips: Ok. I actually thought `Count()` will become alias for `Length` in case of  arrays.

NBitcoin/Secp256k1/ECPubKey.cs
joemphilips: I think that's a weird api (how do we assure that actual length is longer than given n?).
What about using `IList` if you think array is too strict.

NBitcoin/Secp256k1/ECPubKey.cs
knocte: actually Joe is right here: https://stackoverflow.com/a/981292/544947

NBitcoin/Secp256k1/ECPubKey.cs
NicolasDorier: why do you want to ensure the actual length is longer than n in the first place? If it is, then it returns false even if you remove this call.

joemphilips: Ah, I see. I first thought you are saying we should pass n as a parameter to the function.

joemphilips:
Updated

NicolasDorier:
pushed 1.0.3

PR #881, opened 2020-07-14
Strict Bech32 HRP comparison

### LINKED ISSUE #880: Invalid bech32 addresses recognized as valid
Recently the bech32 encoder changed (fixed a incorrect restrictive) the implementation but introduced a new bug.  

```c#
[Fact]
public void CanDetectError()
{
	var bech = Encoders.Bech32("tb");
	Assert.Throws<Bech32FormatException>(() => bech2.Decode("tb1qdtpzu24mlyz3k0vrqxyntq0xqdqva7zxm7r485tb1qdtpzu24mlyz3k0vrqxyntq0xqdqva7zxm7r485", out wit));
}
```

**Bip173 states that:**

> As a result of the previous rules, addresses are always between **14 and 74** characters long, and their length modulo 8 cannot be 0, 3, or 5. Version 0 witness addresses are always 42 or 62 characters, but implementations MUST allow the use of any version. 

Here there is an implementation by sipa that fails with the with the address used in the unit test http://bitcoin.sipa.be/bech32/demo/demo.html 

Related to https://github.com/zkSNACKs/WalletWasabi/issues/3933

lontivero:
Fixes #880

### DIFF
--- a/NBitcoin.Tests/Bech32Test.cs
+++ b/NBitcoin.Tests/Bech32Test.cs
@@ -71,6 +71,9 @@ namespace NBitcoin.Tests
 			Assert.Equal(2, ex.ErrorIndexes.Length);
 			Assert.Equal(8, ex.ErrorIndexes[0]);
 			Assert.Equal(12, ex.ErrorIndexes[1]);
+
+			var bech2 = Encoders.Bech32("tb");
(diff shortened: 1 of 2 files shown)


nopara73:
@NicolasDorier would you mind give out a release with this?

NicolasDorier:
i did

nopara73:
Thank you!

PR #883, opened 2020-07-22
Fix bug with GetMempoolEntry

nopara73:
It threw even when don't throw was specified.


NicolasDorier:
you can fix the conflict?

NicolasDorier:
pushed new @lontivero

nopara73:
Thank you!

PR #884, opened 2020-07-22
Add double spend test

nopara73:
With Bitcoin Knots 19 this test fails. Since Bitcoin Knots consists of PRs from Bitcoin Core, for this reason it may be best if we have a test to notice if the PR would be accidentally rolled out. I'm not sure why Knots doesn't handle double spend properly yet, I'm still investigating it, nevertheless I think this test is good to have.

**Update** This works properly with Core 20 and Knots 20.

### DIFF
--- a/NBitcoin.Tests/RPCClientTests.cs
+++ b/NBitcoin.Tests/RPCClientTests.cs
@@ -1143,6 +1143,40 @@ namespace NBitcoin.Tests
 		}
 
 		[Fact]
+		public void DoubleSpendThrows()
+		{
+			using (var builder = NodeBuilderEx.Create())
+			{
+				var node = builder.CreateNode();
+				var rpc = node.CreateRPCClient();
+				builder.StartAll();
+				var network = node.Network;
+
+				var key = new Key();
+				var blockId = rpc.GenerateToAddress(1, key.PubKey.WitHash.GetAddress(network));
+				var block = rpc.GetBlock(blockId[0]);
+				var coinBaseTx = block.Transactions[0];
+
+				var tx = Transaction.Create(network);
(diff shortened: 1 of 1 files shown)


luke-jr:
This *shouldn't* pass, as the second transaction is a RBF. My guess is you're somehow triggering a different rejection. I was unable to reproduce a reject of the 2nd transaction with Knots 0.20 (but I rewrote the test in Python for the QA test framework).

Can you modify the test to check what the rejection reason is? (In addition to figuring out what you want to test here..)

nopara73:
Neither the first, nor the double spend is RBF:

image

As a summary, my tests have shown that:

|      | Bitcoin Core       | Bitcoin Knots         |
|------|--------------------|-----------------------|
| 0.19 | double spend fails | double spend succeeds |
| 0.20 | double spend fails | double spend fails    |

luke-jr:
The first transaction has an output of 49.9999 BTC, therefore a fee of 0.0001 BTC. The second has an output of 49.998, therefore a fee of 0.002 BTC. Thus, the second has preferable fee and accepted to replace the first.

Core requires BIP 125 signalling to allow replacements. Knots does not unless configured to do so (with `mempoolreplacement=fee,optin`).

lontivero:
> Core requires BIP 125 signalling to allow replacements. Knots does not unless configured to do so (with mempoolreplacement=fee,optin)

Ohh we didn't know that (or at least I didn't). Thank you.

nopara73:
Thank you. I made a mistake. I double checked with Bitcoin Knots 0.20 and double spend indeed succeeds.

dangershony:
Perhaps to remove confusion its a good idea to assert the type of RCP exception in the test?

PR #889, opened 2020-07-24
Renamed Hashes.Hash256 to Hashes.DoubleSHA256, to make it clear what the method actually does

albracko:
Recently i was trying to verify some payload with the signature and lost a few hours because i wasn't aware that Hash256 method is actually doing a double hash, because the name of the method suggests that it does a single hash.

### DIFF
--- a/NBitcoin/Crypto/Hashes.cs
+++ b/NBitcoin/Crypto/Hashes.cs
@@ -18,24 +18,48 @@ namespace NBitcoin.Crypto
 {
 	public static class Hashes
 	{
-		#region Hash256
+		#region DoubleSHA256
+		[Obsolete("Method name changed to DoubleSHA256(byte[]) to make it clearer what the implementation does.")]
 		public static uint256 Hash256(byte[] data)
 		{
 			return Hash256(data, 0, data.Length);
 		}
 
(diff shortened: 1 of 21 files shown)


nopara73:
Concept ACK.

NBitcoin/Crypto/Hashes.cs
```diff
@@ -18,24 +18,24 @@ namespace NBitcoin.Crypto
 {
 	public static class Hashes
 	{
-		#region Hash256
-		public static uint256 Hash256(byte[] data)
+		#region DoubleSHA256
+		public static uint256 DoubleSHA256(byte[] data)
```
knocte, line 22: this is breaking API; you should leave the old methods there but with an `[Obsolete]` attribute

albracko: Agreed...will change it, to be backward compatible

NicolasDorier:
We can't just rename methods, it will break consumers of NBitcoin.

What I advise you is to keep this commit, but add back the `Hash256` method with an `[Obsolete]` argument explaining it is deprecated and use the DoubleSHA256 instead.
I will clean up the old method after one year or two.

knocte:
@NicolasDorier I already told the same to @albracko and he already addresses the issue in his last push ;)

NicolasDorier:
oops indeed sorry!

PR #893, opened 2020-08-03
Add verbosity option to RPC command `GetBlock`

joemphilips:
bitcoin rpc `getblock` has second option "verbosity", by which
we can get not only the block itself but its metadata
(e.g. height, next block hash) We can now specify it by second option
to the corresponding rpc command.

I need this to get the block height with the block itself by one rpc call.

### DIFF
--- a/NBitcoin.Tests/RPCClientTests.cs
+++ b/NBitcoin.Tests/RPCClientTests.cs
@@ -1855,6 +1855,57 @@ namespace NBitcoin.Tests
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


NicolasDorier:
1. The CI does not build
2. Can you add a `///` doc to explain the verbosity parameter? (or/and use an enum instead of int)

joemphilips:
Fixed.

NicolasDorier:
pushed new version. (5.0.49)

PR #904, opened 2020-08-24
Monkey porting I

lontivero:


NicolasDorier:
@lontivero you may want to take a look at https://github.com/MetacoSA/NBitcoin/pull/696 from @joemphilips

I was planning to merge it, but I don't think anyone wanted to use it immediately, so I kept the PR open waiting for the opportunity.

joemphilips:
I abandoned my attempt because miniscript/output-descriptor spec is continuously have been rebuilt from the scratch. So I'm happy if @lontivero is taking after me. I was exhausted by keep scratching my head around for prototypical miniscript.

My attempt was based on rust-miniscript, by I recommend to port from C++ version. It seemed more concise than rust version.

NicolasDorier:
@joemphilips even the miniscript one if I understood is constantly changing? I thought the output descriptors were kind of stable.

joemphilips:
@NicolasDorier The spec for the miniscript contained new version of output-descriptor and that was not stable when I was playing around. The restricted version of output-descriptor (which we already have in bitcoin-rpc) is stable, I guess.

lontivero:
Ohhh! I didn't know about @joemphilips' PR. I've just scrolled up and down his PR and it looks good (that functional approach in the code is cool btw). I can take it and make some progress on top. I cannot promise anything but i will try.

joemphilips:
Speaking of #696 , this contains the restricted version of output descriptor, which works fine and passes every tests in bitcoin-core. But if i could start from scratch, I will probably not use Parser combinator. functional parser combinator is a great tool in functional-first language such as F# but in C# it is hard to read/debug.

PR #917, opened 2020-09-15
Scalar (ulong) ctor

lontivero:
Adds a new Scalar ctor that accepts ulong. 

Nore: This will be useful for WabiSabi. Of course it is not absolutely necessary and we can perfectly live without this so, feel free to reject it if you think it is not okay for some reason.


NicolasDorier:
Can you add a small test about it?

lontivero:
Done.

NicolasDorier:
I think there is a bug.

NicolasDorier:
ah no that's fine

PR #925, opened 2020-10-02
Update seed and attempt to fix POW for Groestlcoin

gruve-p:


NBitcoin.Altcoins/Groestlcoin.cs
```diff
@@ -277,8 +364,8 @@ protected override NetworkBuilder CreateTestnet()
 			.SetMagic(0x0709110b)
 			.SetPort(17777)
 			.SetRPCPort(17766)
-			.SetName("groestl-test")
-			.AddAlias("groestl-testnet")
+			.SetName("grs-test")
```
NicolasDorier, line 367: this is a breaking change. Are you sure you want to break all software?

gruve-p: Ohh. I was not aware this was a breaking change. Can you elaborate what it breaks? Does it break nbxplorer and btcpayserver? I did the name change because we are trying to get Wasabi wallet work for GRS (https://github.com/Groestlcoin/WalletWasabi) and saw it being used there. If nbxplorer and btcpayserver is affected then I will revert the name change

gruve-p: @NicolasDorier Does this change break nbxplorer and btcpayserver? If so, then I will revert the name change.

NicolasDorier: it may break it yes

gruve-p: ok reverted

PR #926, opened 2020-10-21
Expose MultiExponentiation Method

lontivero:
This PR publishes a new method in the NBitcoin.Secp256k1 library. This is necessary for WabiSabi because it implements a range proof algorithm that involves tens of group element multiplications that makes the algorithm performance really bad. With this exposed method the performance is 10x faster:

```
|   Method | count |         Mean |      Error |       StdDev |       Median | Rank |
|--------- |------ |-------------:|-----------:|-------------:|-------------:|-----:|
| MultiExp |     2 |     64.03 us |   2.127 us |     5.965 us |     62.15 us |    1 |
| ExpMulti |     2 |    745.40 us |  16.678 us |    46.768 us |    728.23 us |    3 |
| MultiExp |    10 |    251.95 us |   4.753 us |     4.668 us |    251.49 us |    2 |
| ExpMulti |    10 |  3,640.58 us |  51.402 us |    42.923 us |  3,633.79 us |    5 |
| MultiExp |   100 |  2,686.49 us |  57.672 us |   167.318 us |  2,632.45 us |    4 |
| ExpMulti |   100 | 36,844.23 us | 720.093 us | 1,503.102 us | 36,264.07 us |    6 |
```

*    MultiExp is `multiexponentiation(IE<Scalar> ns, IE<GEJ> gs)`
*    ExpMulti is `n0 *G0 + n1 * G1 + ....+ nn * Gn `

@nothingmuch could you take a look at this?

### DIFF
--- a/NBitcoin/Secp256k1/ECMultContext.cs
+++ b/NBitcoin/Secp256k1/ECMultContext.cs
@@ -224,7 +224,12 @@ namespace NBitcoin.Secp256k1
 		}
 
 #if SECP256K1_LIB
-		public GEJ Mult(in Span<GEJ> @as, in Span<Scalar> nas)
+		public
+#else
+		internal
+#endif
+
+		GEJ Mult(in Span<GEJ> @as, in Span<Scalar> nas)
 		{
 			var n = @as.Length;
 			unsafe
@@ -244,7 +249,6 @@ namespace NBitcoin.Secp256k1
 				return ECMultiply(state, @as, nas, null, n);
 			}
 		}
-#endif
 
 		unsafe GEJ ECMultiply(in StraussState state, in Span<GEJ> a, in Span<Scalar> na, in Scalar? ng, int num)
 		{
--- a/NBitcoin/Secp256k1/ECMultContext.cs
+++ b/NBitcoin/Secp256k1/ECMultContext.cs
@@ -234,19 +234,26 @@ namespace NBitcoin.Secp256k1
 			var n = @as.Length;
 			unsafe
 			{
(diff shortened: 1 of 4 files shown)


NBitcoin/Secp256k1/ECMultContext.cs
```diff
@@ -222,6 +222,34 @@ public GEJ Mult(in GEJ a, in Scalar na, in Scalar ng)
...
+		GEJ Mult(in Span<GEJ> @as, in Span<Scalar> nas)
+		{
+			var n = @as.Length;
+			unsafe
+			{
+				GEJ* prej = stackalloc GEJ[n * ArraySize_A];
+				FE* zr = stackalloc FE[n * ArraySize_A];
+				GE* pre_a = stackalloc GE[n * ArraySize_A];
+				GE* pre_a_lam = stackalloc GE[n * ArraySize_A];
+				StraussPointState* ps = stackalloc StraussPointState[n];
```
NicolasDorier, line 241: The stack space is limited, you should not allocate unless you precisely know that it can't be more than a bounded size, and small enough. (say less than 500 bytes)

But here we do not control `n`.
I don't think just using heap allocated pointers (either managed or unmanaged), will have big performance impact.

lontivero: Yes, you're right. I understand i have to allocate those arrays in the heap,  am i right? Please take a look at my last commit.

lontivero:
Umm I am hunting a problem. Please **DO NOT** merge

lontivero:
This PR makes no sense.

Issue #927, opened 2020-10-21
AddKeyPath adds bip32_derivs to both inputs and outputs in PSBT


molnard:
CC [was complaining](https://github.com/zkSNACKs/WalletWasabi/issues/4460) about `PSBT` in case of multisig address and didn't sign. 

### Repro
1. Start Wasabi and go tx builder
2. Send to an external Multisig address
3. CC will fail to sign, `PSBT` outputs contain an extra field. 

### PSBT changes during the build

Before `psbt.AddKeyPath`

```
  "inputs": [
    {
      "index": 0,
      "partial_signatures": {
        "censored"
      },
      "witness_utxo": {
        "value": "0.xxxxxxx",
        "scriptPubKey": "0 censored"
      },
      "bip32_derivs": []
    }
  ],
  "outputs": [
    {
      "bip32_derivs": []
    }
  ]
```
After psbt.AddKeyPath

```
  "inputs": [
    {
      "index": 0,
      "partial_signatures": {
        "censored"
      },
      "witness_utxo": {
        "value": "0.xxxxxxx",
        "scriptPubKey": "0 censored"
      },
      "bip32_derivs": [
        {
          "pubkey": "PUBKEY A",
          "master_fingerprint": "FINGERPRINT A",
          "path": "PATH A"
        }
      ]
    }
  ],
  "outputs": [
    {
      "bip32_derivs": [
        {
          "pubkey": "PUBKEY A",
          "master_fingerprint": "FINGERPRINT A",
          "path": "PATH A"
        }
      ]
    }
  ]
```

There should not be an extra output. 

Wasabi PR to hack it around https://github.com/zkSNACKs/WalletWasabi/pull/4601

nopara73:
A 2of5 mainnet address one can reproduce the incorrect behavior with: `bc1qeef3jecqytj8j2xnjzduf5mu9c6jsqwd4hmvyv2zw8hzpf7a47nqrws5sn`.

NicolasDorier:
I do not see an extra output here

NicolasDorier:
I don't see why AddKeyPath should not add the KeyPath to the outputs?

nopara73:
[...]

PR #928, opened 2020-10-22
Add failing PSBT test for script segwit

nopara73:
https://github.com/MetacoSA/NBitcoin/issues/927#issuecomment-714514489


NicolasDorier:
Closing this I am on it, will make a commit as well. I think you should remove your current hack then.

PR #929, opened 2020-10-25
Remove this qualification

yahiheb:


knocte:
Personally, nACK.

NicolasDorier:
yeah I won't merge this one, this is too hard for me to review that it is not calling a local variable instead, while not providing too much value.

PR #932, opened 2020-10-28
Only shuffle the tx builders if both flags are enabled

### LINKED ISSUE #931: Setting TransactionBuilder.ShuffleOutputs to false doesn't prevent output shuffling
## Steps to reproduce:

Create a `TransactionBuilder`, set its `ShuffleOutputs` flag to false, add some outputs using `Send` then call `BuildTransaction`.

## Expected behaviour:

The outputs appear in the order that they were added using `Send`.

## Actual behaviour:

The outputs appear in a random order.

---

Note that it's possible to fix this by also setting `TransactionBuilder.RandomShuffle` to `null`. But that will disable all shuffling (including input shuffling) and it also makes it unnecessary to set `ShuffleOutputs` at all. So as it stands the `ShuffleOutputs` flag is just broken.

From a brief look at the code, it seems that the bug is caused by `TransactionBuilder.BuildTransaction` calling `DoShuffle` at the very start of the function, regardless of whether the `ShuffleOutputs` flag is set. It later shuffles the outputs again depending on whether `ShuffleOutputs` is set, but by that point the outputs have already been shuffled.

aarani:
Fixes #931

### DIFF
--- a/NBitcoin/TransactionBuilder.cs
+++ b/NBitcoin/TransactionBuilder.cs
@@ -1153,7 +1153,7 @@ namespace NBitcoin
 
 		private void DoShuffle()
 		{
-			if (ShuffleRandom != null)
+			if (ShuffleRandom != null && ShuffleOutputs)
 			{
 				Utils.Shuffle(_BuilderGroups, ShuffleRandom);
 				foreach (var group in _BuilderGroups)


NicolasDorier:
Can you instead change the `DoShuffle` method to check if shuffling is authorized and output shuffling as well?

knocte:
@aarani CI is broken now, pls check when you. have time

knocte:
@aarani also in your last push you removed the removal of the DoStuff() method at the beginning, so the bug would still be there.

NBitcoin/TransactionBuilder.cs
```diff
@@ -1153,7 +1153,10 @@ public TransactionBuilder Shuffle()
 
 		private void DoShuffle()
 		{
-			if (ShuffleRandom != null)
+			TransactionBuildingContext ctx = new TransactionBuildingContext(this);
+			if (ShuffleRandom != null &&
+				ShuffleInputs && ctx.CanShuffleInputs &&
```
NicolasDorier, line 1158: remove this, this DoShuffle only shuffle outputs.

NicolasDorier:
The test not passing is me that broke the test. Fixing.

NBitcoin/TransactionBuilder.cs
knocte: then let's rename DoShuffle to DoShuffleOutputs?

knocte:
> @aarani also in your last push you removed the removal of the DoStuff() method at the beginning,

Disregard this bit, I misread the last diff; the bug would be fixed by the last diff but maybe it's more readable to put the two shuffles closer to each other.

NBitcoin/TransactionBuilder.cs
```diff
@@ -1153,7 +1153,9 @@ public TransactionBuilder Shuffle()
 
 		private void DoShuffle()
 		{
-			if (ShuffleRandom != null)
+			TransactionBuildingContext ctx = new TransactionBuildingContext(this);
```
NicolasDorier, line 1156: remove this, this is uneeded.

knocte: @NicolasDorier how can he check ctx.CanShuffleOutputs then? should pass ctx as param?

NicolasDorier:
made new version

PR #935, opened 2020-11-02
TransactionBuilder: shuffle inputs too in Shuffle() method

knocte:
While we are at it, we improve the obsolete message of the Shuffle
method and put the shuffle operations closer together in the
BuildTransaction method.

### DIFF
--- a/NBitcoin/TransactionBuilder.cs
+++ b/NBitcoin/TransactionBuilder.cs
@@ -1144,11 +1144,27 @@ namespace NBitcoin
 			return SendAsset(destination, new AssetMoney(assetId, quantity));
 		}
 
-		[Obsolete("Transaction builder is automatically shuffled")]
+		[Obsolete("Transaction builder is automatically shuffled if its Shuffle* properties are true")]
 		public TransactionBuilder Shuffle()
 		{
-			DoShuffleGroups();
+			DoShuffle();
 			return this;
+		}
+
+		private void DoShuffle(TransactionBuildingContext ctx)
+		{
+			DoShuffleGroups();
+			if (ShuffleRandom != null)
+			{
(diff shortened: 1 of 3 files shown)


NicolasDorier:
No do not do this please, things are working already as it should.

knocte:
> No do not do this please, things are working already as it should.

Can you elaborate? AFAIU the code is much more readable this way, did I break something?

NicolasDorier:
Yes `DoShuffle` != `DoShuffleGroups`. `DoShuffleGroups` does not shuffle inputs by design, because that's why it is called `DoShuffleGroups`

knocte:
My change doesn't make DoShuffleGroups shuffle inputs.

NBitcoin/TransactionBuilder.cs
```diff
@@ -1623,7 +1639,6 @@ public Transaction BuildTransaction(bool sign, SigHash sigHash)
 		public Transaction BuildTransaction(bool sign, SigningOptions signingOptions)
 		{
 			int totalRepass = 5;
-			DoShuffleGroups();
```
NicolasDorier, line 1626: groups are shuffled before the rest, there is a reason for that, probably why tests break.

knocte: But in my PR DoShuffleGroups() is the first thing that DoShuffle() does.

NicolasDorier: No your code is not equivalent, you are not shuffling group before building the transaction, which make it useless.

knocte: ah, so with "the rest" you don't mean "shuffling the rest" but more stuff, got it; sorry for taking your time

PR #940, opened 2020-11-16
BIP155 `sendaddrv2` and `addrv2`

lontivero:
Since September 15th Tor started warning onion service operators that v2 is deprecated. In part because of that Bitcoin Core has merged [PR 19954](https://github.com/bitcoin/bitcoin/pull/19954) that implements [BIP155 addrv2 message](https://github.com/bitcoin/bips/blob/master/bip-0155.mediawiki). The new p2p messages allow share nodes' endpoints other than ip:port (onion v3, I2P, etc).

Currently Wasabi connects to third-party nodes through v2 onion services only because v2 addresses can be encoded as IPv6 addresses (onioncat) what makes the connection transparent because it is the tor client the one that knows how to solve the real onion address.

After onion services v2 are completely removed Wasabi will not be able to connect to third-party nodes any more.
 
### This PR includes some WIP code that:

* signals the preference for `addrv2` messages using `sendaddrv2`.
* adds `addrv2` p2p messages.
* adds SHA3 digest (the `NBitcoin.Altcoins` contains an implementation but moving that it is a big change)
* adds code to keep backward compatibility.

### Problems

The p2p code in NBitcoin uses `IPEndpoint` in many places and it seems we have to use the `NetworkAddress` instead and also know when it needs to use a proxy.

Any commment, suggestion, clue or whatever that could help me to implement this is very much welcome.


nopara73:
@lontivero In Wasabi we're using `EndPoint`
image for local Bitcoin Core P2P connections: IIRC people were pasting onion addresses there and report it still works.

For backend connection we're using `Uri`.

Anyway if your solution works, then whatever I just wanted to note this.

NicolasDorier:
@nopara73 In NBitcoin we use Endpoint as well and `TryParseEndpoint` to parse it. (which support onion already)

NBitcoin.Tests/ProtocolTests.cs
```diff
@@ -320,7 +350,7 @@ private static async Task WaitConnected(NodesGroup group)
 				group.ConnectedNodes.Added -= waitingConnected;
 			};
 			group.ConnectedNodes.Added += waitingConnected;
-			CancellationTokenSource cts = new CancellationTokenSource(5000);
+			CancellationTokenSource cts = new CancellationTokenSource(5000 * 100);
```
molnard, line 353: Dispose

NBitcoin/Protocol/NetworkAddress.cs
```diff
@@ -5,134 +5,868 @@
...
+			SetIp(ip);
 		}
 
-		internal uint ntime;
-		ulong service = 1;
-		byte[] ip = new byte[16];
-		ushort port;
+		public bool IsIPv4 => (Network)network == Network.IPv4;
+		public bool IsIPv6 => (Network)network == Network.IPv6;
+		public bool IsTOR => (Network)network == Network.Onion;
```
molnard, line 132: IsTor

molnard (changes requested):

nopara73:
@lontivero You should be able to make it ready for review (non-draft) since https://github.com/MetacoSA/NBitcoin/pull/944 is merged, right?

lontivero:
Yes. I am trying to make it no so big. Also, sipa changed bitcoin core (it is not merged yet) making this PR incompatible, this means that I cannot use the rc2 to test this, I need to compile core and use that.

lontivero:
@nopara73 @NicolasDorier this is ready for review (the failing test is failing for some unrelated reason).

After this PR I need to make a few small changes like signaling `sendaddrv2` before `VerAck` message (this changes the order of the handshake messages), clean up a few duplicated code, increase the protocol version to 70016 and make sure Tor addresses are routed through Tor (I didn't check it).

NicolasDorier:
Please do not change types around, that's too much breaking changes.
Keep `NetworkAddress`, I don't understand what `Service` is.
You can modify `NetworkAddress` to support your changes.

Do not change the IPEndpoint stuff in Node class. Keep the current behavior for what is not Torv3, if it is Torv3 then just throw an exception like InvalidOperationException.

lontivero:
Copied. I will implement everything in `NetworkAddress` then and avoid changes in `Node`. I am closing this PR and I will open other/s.

Issue #941, opened 2020-11-25
MaybeNullWhen is not needed anymore (causes errors)


dangershony:
Hi MaybeNullWhenAttribute causes errors for projects that enable Nullable
https://github.com/MetacoSA/NBitcoin/blob/973fd31be570cb460518ffa268fd7b9c62b44105/NBitcoin/NullableShims.cs

This type exists in the namespace already under `System.Diagnostics.CodeAnalysis` maybe it can now be removed?

nopara73:
Interesting, we haven't encounter the errors it causes.

dangershony:
This is the error, a collision of ns types
```

Error	CS0433	The type 'MaybeNullWhenAttribute' exists in both 
'NBitcoin, Version=5.0.65.0, Culture=neutral, PublicKeyToken=null' and 
'System.Runtime, Version=4.2.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'	```

nopara73:
Anyhow it seems to be functionally the same.

NicolasDorier:
@dangershony in which target you have the problem? It does not always exists in `System.Diagnostics.CodeAnalysis`.

NicolasDorier:
Find something that work @dangershony I can't build with it on my side :/

NicolasDorier:
I can change the namespace though one moment.

NicolasDorier:
ah no if I change namespace it is other problem to fix everywhere. If I put in internal maybe it works? Can you try @dangershony

dangershony:
Hi @NicolasDorier for now I just stopped using he attribute (I can survive without it) but it will become a problem n the future I suspect.
Compiling against netcoreapp3.1 and NBitcoin Version=5.0.65

Perhaps internal will work indeed 
> If I put in internal maybe it works? Can you try

How do I try? ah I can reference NBitcoin locally and check yeah let me try and get back

PR #942, opened 2020-11-26
Change MaybeNullWhenAttribute to internal

dangershony:
This is to avoid collision with the diagnostics ns
Seems to not break the secp256k1 tests


dangershony:
Fixes #941

nopara73 (approved):
LGTM

NicolasDorier:
pushed .67

PR #944, opened 2020-11-30
Update NBitcoin to support RPC changes

lontivero:
## Update NBitcoin to support RPC changes in Bitcoin Core 0.21.0.

Bitcoin core doesn't start a default wallet anymore and for that reason all the RPC tests that need a wallet fail. The solution implemented in this PR is to create an empty wallet called `wallet.dat` using the `bitcoin-wallet` command line tool.

Additionally I have created a new interface called `IRPCWalletClient` that improves usability of the rpc interface when we need to invoke wallet related rpc calls because it lists only the wallet-related calls.

The batch rpc can only work with one wallet at a time, that restriction is not new and it is because the wallet name is parts of the URL.

 I added the `CreateWallet`, `LoadWallet` and `UnloadWallet` rpc calls.

--------------

It would be good if this PR can be reviewed and merged (after all the recommendations/suggestions/fixes) because this is needed to test the other PR https://github.com/MetacoSA/NBitcoin/pull/940 given it needs 0.21.0 which implements the bip155.

### DIFF
--- a/NBitcoin.TestFramework/NodeBuilder.cs
+++ b/NBitcoin.TestFramework/NodeBuilder.cs
@@ -582,8 +582,10 @@ namespace NBitcoin.Tests
 			{
 				UseShellExecute = _Builder.ShowNodeConsole
 			};
-			using var walletToolProcess = Process.Start(info); 
-			walletToolProcess.WaitForExit();
+			using (var walletToolProcess = Process.Start(info))
+			{ 
+				walletToolProcess.WaitForExit();
+			}
 		}
 
 		Process _Process;
--- a/NBitcoin.TestFramework/NodeBuilder.cs
+++ b/NBitcoin.TestFramework/NodeBuilder.cs
@@ -578,17 +578,11 @@ namespace NBitcoin.Tests
 			var walletToolPath = Path.Combine(Path.GetDirectoryName(this._Builder.BitcoinD), "bitcoin-wallet");
 			string walletToolArgs = $"-regtest -wallet=\"wallet.dat\" -datadir=\"{dataDir}\" create";
 
-			Process walletToolProcess; 
-			if (_Builder.ShowNodeConsole)
+			var info = new ProcessStartInfo(walletToolPath, walletToolArgs)
 			{
-				ProcessStartInfo info = new ProcessStartInfo(walletToolPath, walletToolArgs);
-				info.UseShellExecute = true;
-				walletToolProcess = Process.Start(info);
-			}
-			else
-			{
-				walletToolProcess = Process.Start(walletToolPath, walletToolArgs);
-			}
+				UseShellExecute = _Builder.ShowNodeConsole
+			};
+			using var walletToolProcess = Process.Start(info); 
 			walletToolProcess.WaitForExit();
 		}
 
--- a/NBitcoin.TestFramework/NodeBuilder.cs
+++ b/NBitcoin.TestFramework/NodeBuilder.cs
@@ -100,6 +100,11 @@ namespace NBitcoin.Tests
 		}
 		public string RegtestFolderName { get; set; }
 
+		public bool CreateWallet
+		{
+			get; set;
+		}
+
 		/// <summary>
 		/// For blockchains that use an arbitrary chain (e.g. instead of main, testnet and regtest
 		/// Elements can use chain=elementsregtest).
@@ -286,14 +291,6 @@ namespace NBitcoin.Tests
 			_Config = Path.Combine(dataDir, "bitcoin.conf");
 			ConfigParameters.Import(builder.ConfigParameters, true);
 			ports = new int[2];
-
-			if (Version.TryParse(builder.NodeImplementation.Version, out var version))
-			{
-				if (builder.Network.NetworkSet is Bitcoin && version >= new Version(0, 21))
-				{
-					RequiresWalletCreation = true;
-				}
-			}
 
 			if (builder.CleanBeforeStartingNode && File.Exists(_Config))
 			{
@@ -508,7 +505,7 @@ namespace NBitcoin.Tests
 					configStr.AppendLine($"[{NodeImplementation.Chain}]");
 				}
 			}
-			if (RequiresWalletCreation)
+			if (NodeImplementation.CreateWallet)
 				config.Add("wallet", "wallet.dat");
 
 			config.Add("rest", "1");
@@ -543,7 +540,7 @@ namespace NBitcoin.Tests
 		{
 			lock (l)
 			{
-				if (RequiresWalletCreation)
+				if (_Builder.NodeImplementation.CreateWallet)
 					CreateDefaultWallet();
 
 				string appPath = new FileInfo(this._Builder.BitcoinD).FullName;
@@ -679,12 +676,6 @@ namespace NBitcoin.Tests
 			get;
 			set;
 		} = true;
-
-		public bool RequiresWalletCreation
-		{
-			get;
-			set;
-		} = false;
 
 		class TransactionNode
 		{
--- a/NBitcoin.TestFramework/NodeBuilder.cs
+++ b/NBitcoin.TestFramework/NodeBuilder.cs
@@ -287,6 +287,14 @@ namespace NBitcoin.Tests
 			ConfigParameters.Import(builder.ConfigParameters, true);
 			ports = new int[2];
 
+			if (Version.TryParse(builder.NodeImplementation.Version, out var version))
+			{
+				if (builder.Network.NetworkSet is Bitcoin && version >= new Version(0, 21))
+				{
+					RequiresWalletCreation = true;
+				}
+			}
+
 			if (builder.CleanBeforeStartingNode && File.Exists(_Config))
 			{
 				var oldCreds = ExtractCreds(File.ReadAllText(_Config));
@@ -500,7 +508,9 @@ namespace NBitcoin.Tests
 					configStr.AppendLine($"[{NodeImplementation.Chain}]");
 				}
 			}
-			config.Add("wallet", "wallet.dat");
+			if (RequiresWalletCreation)
+				config.Add("wallet", "wallet.dat");
+
 			config.Add("rest", "1");
 			config.Add("server", "1");
 			config.Add("txindex", "1");
@@ -533,7 +543,8 @@ namespace NBitcoin.Tests
 		{
 			lock (l)
 			{
-				CreateDefaultWallet();
+				if (RequiresWalletCreation)
+					CreateDefaultWallet();
 
 				string appPath = new FileInfo(this._Builder.BitcoinD).FullName;
 				string args = "-conf=bitcoin.conf" + " -datadir=" + dataDir + " -debug=net";
@@ -668,6 +679,12 @@ namespace NBitcoin.Tests
 			get;
 			set;
 		} = true;
+
+		public bool RequiresWalletCreation
+		{
+			get;
+			set;
+		} = false;
 
 		class TransactionNode
 		{
--- a/NBitcoin.TestFramework/NodeBuilder.cs
+++ b/NBitcoin.TestFramework/NodeBuilder.cs
@@ -500,6 +500,7 @@ namespace NBitcoin.Tests
 					configStr.AppendLine($"[{NodeImplementation.Chain}]");
 				}
 			}
+			config.Add("wallet", "wallet.dat");
 			config.Add("rest", "1");
 			config.Add("server", "1");
 			config.Add("txindex", "1");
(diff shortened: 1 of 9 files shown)


knocte:
Why not make the commit message have the same nice descriptive text of the PR? It will help when people use git-blame.

nopara73:
image

knocte:
Thanks :) but commit messages have a body too, not just a title :P

NicolasDorier:
Your fix will break the altcoin tests. (we don't run them in CI)

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -91,40 +93,281 @@ wallet			 getbalance
...
+			};
+		}
+
+		public async Task<IRPCWalletClient> CreateWalletAsync(string walletNameOrPath, 
+			bool disablePrivateKeys = false, 
+			bool blank = false, 
+			string passphrase = null, 
+			bool avoidReuse = false,
+			bool descriptors = false,
+			bool? loadOnStartup = null)
```
NicolasDorier, line 330: please use a structure for that with proper nullables.

The defaults change from one release to the other, so we should not try to guess them.

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -91,40 +93,281 @@ wallet			 getbalance
...
 		wallet			 walletlock
 		wallet			 walletpassphrasechange
-		wallet			 walletpassphrase			yes
+		wallet			 walletpassphrase			Yes
 		wallet			 walletprocesspsbt
 		wallet			 walletcreatefundedpsbt
 	*/
-	public partial class RPCClient
+
+	public interface IRPCWalletClient
```
NicolasDorier, line 129: I really don't like this approach for interfaces.

Interfaces with many methods are a code smell, and everytime we add one method it will be a breaking change, I would prefer you remove the interface.

lontivero:
I've made the changes and made sure it doesn't break the altcoins tests. The solution requires a flag to decide whether it needs to create the wallet or not. If anyone have a better idea please let me know and I will implement it.

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -696,6 +782,28 @@ public OutPoint[] ListLockUnspent()
 			return ListLockUnspentAsync().GetAwaiter().GetResult();
 		}
 
+		// abandom transaction
```
yahiheb, line 785:
```suggestion
		// abandon transaction
```

yahiheb (changes requested):

NBitcoin.TestFramework/NodeBuilder.cs
```diff
@@ -647,6 +680,12 @@ public bool CookieAuth
 			set;
 		} = true;
 
+		public bool RequiresWalletCreation
```
NicolasDorier, line 683: Can you instead add this property to `NodeDownloadData`?

NBitcoin.TestFramework/NodeBuilder.cs
```diff
@@ -287,6 +287,14 @@ public CoreNode(string folder, NodeBuilder builder)
 			ConfigParameters.Import(builder.ConfigParameters, true);
 			ports = new int[2];
 
+			if (Version.TryParse(builder.NodeImplementation.Version, out var version))
```
NicolasDorier, line 290: not needed if https://github.com/MetacoSA/NBitcoin/pull/944/files#r533904805

lontivero:
Done.

NBitcoin.TestFramework/NodeBuilder.cs
```diff
@@ -100,6 +100,11 @@ public NodeOSDownloadData Windows
 		}
 		public string RegtestFolderName { get; set; }
 
+		public bool CreateWallet
```
NicolasDorier, line 103: unsused, remove it

NBitcoin/RPC/CreateWalletOptions.cs
```diff
@@ -0,0 +1,16 @@
+#nullable enable
+
+namespace NBitcoin.RPC
+{
+	public class CreateWalletOptions
+	{
+		public bool DisablePrivateKeys { get; set; }
```
NicolasDorier, line 7: In the createwallet RPC call, all optional parameters should be nullable here.

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -91,40 +93,124 @@ wallet			 getbalance
...
+			if (options is { })
+			{
+				parameters.Add("disable_private_keys", options.DisablePrivateKeys.ToString());
+				parameters.Add("blank", options.Blank.ToString());
+				if (!string.IsNullOrEmpty(options.Passphrase))
+					parameters.Add("passphrase", options.Passphrase);
+
+				parameters.Add("avoid_reuse", options.AvoidReuse.ToString());
+				parameters.Add("descriptors", options.Descriptors.ToString());
+				if (options.LoadOnStartup.HasValue)
```
NicolasDorier, line 177: prefer the pattern `options.LoadOnStartup is bool v`

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -91,40 +93,124 @@ wallet			 getbalance
...
+			var parameters = new Dictionary<string, object>();
+			parameters.Add("wallet_name", walletNameOrPath);
+			if (options is { })
+			{
+				parameters.Add("disable_private_keys", options.DisablePrivateKeys.ToString());
+				parameters.Add("blank", options.Blank.ToString());
+				if (!string.IsNullOrEmpty(options.Passphrase))
+					parameters.Add("passphrase", options.Passphrase);
+
+				parameters.Add("avoid_reuse", options.AvoidReuse.ToString());
```
NicolasDorier, line 175: all optional parameters should be nullable and check it.

NBitcoin/RPC/RPCClient.cs
```diff
@@ -1183,7 +1183,7 @@ public async Task<PeerInfo[]> GetPeersInfoAsync()
 					StartingHeight = (int)peer["startingheight"],
 					SynchronizedBlocks = (int)peer["synced_blocks"],
 					SynchronizedHeaders = (int)peer["synced_headers"],
-					IsWhiteListed = (bool)peer["whitelisted"],
+					IsWhiteListed = peer["whitelisted"] != null ? (bool)peer["whitelisted"] : false,
```
NicolasDorier, line 1186: `is bool v ? v : false`

NBitcoin/RPC/RPCClient.cs
```diff
@@ -2279,7 +2279,10 @@ public async Task<uint256[]> GenerateAsync(int nBlocks)
 					var result = (JArray)(await SendCommandAsync(RPCOperations.generate, nBlocks).ConfigureAwait(false)).Result;
 					return result.Select(r => new uint256(r.Value<string>())).ToArray();
 				}
-				catch (RPCException rpc) when (rpc.RPCCode == RPCErrorCode.RPC_METHOD_DEPRECATED || rpc.RPCCode == RPCErrorCode.RPC_METHOD_NOT_FOUND)
+				catch (RPCException rpc) when (
+					rpc.RPCCode == RPCErrorCode.RPC_METHOD_DEPRECATED 
+					|| rpc.RPCCode == RPCErrorCode.RPC_METHOD_NOT_FOUND 
+					|| (rpc.RPCCode == RPCErrorCode.RPC_MISC_ERROR && rpc.Message.StartsWith("generate\nhas been replaced by the -generate cli option")))
```
NicolasDorier, line 2285: please remove the check `StartsWith`, this is something that will break easily.

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -696,6 +782,28 @@ public OutPoint[] ListLockUnspent()
...
 
+		// abandon transaction
+
+		/// <summary>
+		/// Marks a transaction and all its in-wallet descendants as abandoned which will allow
+		/// for their inputs to be respent.
+		/// </summary>
+		/// <param name="txId">the transaction id to be marked as abandoned.</param>
+		public void AbandonTransaction(uint256 txId)
+		{
```
NicolasDorier, line 793: ArgumentNullException if txId is null

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -696,6 +782,28 @@ public OutPoint[] ListLockUnspent()
...
+			SendCommand(RPCOperations.abandontransaction, txId.ToString());
+		}
+
+		/// <summary>
+		/// Marks a transaction and all its in-wallet descendants as abandoned which will allow
+		/// for their inputs to be respent.
+		/// </summary>
+		/// <param name="txId">the transaction id to be marked as abandoned.</param>
+		public async Task AbandonTransactionAsync(uint256 txId)
+		{
```
NicolasDorier, line 803: ArgumentNullException if txId is null

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -91,40 +93,124 @@ wallet			 getbalance
...
+		}
+
+		public async Task<RPCClient> LoadWalletAsync(string filename, bool? loadOnStartup = null)
+		{
+			var result =  await SendCommandAsync(RPCOperations.loadwallet, filename, loadOnStartup).ConfigureAwait(false);
+			return GetWallet(result.Result.Value<string>("name"));
+		}
+
+		public RPCClient LoadWallet(string filename, bool? loadOnStartup = null)
+		{
```
NicolasDorier, line 196: ArgumentNullException if filename is null

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -91,40 +93,124 @@ wallet			 getbalance
...
+				Capabilities = Capabilities,
+				RequestTimeout = RequestTimeout,
+				_HttpClient = _HttpClient,
+				AllowBatchFallback = AllowBatchFallback
+			};
+		}
+
+		public async Task<RPCClient> CreateWalletAsync(string walletNameOrPath, CreateWalletOptions? options = null)
+		{
+			var parameters = new Dictionary<string, object>();
```
NicolasDorier, line 166: ArgumentNullException if walletNameOrPath is null

NBitcoin.TestFramework/NodeBuilder.cs
lontivero: It is not unused.

NBitcoin/RPC/RPCClient.cs
lontivero: Yes, it sucks really bad. I didn't find any other way to check for that specific error. I will remove it.

NBitcoin/RPC/RPCClient.cs
lontivero: I prefer that style too but in this case that doesn't compile and this is the style that is already used in this same method.

molnard (changes requested):

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -91,40 +93,135 @@ wallet			 getbalance
...
 		wallet			 walletprocesspsbt
 		wallet			 walletcreatefundedpsbt
 	*/
+
 	public partial class RPCClient
 	{
-		// backupwallet
+		#nullable enable
+
+		public RPCClient GetWallet(string walletName)
```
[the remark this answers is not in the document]
lontivero: That's why I created the `IRCPWalletClient` with only the rpc methods that involve Wallet operations but @NicolasDorier and @nopara73 think it is a bad idea so I remove it after all the only benefit is the intellisense and the semantic in methods like this one that could return a  `public IRCPWalletClient GetWallet(string walletName)`. That is probably no enough to justify the new interface and the maintainability costs.

NBitcoin.TestFramework/NodeBuilder.cs
```diff
@@ -561,6 +573,24 @@ private async Task Run()
 			}
 		}
 
+		private void CreateDefaultWallet()
+		{
+			var walletToolPath = Path.Combine(Path.GetDirectoryName(this._Builder.BitcoinD), "bitcoin-wallet");
+			string walletToolArgs = $"-regtest -wallet=\"wallet.dat\" -datadir=\"{dataDir}\" create";
+
+			Process walletToolProcess;
```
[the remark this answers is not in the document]
lontivero: It is not necessary because the destructor disposes the handler. I will do it anyway.

lontivero:
Done.

PR #948, opened 2020-12-09
Make sure to clean the key material

lontivero:
The `Key` type was not always `IDisposable` and for that reason I would expect many projects do not dispose the keys explicitly nor using an `using` statement neither. However, I think it is important to clean the key from memory so keys are not leaked by a core crash dump or other means.

This PR implements a destructor/finalizer that disposes the private keys in case the `Dispose` is not invoked, what I think could be the case in some projects.

### DIFF
--- a/NBitcoin/Secp256k1/ECPrivKey.cs
+++ b/NBitcoin/Secp256k1/ECPrivKey.cs
@@ -879,9 +879,23 @@ namespace NBitcoin.Secp256k1
 			return true;
 		}
 
+		protected virtual void Dispose(bool disposing)
+		{
+			if (this.cleared)
+				return;
+				
+			Clear();
+		}
+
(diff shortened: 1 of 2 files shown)


NBitcoin/Secp256k1/ECPrivKey.cs
```diff
@@ -879,11 +879,25 @@ private bool secp256k1_eckey_privkey_tweak_mul(ref Scalar key, in Scalar tweak)
...
-		public void Dispose()
+		protected virtual void Dispose(bool disposing)
 		{
+			if (this.cleared)
+				return;
+				
 			Clear();
 		}
 
+		public void Dispose()
```
NicolasDorier, line 890: don't do this pattern, this is for unmanaged resources.

PR #964, opened 2021-01-28
Fix to respect NONATIVEHASH in Crypto.Hashes.cs.

JVanloofsvelt:
NONATIVEHASH was not respected everywhere, causing the project to fail unexpectedly in a target platform (Blazor wasm) where System.Security.Cryptography is not supported.

Added a missing overload for HMACSHA512, and some small fixes to restore the build.

### DIFF
--- a/NBitcoin/Crypto/Hashes.cs
+++ b/NBitcoin/Crypto/Hashes.cs
@@ -1,4 +1,4 @@
-﻿#if !NO_BC
+#if !NO_BC
 using NBitcoin.BouncyCastle.Crypto.Parameters;
 using NBitcoin.BouncyCastle.Security;
 #endif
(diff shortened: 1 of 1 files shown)


NBitcoin/Crypto/Hashes.cs
```diff
@@ -801,17 +801,29 @@ public static byte[] HMACSHA512(byte[] key, byte[] data)
...
 			var mac = new NBitcoin.BouncyCastle.Crypto.Macs.HMac(new Sha512Digest());
 			mac.Init(new KeyParameter(key));
-			mac.Update(data);
+			mac.BlockUpdate(data, 0, data.Length);
 			byte[] result = new byte[mac.GetMacSize()];
 			mac.DoFinal(result, 0);
 			return result;
 		}
+#if HAS_SPAN
+		public static bool HMACSHA512(byte[] key, ReadOnlySpan<byte> data, Span<byte> output, out int outputLength)
```
NicolasDorier, line 810: `ReadOnlySpan<byte> key`

NicolasDorier, line 810: the code does not seems correct

NicolasDorier: fixed it on https://github.com/MetacoSA/NBitcoin/commit/9326fdb18f1ecdb2388f1136440ab17315de9483

PR #965, opened 2021-01-30
Adapt Hashes.HMACSHA256() so that it uses native hashes when they are…

JVanloofsvelt:
… present, even if native HMAC is not.

If there is no native HMACSHA256, but there is native SHA256, then use Bouncy Castle's HMAC implementation in combination with the native Sha256 digest (instead of Bouncy Castle's SHA256 implementation).


JVanloofsvelt:
Using following compilation symbols to get a build that works in Blazor wasm:
NO_NATIVE_HMACSHA512;NO_NATIVESHA1;NO_NATIVE_RFC2898_HMACSHA512;NOCUSTOMSSLVALIDATION;NO_NATIVERIPEMD160;NETCORE;HAS_SPAN;NO_BC

The gist being that you need to add NO_NATIVE_HMACSHA512 and NO_NATIVE_RFC2898_HMACSHA512.

NBitcoin/Crypto/Hashes.cs
```diff
@@ -12,6 +12,8 @@
 
 #if !NONATIVEHASH
 using System.Security.Cryptography;
+using NBitcoin.Crypto.digests; // use managed SHA implementations
+using NBitcoin.BouncyCastle.Crypto.Parameters;  // we'll need this to create HMACs based o the managed SHAs.
```
knocte, line 16: introduced 1 warning here: `The using directive for 'NBitcoin.BouncyCastle.Crypto.Parameters' appeared previously in this namespace`

JVanloofsvelt: I'll fix it

JVanloofsvelt:
I have no idea why the tests timed out on .NET 4.6

NicolasDorier:
Remove changes on `ECPrivKey.cs` I fixed properly on  58307cca5f5520a306e6f63bb7cc1f20196a380b

NBitcoin/Crypto/NativeDigests/ManagedSha256Digest.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		public void BlockUpdate(byte[] input, int inOff, int length)
+		{
+			this.input = new byte[length];
+			Array.Copy(input, inOff, this.input, 0, length);
+		}
+
+		public int DoFinal(byte[] output, int outOff)
+		{
+			var hash = nativeSha256.ComputeHash(input, 0, input.Length);
```
NicolasDorier, line 35: Can you instead rely on NBitcoin.Secp256k1.SHA256 class which properly implement DoFinal and BlockUpdate and is widely tested and memory efficient?

NBitcoin/Crypto/Hashes.cs
```diff
@@ -832,23 +832,60 @@ public static byte[] HMACSHA256(byte[] key, byte[] data)
...
+				return false;
+
+			outputBuffer.CopyTo(output);
+			return true;
+		}
+#endif
+
+		public static byte[] HMACSHA256(byte[] key, byte[] data)
+		{
+			var mac = new NBitcoin.BouncyCastle.Crypto.Macs.HMac(new ManagedSha256Digest());
```
NicolasDorier, line 862: Can you dispose all the `ManagedSha256Digest` you instanciate?

NBitcoin/Crypto/NativeDigests/ManagedSha256Digest.cs
JVanloofsvelt: Right, I'll take care of it tonight.

JVanloofsvelt: @NicolasDorier I found the NBitcoin.Secpk1.SHA class, but it doesn't look like it implements the IDigest interface, nor the BlockUpdate/DoFinal methods.

NBitcoin/Crypto/Hashes.cs
JVanloofsvelt: I made another update to dispose these.

knocte:
There are conflicts, you need to rebase.

JVanloofsvelt:
@knocte I rebased

NBitcoin/Crypto/NativeDigests/ManagedSha256Digest.cs
NicolasDorier: I am not saying to use it instead, but to use it rather than that `nativeSha256`

NicolasDorier:
https://github.com/MetacoSA/NBitcoin/pull/965#discussion_r570892153

NicolasDorier:
Ping @JVanloofsvelt

NBitcoin/Crypto/NativeDigests/ManagedSha256Digest.cs
JVanloofsvelt: Hello @NicolasDorier, I understand now. There is no NBitcoin.Secp512k1.SHA512 equivalent in the Secpk1 project though (to go along with the SHA256). Besides, wouldn't it be great if NBitcoin could rely on the framework's builtins? 

It's not clear to me if DoFinal() and BlockUpdate() are BouncyCastle-specific interface methods, or if they are a more wider known concept. I'm not sure how to implement them correctly. I was hoping someone here would review it, and fill in the gaps for me.

PR #968, opened 2021-02-03
Allow restrict connections depending their height

lontivero:
This PR is for allowing clients to restrict connections to nodes that are at a minimum height (they can be in the synchronizing, for example). Clients that need to download mostly recent blocks could be interested in connecting only with those nodes that have more blocks than what they already know.

For example, here we can see nodes with different heights (one is synchronizing):

image

### DIFF
--- a/NBitcoin.Tests/ProtocolTests.cs
+++ b/NBitcoin.Tests/ProtocolTests.cs
@@ -277,6 +277,68 @@ namespace NBitcoin.Tests
 
 		[Fact]
 		[Trait("Protocol", "Protocol")]
+		public void CanHandshakeRestrictNodes()
+		{
+			using (var builder = NodeBuilderEx.Create())
+			{
+				var node = builder.CreateNode(true);
+				var manager = new AddressManager();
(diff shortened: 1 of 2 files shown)


knocte:
Why don't you make the commit message the same as the PR description? It can have a body too, not just title. This would help immensely to people that use git blame.

lontivero:
> Why don't you make the commit message the same as the PR description? 

Because messages in the commits are to explain the commits while the message in the PR is to describe the intention of the PR. For example, in a PR description I can write something like "This is probably an stupid idea, do you maintainer think I should write a test? Please feel free to whatever" but that's not something i would put in a commit.

knocte:
> in a PR description I can write something like "This is probably an stupid idea, do you maintainer think I should write a test? Please feel free to whatever" but that's not something i would put in a commit.

I agree with the particular example you just have mentioned, but this case doesn't match with this example. There's useful information in this PR description that would be great to have in the commit message. If you don't do it: 1) people using git blame will not see it; 2) if this repo moves away from github in the future, the information is lost forever too

PR #969, opened 2021-02-05
Key: add helper method GetSeedBytes

knocte:
My team uses this in more than one place and would be good to have it in NBitcoin out of the box.


NBitcoin/Key.cs
```diff
@@ -407,6 +408,18 @@ public BitcoinSecret GetWif(Network network)
 			return new BitcoinSecret(this, network);
 		}
 
+		public byte[] GetSeedBytes()
```
NicolasDorier, line 411: Call it `ToBytes` and follow how I implemented it in `PubKey`.
Your version is super inefficient.

NBitcoin/Key.cs
```diff
@@ -407,6 +408,18 @@ public BitcoinSecret GetWif(Network network)
 			return new BitcoinSecret(this, network);
 		}
 
+		public byte[] GetSeedBytes()
+		{
+			AssertNotDiposed();
+			byte[] bytes = Enumerable.Repeat((byte)0x00, KEY_SIZE).ToArray();
```
NicolasDorier, line 414: `new byte[32]` is  clearly overrated

NBitcoin/Key.cs
knocte: Ah sure

NBitcoin/Key.cs
knocte: oops! didn't realize that the bytes would be overriden later anyway, was just monkey-translating F# code to C# code from somewhere else

knocte:
@NicolasDorier hey, I just realized that there already is a .ToBytes() method for Key, however:

1. It's not really in Key class, but it's an extension method of IBitcoinSerializable. Is it enough to have it there, or is it better that it lives in Key class for better visibility? (so that you don't need to import IBitcoinSerializable's namespace) We could even call the extension method from Key.ToBytes() but this leads me to the next point:
2. Is the extension method as bad (performance wise) as my initial approach? If yes, I'll just go and do what you suggested (use same approach there is in PubKey).

Let me know what's best please.

NicolasDorier:
> It's not really in Key class, but it's an extension method of IBitcoinSerializable. Is it enough to have it there, or is it better that it lives in Key class for better visibility?

Yes.

> Is the extension method as bad (performance wise) as my initial approach? If yes, I'll just go and do what you suggested (use same approach there is in PubKey).

Yes it is bad, that said in practice this should hardly matter but I think just using the PubKey approach is also simpler.

queilawithaQ (approved):

knocte:
> Yes.

Sorry, if I ask (A or B)? you cannot answer "Yes" unless you're trying to confuse me haha.

NicolasDorier:
sorry lol, I think we should not add `GetSeedBytes` when the extension method already have a `ToBytes` that works.
This is very confusing for users, as they would not know which one to call.

julioAk47 (approved):

PR #972, opened 2021-02-10
update althash genesis hex and reverse magic

denuoweb:
Sorry about the number of commits here, I finally got NBXplorer running correctly, now to pass these tests. Thanks for providing this package!


NicolasDorier:
there is conflicts

PR #976, opened 2021-02-24
Allow setting version number with TransactionBuilder

canndrew:
This PR allows setting the version number of a transaction built using `TransactionBuilder` via a new `TransactionBuilder.SetVersion(uint version)` method.

A test for this has also been added.


knocte:
@NicolasDorier thanks for merging, can you bump&release?

NicolasDorier:
done

PR #982, opened 2021-03-19
Key.Parse - network -> expectedNetwork

molnard:
I was confused if this is an option or a constraint. It is cleaner in this way. 

`Network.Parse` uses the same naming which is `expectedNetwork`.


lontivero:
After Named Parameters were introduced in C# this kind of improvement are potential breaking changes, this means that it could break those projects that use this method as follow:

```c#
var sk = Key.Parse(wif: aliceKey, network: Network.Main);
```

I know that is extremely unlikely but it is something we have to have in mind.

Anyway, if you change this method then you should also change the other `Parse` method because after this PR they will be different.

nopara73 (approved):
Good point by @lontivero, but it's not silently breaking and the fix is easy.

PR #989, opened 2021-04-08
Unload Method update

methrammar:


NicolasDorier:
Can you instead keep unload but mark it with `[Obsolete("Use UnloadWallet instead")]` ?

NicolasDorier:
@methrammar sure feel free to do on others.

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -209,14 +221,14 @@ public RPCClient LoadWallet(string filename, bool? loadOnStartup = null)
 			return GetWallet(result.Result.Value<string>("name"));
 		}
 
-		public async Task UnloadAsync(bool? loadOnStartup = null)
+		public async Task UnloadWalletAsync(string filename, bool? loadOnStartup = null)
```
NicolasDorier, line 224: it is `walletName` not `fileName`.

NBitcoin/RPC/RPCClient.Wallet.cs
```diff
@@ -209,14 +221,14 @@ public RPCClient LoadWallet(string filename, bool? loadOnStartup = null)
...
 		{
-			var result = await SendCommandAsync(RPCOperations.loadwallet, loadOnStartup).ConfigureAwait(false);
+			var result = await SendCommandAsync(RPCOperations.loadwallet, filename, loadOnStartup).ConfigureAwait(false);
```
NicolasDorier, line 226: If `loadOnStartup` is `null`, then the best is to call `SendCommandAsync` without it, leaving the default to bitcoin core rather than passing null.

We need to do this to protect against breaking change which might come from bitcoin core. (and also increase compatiblity of RPC Client with other shitcoins)

NicolasDorier:
My apologize @methrammar the `UnloadWallet` could not possibly work before so we don't have to make it obsolete.

NBitcoin.Tests/RPCClientTests.cs
```diff
@@ -2102,12 +2102,12 @@ public void ShouldCreateLoadAndUnloadWallet()
 
 			var wallet0 = rpc.CreateWallet("w0");
 			var address = wallet0.GetNewAddress();
-			wallet0.Unload();
+			wallet0.UnloadWallet("w0");
 			Assert.Throws<RPCException>(()=> wallet0.GetNewAddress());
 
 			wallet0 = rpc.LoadWallet("w0");
 			address = wallet0.GetNewAddress();
-			wallet0.Unload();
```
NicolasDorier, line 2110: This should not change, `w0` is implied by the rpc path. This delete the right wallet.

NBitcoin/RPC/RPCClient.Wallet.cs
NicolasDorier: I noticed I did not do it for the other methods. Will do.

NicolasDorier:
I implemented your idea on https://github.com/MetacoSA/NBitcoin/commit/d41d70168434fab83110495a84dde3e7f4d5e13b
As well as cleaning up Load/UnloadWallet to be more breaking change proof. Added tests as well.

Issue #977, opened 2021-03-02
DNS seed queries are not done over Tor 


lontivero:
It would be good to query dns seeds using Tor. https://www.reddit.com/r/WasabiWallet/comments/lvld6s/wasabi_not_going_over_tor/

NicolasDorier:
It should if `SocksSettingsBehavior.OnlyForOnionHosts` is set to `true`.

NicolasDorier:
Can see the code related to this at `DefaultEndpointConnector`

lontivero:
I didn't explain myself clearly enough. What I say is that NBitcoin queries the DNS servers in the clearnet, this is done with `Dns.GetHostAddressesAsync(dns.Host)`. 

Sadly, I don't see an easy way to route those request through a proxy.

NicolasDorier:
ah shit I see. Yeah, I don't see how we can do that.
That said NBitcoin could disallow DNS endpoints if Tor must be used.
Those DNS seed are not really useful, most seeds are ips, hard coded to bitcoin.

DemonRx:
I don't know if this might be of much help, but I think you need to have the seeds actually running over Tor as well.

Refer to this article: https://developers.cloudflare.com/1.1.1.1/fun-stuff/dns-over-tor

nopara73:
One can ask Tor to resolve DNS through the socks port. See an old C# implementation of mine: https://github.com/nopara73/DotNetTor/blob/fc438daa573145b12d4ae1d93364797858bc244f/src/DotNetTor/TorSocks5Client.cs#L391

Spec: https://gitweb.torproject.org/torspec.git/tree/socks-extensions.txt#n44

> As an extension to SOCKS4A and SOCKS5, Tor implements a new command value,
  "RESOLVE" [F0].  When Tor receives a "RESOLVE" SOCKS command, it initiates
  a remote lookup of the hostname provided as the target address in the SOCKS
  request.  The reply is either an error (if the address couldn't be
  resolved) or a success response.  In the case of success, the address is
  stored in the portion of the SOCKS response reserved for remote IP address.

PR #991, opened 2021-04-08
Port seed data from bitcoin core 

lontivero:
Wasabi wallet users using Tail OS and similars (sorry I don't remember the names) cannot connect to the bitcoin p2p network because NBitcoin is not able to discover peers using the DNS seeds given that those systems cannot have access to the clear net, only Tor connections are possible.

One of those cases was the trigger to create this issue: https://github.com/MetacoSA/NBitcoin/issues/977 however, recently other users have reported a the problem that they cannot use Wasabi: https://github.com/zkSNACKs/WalletWasabi/issues/5512

NBitcoin uses two mechanism to bootstrap the initial list of node: 
* DNS seeds and,
* Fixed addresses

This PR simply imports the list of fixed addresses from bitcoin core which includes onion v2 and v3 addresses that can be used to connect to those and get more addresses.

@nopara73 is going to work in resolving dns lookups though Tor in case the application is using socks (Tor)


nopara73 (approved):
LGTM

adamPetho:
The code LGTM, utACK

Szpoti (approved):
utACK, the code looks good

gruve-p:
Hi @NicolasDorier 
Do the altcoins also need to update the seed data since the pnSeeds array is now converted differently?
I can send a PR to replace https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin.Altcoins/Groestlcoin.cs#L197-L283 with https://github.com/Groestlcoin/groestlcoin/blob/master/src/chainparamsseeds.h#L9-L104 

Please advice

NicolasDorier:
I think altcoin does not need to update it if nobody uses it. Those seeds are used for a very specific case, which, as far as I know, only Wasabi ever needed.

PR #998, opened 2021-04-23
Set Chain Name for liquid

Kukks:


NicolasDorier:
Closing this, the breaking change in liquid was not caused by this but by a change in nbx

PR #1003, opened 2021-05-12
Make sure DNS requests from the addrman are passing through Socks if …

NicolasDorier:
…behavior is set up


NicolasDorier:
Should fix https://github.com/MetacoSA/NBitcoin/issues/977

NicolasDorier:
If you use `Node.Connect` nothing should change in your code.
If not, `AddressManager` should have the `DnsResolver` property set to the `DnsSocksResolver`. (This is done automatically in `Node.Connect`)

@lontivero @nopara73

NicolasDorier:
I made new package

PR #1009, opened 2021-05-27
Override Script Eval Context for Elements

Kukks:
Adds: CHECKSIGFROMSTACK, CHECKSIGFROMSTACKVERIFY
Re-enables: CAT

I haven't figured out a good way to plug this into the system painlessly though. I think I need to add `CreateScriptEvaluationContext` to `ConsensusFactory` but that requires I need to do method signature changes across the lib


NicolasDorier:
@Kukks do we need this btw? I am not sure it makes sense to validate the scripts with NBitcoin, even if we support it.
Probably better to rely on Bitcoin Core for this.

PR #1014, opened 2021-06-15
Fix the portuguese word list to use the correct and accepted version

johnnyasantoss:
Implements the correct Portuguese word list (see #385 & https://github.com/bitcoin/bips/pull/998)


NicolasDorier:
I also announced it on https://twitter.com/NicolasDorier/status/1406805465686675458?s=20

That suck... :(

johnnyasantoss:
@NicolasDorier I made a tool that makes the conversion of words from one list to another. Check it out https://github.com/johnnyasantoss/bip39-nbitcoin-fixer

NicolasDorier:
@johnnyasantoss  mmh maybe I should add a method as well to convert it.

johnnyasantoss:
> @johnnyasantoss mmh maybe I should add a method as well to convert it.

I don't see wallets doing this type of recovery, given that it is super specific. I think that people would already be able to find this PR with ease :)

PR #1016, opened 2021-06-24
Add taproot addresses

NicolasDorier:


NicolasDorier:
My main issue right now is that I don't know if we should create a new PubKey class or reuse the existing one...

NBitcoin/TaprootAddress.cs
```diff
@@ -0,0 +1,119 @@
...
+	public class TaprootAddress : BitcoinAddress, IBech32Data, IPubkeyHashUsable
+	{
+		public TaprootAddress(string bech32, Network expectedNetwork)
+				: base(Validate(bech32, expectedNetwork), expectedNetwork)
+		{
+			if (expectedNetwork.GetBech32Encoder(Bech32Type.TAPROOT_ADDRESS, false) is Bech32Encoder encoder)
+			{
+				var decoded = encoder.Decode(bech32, out _);
+				if (ECXOnlyPubKey.TryCreate(decoded, out var k))
+					_PubKey = new PubKey(k);
```
dangershony, line 22: Are you referring to this code with reusing the pukey?

NBitcoin/TaprootAddress.cs
```diff
@@ -0,0 +1,119 @@
...
+		{
+			_PubKey = pubKey;
+		}
+
+		private static string? NotNull(PubKey pubKey)
+		{
+			if (pubKey == null)
+				throw new ArgumentNullException(nameof(pubKey));
+			return null;
+		}
```
lontivero, line 76: wtf!?

NBitcoin/TaprootAddress.cs
lontivero: Yes. As you know a pubkey is just a point  (x, y) in the elliptic curve and given it is symmetric there are two point for each x coordinate (higher and lower). Before taproot we had that a public key was "compressed"  meaning that it was a 33 bytes pubkey (32 bytes corresponding to the X coordinate plus and additional byte to indicate which of the two points to use).

In Taproot we have a 32 bytes public key and the extras bit was just dropped, that's the `XOnlyPubKey`. Nicolas doubt makes sense to me because I don't know what the best approach is neither.

dangershony: Ah I see what you mean, intuitively I would say we should create a new pubkey type.

The options I see are:
-  new pubkey it can perhaps inherit from the pubkey and check the array limits to 32 bytes (then its reusable).
-  new type all together (no inheritance) this mean les compatibility with other components.
- dummy last byte in the array (but this is tricky as we cant find a dummy byte that is invalid).

Intuitively I think inheritance is the easiest option.

NBitcoin/TaprootAddress.cs
NicolasDorier: yes, this is to throw a proper exception. in case pubKey is null. I do this, because I use it in the ctor initializer, where I can't have code blocks.

That's a bit weird, but just reusing the code of BitcoinAddress.

NicolasDorier:
It's kind of complicated. Because when you are using the taproot script path you NEED the PubKey with both coordinate to properly sign it.

You don't need it for the key path spend though.
But on chain, you only get the 32 bytes.

NicolasDorier:
I am thinking of having a type `TaprootOutputPubKey` which represent this 32 bytes.
But this type will be more like `KeyId/ScriptId` than it looking like `PubKey`.

I would add `Key.SignTaproot()` and `Key.GetTaprootOutputPubKey()`.
Those two methods assume the user of taproot is doing single sig operation.

More complicated stuff (interpreter), will be left for later PRs

dangershony:
> More complicated stuff (interpreter), will be left for later PRs

Great, sounds reasonable.

NBitcoin/Network.cs
```diff
@@ -2562,7 +2566,8 @@ public IBitcoinString Parse(string str, Type? targetType)
 						candidate = new BitcoinWitPubKeyAddress(str.ToLowerInvariant(), bytes, this);
 					if (witVersion == 0 && bytes.Length == 32 && type == Bech32Type.WITNESS_SCRIPT_ADDRESS)
 						candidate = new BitcoinWitScriptAddress(str.ToLowerInvariant(), bytes, this);
-
+					if (witVersion == 1 && bytes.Length == 32)
```
prusnak, line 2569: Aren't we missing this?

```suggestion
					if (witVersion == 1 && bytes.Length == 32 && type == Bech32Type.TAPROOT_ADDRESS)
```

NicolasDorier: yes, I need a bunch of tests as well. I want to do this week.

NicolasDorier:
Ok this is ready. This is a really basic implementation only meant for software to be able to parse and send to Taproot addresses.

Note I support this in the net framework, but I do not perform any complicated crypto check whether the taproot pubkey is correct. This mean that legacy .net framework will still be able to parse addresses, and send to taproot, but that I will not support anything else related to taproot for such environment.

NicolasDorier:
Merging this for now. Will probably make a breaking change to NBitcoin. The `TxDestination` class is not really good.

PR #1019, opened 2021-06-29
[WIP] Taproot scripts validation

lontivero:
This is just a work-in-progress PR to get early feedback. I will try to make small commits to make a commit-by-commit review easier.

### DIFF
--- a/NBitcoin/Script.cs
+++ b/NBitcoin/Script.cs
@@ -118,6 +118,16 @@ namespace NBitcoin
 		WitnessPubkeyType = (1U << 15),
 
 		/// <summary>
+		/// Taproot/Tapscript validation (BIPs 341 & 342)
+		/// </summary>
+		Taproot = (1U << 17),
+
+		/// <summary>
+		/// Making unknown Taproot leaf versions non-standard
+		/// </summary>
+		DiscourageUpgradableTaprootVersion = (1U << 18),
+
+		/// <summary>
 		/// Some altcoins like BCash and BGold requires ForkId inside the sigHash
 		/// </summary>
 		ForkId = (1U << 29),
--- a/NBitcoin/Script.cs
+++ b/NBitcoin/Script.cs
@@ -337,7 +337,9 @@ namespace NBitcoin
 		Original = 0,
 		[Obsolete("Use HashVersion.WitnessV0 instead")]
 		Witness = 1,
-		WitnessV0 = 1
+		WitnessV0 = 1,
+		Taproot = 2,
+		Tapscript = 3
 	}
 
 	public enum ScriptType
--- a/NBitcoin/ScriptEvaluationContext.cs
+++ b/NBitcoin/ScriptEvaluationContext.cs
@@ -565,7 +565,7 @@ namespace NBitcoin
 		private bool VerifyWitnessProgram(WitScript witness, WitProgramParameters wit, TransactionChecker checker)
 		{
 			List<byte[]> stack = new List<byte[]>();
-			Script scriptPubKey;
+			Script execScript;
 
 			if (wit.Version == 0)
 			{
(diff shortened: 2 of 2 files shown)


dangershony:
Great job starting on this.
I assume you are porting form core, in that case perhaps posting the links to where in core the code is ported form will be useful? (it will make it easier to help with review)

lontivero:
Everything is in this PR: [Merge #19953: Implement BIP 340-342 validation (Schnorr/taproot/tapscript)](https://github.com/bitcoin/bitcoin/pull/19953).  

The changes contained in this PR up to now can be found in [Implement Tapscript script validation rules (BIP 342)](https://github.com/bitcoin/bitcoin/commit/72422ce396b8eba7b1a72c171c2f07dae691d1b5) I would have sound more logical to start with [Implement Taproot validation (BIP 341)](https://github.com/bitcoin/bitcoin/commit/8bbed4b7acf4c76eaea8c0e10f3cbf6ba4e53809) but I suspected that could overlap with what Nicolas is working on.
 
Anyway, my approach for this is:  I read the BIP to understand what has to be done then check Bitcoin Core (`interpreter.cpp` mostly) code to see how that translates to real-world code and whether it is small enough to be ported at the moment or not;  go back to read the BIP again.

lontivero:
Yes, that's true.

NicolasDorier:
You need to add the test vectors probably updating `script_tests.json` and `tx_valid.json`, `tx_invalid.json` is enough to test all the paths. (you can pick those on Bitcoin core repo)

onvej-sl:
I think [`ExtractScriptPubKeyParameters2`](https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/StandardScriptTemplate.cs#L1061-L1073) erroneously returns `Version = 0x51` (instead of `Version = 0x01`) for taproot script pubkeys.

This is because the first byte of a taproot script pubkey is `0x51` (rather than `0x01`), which is the opcode for `OP_1`. This is something that it's not written very clearly in BIP141:
> A scriptPubKey (or redeemScript as defined in BIP16/P2SH) that consists of a 1-byte push opcode (for 0 to 16) followed by a data push between 2 and 40 bytes gets a new special meaning. The value of the first push is called the "version byte". The following byte vector pushed is called the "witness program".

Nevertheless, it's obvious from how it is [implemented](https://github.com/bitcoin/bitcoin/blob/master/src/script/script.cpp#L220) in bitcoin core.

NicolasDorier:
@onvej-sl this pass.

```csharp
[Fact]
public void CanExtractTaprootScript()
{
	var a = BitcoinAddress.Create("bc1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vqzk5jj0", Network.Main);
	var res = PayToWitTemplate.Instance.ExtractScriptPubKeyParameters2(a.ScriptPubKey);
	Assert.Equal(OpcodeType.OP_1, res.Version);
}
```

onvej-sl:
> @onvej-sl this pass.

The test confirms what I say, since it asserts that the version of the script pub key is `OpcodeType.OP_1 = 0x51`.

The version byte is compared with `0x01` [here](https://github.com/MetacoSA/NBitcoin/blob/b88ff8b24417b5ba71513b72f465d2c050dfad07/NBitcoin/ScriptEvaluationContext.cs#L662). So either
  * `ExtractScriptPubKeyParameters2` must return `0x01` or
  * the version byte must be compared with `0x51`.

NicolasDorier:
I'll take over this.
All the taproot signing stuff for key spend is possible, now I just need the script evaluator.

NicolasDorier:
supersede by https://github.com/MetacoSA/NBitcoin/pull/1026

I did not ported everything yet, just key spend. Will do the rest in a later version of NBitcoin.

NicolasDorier:
This complete the work: https://github.com/MetacoSA/NBitcoin/pull/1037

PR #1020, opened 2021-07-04
Cancellable RPC methods

lontivero:
## Content

This PR adds a cancellation token to most async methods and pass it to the `HttpClient.SendAsync` method. Depending on the framework's version there are other methods that could also receive the cancellation token but that is out of the scope of this PR, however, the work done here can make that future work easier.

## Background

In Wasabi many times we propagate a cancellation token all the way down to all async methods and some times the last call in that chain is an async call to the `RPCClient` that do not accept a cancellation token.

I assume having an uniform pattern where most methods accept a cancellation token and can cancel a IO task is desirable (not always possible).

## A tiny huge problem 

For reasons of usability the `RPCClient::SendCommandAsync` methods receive optional parameters with `params object[]` what makes these methods really easy to use but that comes with the problem that it is not possible to add a `CancellationToken`  parameter at the end of the list (what is the recommended place to put it) because `params` must be the latest one.

There is no good solution to that problem and for that reason it is not easy to add `CancellationToken` to the methods without breaking compatibility so, I decided to DO NOT break it (except for that can be seen in one test). I had to create a new (private) method called `InternalSendCommandAsync` that is used internally instead of the already published `SendCommandAsync`.
The name `InternalSendCommandAsync` doesn't sound great, I know.

**Update:**

This work is needed for https://github.com/zkSNACKs/WalletWasabi/issues/5991

### DIFF
--- a/NBitcoin/RPC/RPCClient.cs
+++ b/NBitcoin/RPC/RPCClient.cs
@@ -828,7 +828,7 @@ namespace NBitcoin.RPC
 		/// <summary>
 		///	Send all commands in one batch
 		/// </summary>
-		public async Task SendBatchAsync()
+		public async Task SendBatchAsync(CancellationToken cancellationToken = default)
 		{
 			Tuple<RPCRequest, TaskCompletionSource<RPCResponse>> req;
 			List<Tuple<RPCRequest, TaskCompletionSource<RPCResponse>>> requests = new List<Tuple<RPCRequest, TaskCompletionSource<RPCResponse>>>();
@@ -842,7 +842,7 @@ namespace NBitcoin.RPC
 			}
 			if (requests.Count == 0)
 				return;
-			await SendBatchAsyncCore(requests).ConfigureAwait(false);
+			await SendBatchAsyncCore(requests, cancellationToken).ConfigureAwait(false);
 		}
 
 		/// <summary>
@@ -858,7 +858,7 @@ namespace NBitcoin.RPC
 		/// </summary>
 		public bool AllowBatchFallback { get; set; }
 
-		private async Task SendBatchAsyncCore(List<Tuple<RPCRequest, TaskCompletionSource<RPCResponse>>> requests)
+		private async Task SendBatchAsyncCore(List<Tuple<RPCRequest, TaskCompletionSource<RPCResponse>>> requests, CancellationToken cancellationToken)
 		{
 			var writer = new StringWriter();
 			writer.Write("[");
@@ -881,7 +881,7 @@ namespace NBitcoin.RPC
 			{
 			retry:
 				var webRequest = CreateWebRequest(writer.ToString());
-				using (var cts = new CancellationTokenSource(RequestTimeout))
+				using (var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, new CancellationTokenSource(RequestTimeout).Token))
 				{
 					using (var httpResponse = await HttpClient.SendAsync(webRequest, cts.Token).ConfigureAwait(false))
 					{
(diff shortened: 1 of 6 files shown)


NBitcoin/RPC/RPCClient.cs
```diff
@@ -574,7 +575,12 @@ public async Task<BitcoinAddress> GetRawChangeAddressAsync()
 
 		public Task<RPCResponse> SendCommandAsync(RPCOperations commandName, params object[] parameters)
 		{
-			return SendCommandAsync(commandName.ToString(), parameters);
+			return InternalSendCommandAsync(commandName, parameters, CancellationToken.None);
```
dangershony, line 578: What if you make a `CancellationToken ` type check on the last param (and default to None otherwise) then remove it form the params.
Ugly hack I agree

lontivero: Well, I thought about that too before jumping to code and the problem that I saw is that it is a hack (it is not so bad imo) and the IDE cannot provide you the autocomplete help. Anyway, That could have made the changeset a little less aggressive. Let see what @NicolasDorier says, I am not opposed to that change and can do it.

augustoproiete: :thinking: seems a little obscure for a `public` API, and not very easy to document. Perhaps adding a method overload that includes the `CancellationToken` before the `params` would be good compromise, and also seems to the accepted pattern Microsoft uses in their own codebase (e.g. [1](https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.dbset.findasync?view=entity-framework-6.2.0#System_Data_Entity_DbSet_FindAsync_System_Threading_CancellationToken_System_Object___), [2](https://github.com/dotnet/aspnetcore/blob/v5.0.7/src/JSInterop/Microsoft.JSInterop/src/JSObjectReferenceExtensions.cs#L66))

dangershony: @augustoproiete proposal seems good, it's not a big deal to have an overload and inside call `InternalSendCommandAsync` and no ugly hack.

augustoproiete: I'm thinking if you go the overload route, you wouldn't need `InternalSendCommandAsync` anymore as the compiler would continue to bind calls to the old method unless a `CancellationToken` is passed through as a second argument keeping the API & ABI backwards compatible.

```diff
public Task<RPCResponse> SendCommandAsync(string commandName, params object[] parameters)
{
-    return InternalSendCommandAsync(commandName, parameters, CancellationToken.None);
+    return SendCommandAsync(commandName, CancellationToken.None, parameters);
 }

-private Task<RPCResponse> InternalSendCommandAsync(string commandName, object[] parameters, CancellationToken cancellationToken)
+public Task<RPCResponse> SendCommandAsync(string commandName, CancellationToken cancellationToken, params object[] parameters)
{
    return SendCommandAsync(new RPCRequest(commandName, parameters), cancellationToken: cancellationToken);
}
```

lontivero: done

NicolasDorier:
Adding optional parameter is actually a breaking change if the version of NBitcoin at runtime is different from the one at build time. (this can happen for various reason when there is transitory dependency references, it already happened to me and users of NBitcoin)

I would say, if you really need it, what about adding a `CancellationToken` to the RPCClient class and use this one in the HttpClient. That's not really clean, I have to admit... or maybe should we bite the bullet and make the breaking changes.

For `SendCommandAsync` you can just add another overload where the CancellationToken is not the last parameter.

lontivero:
Yep.

NicolasDorier:
Reopening this, due to taproot I am doing some breaking changes accross NBitcoin, so let's do this as well.

PR #1021, opened 2021-07-09
Fix bounds for ContextStack.Top() and ContextStack.Swap()

onvej-sl:
Furthermore, the comments such as `Returns the i-th element from the top of the stack.` are quite misleading since `i` is negative.

### DIFF
--- a/NBitcoin/ScriptEvaluationContext.cs
+++ b/NBitcoin/ScriptEvaluationContext.cs
@@ -2132,7 +2132,7 @@ namespace NBitcoin
 		/// <exception cref="System.IndexOutOfRangeException">topIndex</exception>
 		public T Top(int i)
 		{
-			if (i > 0 || -i > Count)
+			if (i >= 0 || -i > Count)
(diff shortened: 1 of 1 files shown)


NBitcoin/ScriptEvaluationContext.cs
```diff
@@ -2147,9 +2147,9 @@ public T Top(int i)
 		/// </exception>
 		public void Swap(int i, int j)
 		{
-			if (i > 0 || -i > Count)
+			if (i >= 0 || -i > Count)
 				throw new IndexOutOfRangeException("i");
-			if (i > 0 || -j > Count)
+			if (i >= 0 || -j > Count)
```
lontivero, line 2152: It should check j instead of i

NicolasDorier:
Nice one, can you fix  https://github.com/MetacoSA/NBitcoin/pull/1021/files#r667120029 ?
Damn that's old code... I wonder why is the context stack even a public class though... it is only useful for the script interpreter.

PR #1022, opened 2021-07-12
Fix copy index in TaprootPubKey.ScriptPubKey

onvej-sl:


NicolasDorier:
wow why isn't it caught by tests. Let me check.

NicolasDorier:
ok because I forgot to add the trait https://github.com/MetacoSA/NBitcoin/commit/5836dc97a4deceedbf2d3a8908a27f1d44dd71f5

PR #1026, opened 2021-07-19
Can Validate key spend

NicolasDorier:
Ability to validate key spend via the script evaluator.

### DIFF
--- a/NBitcoin/ScriptEvaluationContext.cs
+++ b/NBitcoin/ScriptEvaluationContext.cs
@@ -463,7 +463,6 @@ namespace NBitcoin
 		public ScriptEvaluationContext()
 		{
 			ScriptVerify = NBitcoin.ScriptVerify.Standard;
-			SigHash = NBitcoin.SigHash.Undefined;
 			Error = ScriptError.UnknownError;
 		}
 		public ScriptVerify ScriptVerify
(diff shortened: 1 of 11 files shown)


NBitcoin/ScriptEvaluationContext.cs
```diff
@@ -559,47 +601,76 @@ public bool VerifyScript(Script scriptSig, Script scriptPubKey, TransactionCheck
...
 					scriptPubKey = PayToPubkeyHashTemplate.Instance.GenerateScriptPubKey(new KeyId(wit.Program));
-					stack = witness.Pushes.ToList();
+					return ExecuteWitnessScript(stack, scriptPubKey, HashVersion.WitnessV0, checker);
 				}
 				else
 				{
 					return SetError(ScriptError.WitnessProgramWrongLength);
 				}
 			}
+			else if (wit.Version == OpcodeType.OP_1 && wit.Program.Length == 32 && !isP2SH)
```
lontivero, line 641: I think this was discussed somewhere else but the `wit.Version` should be `1` instead of `OP_1`

NicolasDorier: On the contrary, it should be OP_1 but not 1

PR #1030, opened 2021-07-20
Make SigningOptions a property rather than passing explicitely to methods

NicolasDorier:
This allow us to get rid of the method overload explosion, and prevent foot shooting.


NBitcoin.TestFramework/WellknownNodeDownloadData.cs
```diff
@@ -1,12 +1,37 @@
...
+using System.Linq;
 
 namespace NBitcoin.Tests
 {
 	public partial class NodeDownloadData
 	{
 		public bool SupportCookieFile { get; set; } = true;
 
-		public class BitcoinNodeDownloadData
+		public class NodeDownloadDataBase
```
knocte, line 10: this seems unrelated to the PR, pushed by mistake?

NicolasDorier: Unrelated to this PR, I did a separate commit.

knocte: but the NodeDownload commit had signing-related things too xD

PR #1031, opened 2021-07-20
Added support for Triptourcoin

macloviw:


NicolasDorier:
You did not add a commented line for your alt in `NodeBuilderEx`.
You also need to add your binaries in `WellknownNodeDownloadData.cs`, and run the Altcoin tests yourself by changing the line in NodeBuilderEx which use Core by default.

NicolasDorier:
need rebase

PR #1042, opened 2021-08-02
Add getblockstats to RPCClient

### LINKED ISSUE #1029: Missing RPC getblockstats
Any plans to add this to the RPC client?

See https://bitcoincore.org/en/doc/0.17.0/rpc/blockchain/getblockstats/

bezysoftware:
Resolves #1029

### DIFF
--- a/NBitcoin.Tests/RPCClientTests.cs
+++ b/NBitcoin.Tests/RPCClientTests.cs
@@ -505,6 +505,24 @@ namespace NBitcoin.Tests
 		}
 
 		[Fact]
+		public void CanGetStatsFromRPC()
+		{
+			using (var builder = NodeBuilderEx.Create())
+			{
+				var rpc = builder.CreateNode().CreateRPCClient();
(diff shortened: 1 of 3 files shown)


NicolasDorier:
Thanks, can you add one test?

bezysoftware:
> 
> 
> Thanks, can you add one test? This help to spot when code break because of breaking change in new release of nodes.

Can you help me with the setup of the test? When I try to call the method I get "Can't read undo data from disk" error. I tried running "bitcoin-qt" as described in the `RPCClientTests.cs` file but I get the same error.

NicolasDorier:
@bezysoftware just look at all the RPC tests that already exists in NBitcoin and do the same (example `CanGetBlockFromRPC`)

bezysoftware:
@NicolasDorier turns out I cannot run the `getblockstats` on the genesis block but must generate at least one extra, the above error was coming as a result of the RPC call

PR #1052, opened 2021-08-17
Support Elements DynFed

Kukks:


NBitcoin/BitcoinStream.cs
```diff
@@ -18,9 +18,9 @@ namespace NBitcoin
 {
 	public enum SerializationType
 	{
-		Disk,
```
NicolasDorier, line 21: I will remove this, this is breaking change

Issue #1050, opened 2021-08-13
Exception on Ubuntu 20.04


xclud:
I run the following code from the blockchain book:

```
var privateKey = new Key(); // generate a random private key
var publicKey = privateKey.PubKey;

return publicKey.GetAddress(ScriptPubKeyType.Legacy, Network.Main).ToString();
```
And i get the following exception:
```
at NBitcoin.Bitcoin.InitSignet() in NBitcoin.dll:token 0x60002fa+0x0
at NBitcoin.Network..cctor() in NBitcoin.dll:token 0x60006be+0x108
--- End of inner exception stack trace ---
at NBitcoin.Network.get_Main() in NBitcoin.dll:token 0x60006bf+0x0
```
I tried `5.x.x` and `6.x.x` versions of the library and i get the same results.

NicolasDorier:
@xclud can you update and try again? I fixed this bug normally.

xclud:
Nop. This is not working.

I installed the version `6.0.9`. My `Main()` has only one line:

```
void Main()
{
  NBitcoin.Network network = NBitcoin.Network.Main;
}
```

and I get this exception:
```
---> System.ArgumentNullException: Value cannot be null. (Param>
  at System.IO.Path.Combine(String path1, String path2, String >
  at NBitcoin.Bitcoin.CreateSignet()
  at NBitcoin.Bitcoin.InitSignet()
  at NBitcoin.Network..cctor()
  --- End of inner exception stack trace ---
  at NBitcoin.Network.get_Main()
```

I have this project as a daemon (background service) in Ubuntu. My guess is this project is trying to get the user's `Home` directory and do `System.IO.Path.Combine`. Home directory does not mean anything in a background service as we are not logging in with a user account. So the value `null` is passed to the `System.IO.Path.Combine`.

BTW, what I dislike about NBitcoin is it relates to the filesystem and network and even thought I only want to generate an offline wallet, it still initializes some directories and network connections without developer's consent.

xclud:
Here is the source of the exception:

https://github.com/MetacoSA/NBitcoin/blob/17ce930b9029ceeed3d0a9d8e60fda2f8631f86e/NBitcoin/Network.cs#L2075

augustoproiete:
[...]

PR #1063, opened 2021-09-27
Update Network GetDefaultDataFolder to fallback to OS Temp folder when HOME and APPDATA not available

augustoproiete:
Fallback to the operating system's TEMP folder when both `HOME` and `APPDATA` are not available, as a workaround.

A better approach for a future improvement would be to remove any filesystem dependencies when accessing networks.

---

Repro steps: https://github.com/augustoproiete-repros/nbitcoin-issue-1050-ubuntu-exception

`Network.GetDefaultDataFolder` expects either `HOME` or `APPDATA` environment variables to be available, which might not be the case in some scenarios as described in https://github.com/MetacoSA/NBitcoin/issues/1050 [in which case it returns `null`](https://github.com/MetacoSA/NBitcoin/blob/v6.0.9/NBitcoin/Network.cs#L2077-L2078).

`CreateSignet` [relies](https://github.com/MetacoSA/NBitcoin/blob/v6.0.9/NBitcoin/Bitcoin.cs#L64-L65) on `Network.GetDefaultDataFolder` to set the path of the signet cookie and currently [throws an exception when the value is `null`](https://github.com/augustoproiete-repros/nbitcoin-issue-1050-ubuntu-exception/runs/3715694401?check_suite_focus=true#step:6:22).

[image](https://github.com/augustoproiete-repros/nbitcoin-issue-1050-ubuntu-exception/runs/3715694401?check_suite_focus=true#step:6:22)

---

Closes https://github.com/MetacoSA/NBitcoin/issues/1050

### DIFF
--- a/NBitcoin/Network.cs
+++ b/NBitcoin/Network.cs
@@ -2073,7 +2073,7 @@ namespace NBitcoin
 		public static string? GetDefaultDataFolder(string folderName)
 		{
 			var home = Environment.GetEnvironmentVariable("HOME");
-			var localAppData = Environment.GetEnvironmentVariable("APPDATA");
+			var localAppData = Environment.GetEnvironmentVariable("APPDATA") ?? Path.GetTempPath();
 			if (string.IsNullOrEmpty(home) && string.IsNullOrEmpty(localAppData))
 				return null;
 			if (!string.IsNullOrEmpty(home) && string.IsNullOrEmpty(localAppData))


knocte:
But reverting to the tmp folder might let the user think that their data is safe while it's not.
IMHO the better way to solve this is an optional DirectoryInfo parameter in InitSignet() maybe?

knocte:
>  that their data is safe while it's not.

Mmm or maybe signet related data is no big deal to lose because it's not big anyway?

augustoproiete:
> an optional DirectoryInfo parameter in InitSignet() maybe?

The issue is that the code runs in a static context during initialization, before the user had the opportunity to send any parameters ([it's a private constructor](https://github.com/MetacoSA/NBitcoin/blob/v6.0.9/NBitcoin/Bitcoin.cs#L17)).

Perhaps a better approach would be to introduce an "Options" class with some defaults, which the user can override?

Something along those lines:

```csharp
var myBitcoinOptions = new BitcoinOptions
{
    DataFolderPath = "./myapp/bitcoin-data",
};

Bitcoin.DefaultOptions = () => myBitcoinOptions;

// Static context relies on Bitcoin.DefaultOptions().DataFolderPath
Network network = Network.Main;

// ---

public class BitcoinOptions
{
    public string DataFolderPath { get; set; }
}

public class Bitcoin
{
    public static Func<BitcoinOptions> DefaultOptions { get; set; } = () => new BitcoinOptions();
}
```

We could then update the code to rely on the `DataFolderPath` of these options, and throw an exception if we couldn't detect the HOME/APPDATA path, so that the user knows they have to set it.

knocte:
Ughh so ugly lol, I prefer the GetTempPath() approach better now 😅

NicolasDorier:
I did this e4a0b4baa10b4f4c97523aa321d59c7653679f30 instead.

As I was checking how `GetDefaultDataFolder` is currently used, I noticed other clients of this method, handle the case when it's null. So I followed this.

PR #1064, opened 2021-09-29
Added LitecoinUrlBuilder

su8898:
Litecoin also supports BIP0021

### DIFF
--- a/NBitcoin.Altcoins/Litecoin.cs
+++ b/NBitcoin.Altcoins/Litecoin.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Rather than making new type, just add a new property in the `Network` class `UrlScheme` which default to bitcoin, and make the BitcoinUrlBuilder use this instead of hardcoded `bitcoin`.

knocte:
> Rather than making new type, just add a new property in the Network class UrlScheme which default to bitcoin, and make the BitcoinUrlBuilder use this instead of hardcoded bitcoin.

Mmm, but there is a constructor (the empty one) that doesn't receive a Network object. Should we have a default Network object in this case?

Richlilyjoy100 (approved):
Dc

PR #1069, opened 2021-11-09
Support net6.0 for NBitcoin

NicolasDorier:
This also remove support to .NET 2.1 and net452.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -14,11 +14,26 @@ env:
   DOTNET_CLI_TELEMETRY_OPTOUT: 'true'
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
+  net60:
+    name: '6.0'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '6.0.x'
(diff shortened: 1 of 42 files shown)


.github/workflows/master.yml
```diff
@@ -27,35 +42,35 @@ jobs:
...
-          dotnet-version: '3.1.x'
+          dotnet-version: '6.0.x'
       - run: ./Build/CI/tests.sh
         env:
           BUILD_ARGS: /p:TargetFrameworkOverride=netstandard2.0
-          Framework: netcoreapp3.1
+          Framework: net6.0
   dotnetcore31macod:
-    name: '3.1 on Mac-OS'
+    name: '6.0 on Mac-OS'
```
knocte, line 57: @NicolasDorier pls note with this change, name of the job doesn't match anymore with the job's caption, should be dotnet6macos

NicolasDorier: fixed

fyziktom:
Hello. 
I think the update to .NET 6.0 can be little issue. It will probably not work in the Blazor WebAssembly. It has some trouble with the  MS Cryptorgraphy library. [here is more info](https://docs.microsoft.com/en-us/dotnet/core/compatibility/cryptography/5.0/cryptography-apis-not-supported-on-blazor-webassembly)

PR #1070, opened 2021-11-09
Fix typo in job's caption

yahiheb:


knocte:
> System.InvalidOperationException : The node is not in a connected state

Hey @NicolasDorier can we make that test retry 2 or 3 times? it's always making noise.

PR #1082, opened 2022-01-24
Enhance PayToTaprootTemplate

onvej-sl:
This PR makes some enhancement in `PayToTaprootTemplate`:
  * It deprecates ~~`GenerateScriptPubKey`~~`GenerateScriptSig` in favour of `GenerateWitScript`. Rationale: A taproot output is a version one segwit output (see [BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#script-validation-rules)). That means its `scriptSig` is empty and its redeem script (Schnorr signature) is placed in `witness`. The function `GenerateScriptPubKey` remains present in the code to ensure backward compatibility. 
  * It implements support of annex (see [BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_ref-5-0)).
  * It implements `ExtractWitScriptParameters`.
  * It introduces tests of `PayToTaprootTemplate`.

### DIFF
--- a/NBitcoin/StandardScriptTemplate.cs
+++ b/NBitcoin/StandardScriptTemplate.cs
@@ -194,7 +194,7 @@ namespace NBitcoin
 			if (signature == null)
 				throw new ArgumentNullException(nameof(signature));
 			if (annex == null)
-				throw new ArgumentNullException(nameof(annex));
+				return GenerateWitScript(signature);
 			if (!CheckAnnex(annex))
 				throw new ArgumentException("The first byte of annex must be 0x50", "annex");
 
--- a/NBitcoin/StandardScriptTemplate.cs
+++ b/NBitcoin/StandardScriptTemplate.cs
@@ -148,7 +148,7 @@ namespace NBitcoin
 			throw new NotSupportedException();
 		}
 
-		[Obsolete("Use GenerateWitScript instead")]
+		[Obsolete("Use GenerateWitScript(TaprootSignature signature) instead")]
 		public Script GenerateScriptSig(TaprootSignature signature)
 		{
 			if (signature == null)
@@ -158,6 +158,9 @@ namespace NBitcoin
 
 		public PayToTaprootScriptSigParameters ExtractWitScriptParameters(WitScript witScript)
 		{
+			if (witScript == null)
+				throw new ArgumentNullException(nameof(witScript));
+
 			if (witScript.PushCount != 1 && witScript.PushCount != 2)
 				return null;
 			var hasAnex = false;
@@ -191,7 +194,7 @@ namespace NBitcoin
 			if (signature == null)
 				throw new ArgumentNullException(nameof(signature));
 			if (annex == null)
-				throw new ArgumentNullException(nameof(signature));
+				throw new ArgumentNullException(nameof(annex));
 			if (!CheckAnnex(annex))
 				throw new ArgumentException("The first byte of annex must be 0x50", "annex");
 
--- a/NBitcoin/StandardScriptTemplate.cs
+++ b/NBitcoin/StandardScriptTemplate.cs
@@ -151,9 +151,7 @@ namespace NBitcoin
 		[Obsolete("Use GenerateWitScript(TaprootSignature signature) instead")]
 		public Script GenerateScriptSig(TaprootSignature signature)
 		{
-			if (signature == null)
-				throw new ArgumentNullException(nameof(signature));
-			return new Script(Op.GetPushOp(signature.ToBytes()));
+			return GenerateWitScript(signature);
 		}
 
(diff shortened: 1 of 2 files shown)


yahiheb:
> * It depracetes `GenerateScriptPubKey` in favour of `GenerateWitScript`.

It depracetes `GenerateScriptSig`.

NicolasDorier:
Awesome, will merge that tomorrow after reviewing.

NBitcoin/StandardScriptTemplate.cs
```diff
@@ -148,12 +148,55 @@ protected override bool CheckScriptSigCore(Script scriptSig, Op[] scriptSigOps,
...
+		{
+			return GenerateScriptSig(signature);
+		}
+
+		public WitScript GenerateWitScript(TaprootSignature signature, byte[] annex)
+		{
+			if (signature == null)
+				throw new ArgumentNullException(nameof(signature));
+			if (annex == null)
+				throw new ArgumentNullException(nameof(signature));
```
NicolasDorier, line 194: It should be `throw new ArgumentNullException(nameof(annex));`

But more importantly, I think annex should be able to be null. I am familiar in Bitcoin but I don't even remember what should be passed as annex.

NBitcoin/StandardScriptTemplate.cs
```diff
@@ -148,12 +148,55 @@ protected override bool CheckScriptSigCore(Script scriptSig, Op[] scriptSigOps,
 			throw new NotSupportedException();
 		}
 
+		[Obsolete("Use GenerateWitScript instead")]
 		public Script GenerateScriptSig(TaprootSignature signature)
 		{
 			if (signature == null)
 				throw new ArgumentNullException(nameof(signature));
 			return new Script(Op.GetPushOp(signature.ToBytes()));
```
NicolasDorier, line 156: this code should move in `GenerateWitScript`

NBitcoin/StandardScriptTemplate.cs
```diff
@@ -148,12 +148,55 @@ protected override bool CheckScriptSigCore(Script scriptSig, Op[] scriptSigOps,
...
+		[Obsolete("Use GenerateWitScript instead")]
 		public Script GenerateScriptSig(TaprootSignature signature)
 		{
 			if (signature == null)
 				throw new ArgumentNullException(nameof(signature));
 			return new Script(Op.GetPushOp(signature.ToBytes()));
 		}
+
+		public PayToTaprootScriptSigParameters ExtractWitScriptParameters(WitScript witScript)
+		{
```
NicolasDorier, line 160: need to check witScript isn't null

NBitcoin/StandardScriptTemplate.cs
onvej-sl: > It should be throw new ArgumentNullException(nameof(annex));

Fixed in https://github.com/MetacoSA/NBitcoin/pull/1082/commits/a24e96ca435db10e225e8eee9e2584ca7fb9ece9.

> I am familiar in Bitcoin but I don't even remember what should be passed as annex.

The annex is reserved for future use and is not used yet. See [this](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_note-5).

> But more importantly, I think annex should be able to be null.

This is a valid point. My intention was the following: If you want to generate the witness without an annex, you should call `GenerateWitScript(TaprootSignature signature)` instead. But I agree it would be more comfortable if `GenerateWitScript(TaprootSignature signature, byte[] annex)` was able to handle `NULL` annex itself, so I fixed it in https://github.com/MetacoSA/NBitcoin/pull/1082/commits/d00e4ec06e5b2a3c836dd7e3d5196513ffbda56a.

NBitcoin/StandardScriptTemplate.cs
onvej-sl: I think it's better to call `GenerateWitScript(TaprootSignature signature)`, fixed in https://github.com/MetacoSA/NBitcoin/pull/1082/commits/a24e96ca435db10e225e8eee9e2584ca7fb9ece9.

NBitcoin/StandardScriptTemplate.cs
onvej-sl: Fixed in https://github.com/MetacoSA/NBitcoin/pull/1082/commits/a24e96ca435db10e225e8eee9e2584ca7fb9ece9.

onvej-sl:
In https://github.com/MetacoSA/NBitcoin/pull/1082/commits/ca008dae959b34bcb791459c77f173c863a076da I swap bodies of the functions `GenerateScriptSig(TaprootSignature signature)` and `GenerateWitScript(TaprootSignature signature)`, so the obsolete function  `GenerateScriptSig(TaprootSignature signature)` can be easily removed once you decide to do that.

PR #1085, opened 2022-02-14
Use .NET `Socket.ConnectAsync(endpoint, cancellationToken)` library i…

kiminuo:
…mplementation in .NET 6+.

See https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.connectasync?view=net-6.0#system-net-sockets-socket-connectasync(system-net-endpoint-system-threading-cancellationtoken)


NicolasDorier:
Thanks for finding that, I pushed a different way more consistent with how we handle build in this project https://github.com/MetacoSA/NBitcoin/commit/3f5420eeaea443eeba6dc44627bac1165678b3d4

PR #1089, opened 2022-03-07
Await send message operations to avoid unobserved exceptions

kiminuo:
This PR attempts to fix some places where unobserved exceptions can be thrown.

note: This PR does not specifically fix https://github.com/zkSNACKs/WalletWasabi/issues/2554 but it is somewhat related to the issue.

note: This PR is not complete in the sense that it does not fix all `SendMessageAsync` calls. The PR is meant to be a preview really. Anyway, I believe that awaiting https://github.com/MetacoSA/NBitcoin/blob/cb6d64664b9f38d0442a4a4ad5edaed9e310bf40/NBitcoin/Protocol/Node.cs#L1085 would lead to significant decrease of unobserved exceptions in Wasabi.

### DIFF
--- a/NBitcoin/Protocol/Node.cs
+++ b/NBitcoin/Protocol/Node.cs
@@ -201,8 +201,8 @@ namespace NBitcoin.Protocol
 					{
 						Utils.SafeSet(ar);
 					};
-					try
 
+					try
 					{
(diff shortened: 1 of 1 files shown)


Szpoti (approved):
LGTM

MaxHillebrand:
what's up with this?

PR #1090, opened 2022-03-20
Remove dead code in Network.cs

yahiheb:
The PR just removes some code that is no longer used, and fixes a typo.

### DIFF
--- a/NBitcoin/Network.cs
+++ b/NBitcoin/Network.cs
@@ -2787,17 +2787,6 @@ namespace NBitcoin
 		{
 			return new BitcoinSecret(key, this);
 		}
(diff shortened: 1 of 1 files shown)


NBitcoin/Network.cs
```diff
@@ -2787,17 +2750,6 @@ public BitcoinSecret CreateBitcoinSecret(Key key)
 		{
 			return new BitcoinSecret(key, this);
 		}
-		public BitcoinPubKeyAddress CreateBitcoinAddress(KeyId dest)
```
kiminuo, line 2790: Not sure about this. It's a public method. How do you know nobody is using it?

Kukks (approved):

PR #1095, opened 2022-04-25
Update RPCClient.Wallet.cs

alphons:
bool translating to strings and feeding them as json to the rpc server never works!
dont stringify , bools are bools, now it works ;-)


NicolasDorier:
damn, should have tested that

NicolasDorier:
pushed 7.0.7

PR #1106, opened 2022-06-16
Use BC for hmac calculation on the extended key class

dangershony:
It seem the extended key type was missing the `NO_NATIVE_HMACSHA512`


NicolasDorier:
hey @dangershony sorry just saw this PR, can I know why you need this? none of our supported platform seems to fall into this case.

dangershony:
I was making a bIp39 tool in blazor wasm and I needed this change because some of the crypto apis are not yet supported with blazor (they might be in dotnet7)
This is the link to the tool, its not very urgent but can be cool to also support bip39 in c# wasm
https://github.com/block-core/blockcore-tools/tree/master/src/Blockcore.Web.Bip39#bip39-tool-using-c-wasm-blazor

PR #1121, opened 2022-08-08
Add ecash

mishfit:


NicolasDorier:
Altcoins integration need to be done in a single file I can easily review like the billion others that have been integrated.
I can't accept a PR that modify code outside of it.

mishfit:
@NicolasDorier this would allow all Altcoins to work like plugins instead of requiring PRs to implement them.

PR #1139, opened 2022-10-20
Resolve "Support for feefilter message" (see issue #1135)

### LINKED ISSUE #1135: Support for `feefilter` message
I would like to use (de)serialization of NBitcoin library for `feefilter` message. However I found out, that it is not present in supported messages (no `FeeFilterPayload` type) in NBitcoin. **So I suggest to add support for `feefilter` message.**

[FeeFilter](https://developer.bitcoin.org/reference/p2p_networking.html#feefilter) message was added in Bitcoin Core 0.13 (protocol 70013) and is described in [BIP133](https://github.com/bitcoin/bips/blob/master/bip-0133.mediawiki).

Quote from BIP133 specification:
> 2. Upon receipt of a "feefilter" message, the node will be permitted, but not required, to filter transaction invs for transactions that fall below the feerate provided in the feefilter message interpreted as satoshis per kilobyte.

So from point of view of implementation, it seems pretty straightforward - we do not need to implement _real filtering_ of TX to send.

I can make PR, if you are also interested in this feature.

vlada-cz:
Closes #1135

### DIFF
--- a/NBitcoin.Tests/AltcoinTests.cs
+++ b/NBitcoin.Tests/AltcoinTests.cs
@@ -112,6 +112,19 @@ namespace NBitcoin.Tests
 				new ConcurrentChain(builder.Network);
 			}
 		}
+		[Theory]
+		[Trait("UnitTest", "UnitTest")]
+		[InlineData(1234)]
+		[InlineData(0.9)]
+		public void CanSerializeDeserializeFeeFilter(decimal satPerBytes)
+		{
+			var p = new FeeFilterPayload();
+			p.FeeRate = new FeeRate(satPerBytes);
+			var bytes = p.ToBytes();
+			p = new FeeFilterPayload();
+			p.ReadWrite(new BitcoinStream(bytes));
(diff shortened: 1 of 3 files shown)


vlada-cz:
@NicolasDorier [Question] - Can it be like this? 
- Or do you prefer also some tests? 
- If so where I can put them ([`NBitcoin.Tests/ProtocolTests`](https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin.Tests/ProtocolTests.cs)) and what is your idea to test.

NicolasDorier:
It isn't easy to test, as Bitcoin core doesn't easily issue this message I guess?

NicolasDorier:
* Fixed a bug, you were affecting the long as a sat/b, but returning it as sat/kb.
* Added a small deserialization reserialization test
* Added nullable
* Use FeeRate rather than long in the private field.

NicolasDorier:
Pushed on 7.0.15

vlada-cz:
Thank you @NicolasDorier!

PR #1141, opened 2022-10-31
support taproot OutputDescriptor

joemphilips:
Not ready for review yet.


joemphilips:
Ready for review.
Two things to note.

1. I'm planning to add some tests. Since some basic tests for descriptor with tapscript are missing in bitcoind side, I should probably create a PR to bitcoind first.
2. it does not support signing/spending from taproot. Since TransactionBuilder does not support tapscript yet.

NicolasDorier:
I only saw that now, will look closer to this.
I don't really like the new members in `ScriptPubKeyType`. Those values should represent how an address can be derived directly from a pubkey.
The new values don't have enough information to derive from pubkey.

lontivero:
Is there something that I can do to help making this PR into NBitcoin?

lontivero:
image

NicolasDorier:
@lontivero you didn't pinged me enough :p

PR #1151, opened 2022-12-05
Remove "3 pubkeys in top level multisig" limit

dennisreimann:
This was introduced in #696 and it seems like an arbitrary limit. 

@joemphilips What is the reason for this?

### DIFF
--- a/NBitcoin/Scripting/OutputDescriptorParser.cs
+++ b/NBitcoin/Scripting/OutputDescriptorParser.cs
@@ -281,11 +281,6 @@ namespace NBitcoin.Scripting
 				return false;
 			}
 			result = res.Value;
-			if (result is OutputDescriptor.Multi multi && multi.PkProviders.Count > 3)
-			{
-				whyFailure = "You can not have more than 3 pubkeys in top level multisig.";
-				return false;
-			}
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


joemphilips:
This limit is taken from original implementation in bitcoind
https://github.com/bitcoin/bitcoin/blob/71abee86dba8f311775104c518644c55c3e123cf/src/test/descriptor_tests.cpp#L440

Why you want to use bare multisig? IIUC it is deprecated and you should use `wsh(multi())` or `sh(multi())` or `sh(wsh(multi()))`

dennisreimann:
This isn't a bare multisig, I'm coming across it in a `wsh(sortedmulti(2,[4 PUBKEYS]))` scenario when importing a wallet file containing that output descriptor in BTCPay Server.

But now that you hinted at that, I think the error isn't in the implementation here, but in our `DerivationSchemeParser`, which [uses `OutputDescriptor.Parse` recursively](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/DerivationSchemeParser.cs#L85), @NicolasDorier.

dennisreimann:
@joemphilips I was able to fix this issue on our end — thanks for the pointer, you are correct!

PR #1156, opened 2022-12-26
[Trivial] Update license

yahiheb:


knocte:
This change is done in 2022 so it's technically a lie.

NicolasDorier:
Now the code is correct!

Issue #138, opened 2016-01-16
BIP-0021 Payment URI builder


Avalaxy:
As far as I'm aware there's currently no builder that allows us to generate BIP-0021 payment URIs (please point it out if I'm wrong, I haven't found it though). Is there any interest for this? I'd like to fork the project and submit a PaymentUriBuilder as a pull request.

NicolasDorier:
There is already : BitcoinUrlBuilder :)

Avalaxy:
@NicolasDorier awesome, didn't know about it. It's not in the documentation. Saves me some work then! Open ended question: are there any specific things you'd like people to contribute to NBitcoin?

NicolasDorier:
Some documentation in my book : https://blockchainprogramming.azurewebsites.net/ but indeed, payment url is not mentioned.
Nothing in particular, but if you think something is missing I'm open to suggestions. Right now I'm working on segwit implementation.

Avalaxy:
Cool, thank you so much for your work.

kristapsk:
> There is already : BitcoinUrlBuilder :)

That is wrong naming, as it is [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) not [URL](https://en.wikipedia.org/wiki/URL), but guess that cannot be changed anymore due to backwards compatibility issues...

knocte:
> guess cannot be changed anymore due to backwards compatibility issues...

It could be changed, so long as you leave a BitcoinUrlBuilder type that inherits from the (renamed)BitcoinUriBuilder and has an `[Obsolete("Use BitcoinUriBuilder instead")]`. Want to propose a PR?

PR #1168, opened 2023-04-12
Rename BitcoinUrlBuilder to BitcoinUriBuilder

kristapsk:
BIP21 contains [URIs](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier), not [URLs](https://en.wikipedia.org/wiki/URL), this is correct terminology (every URL is URI, but not every URI is URL).

Original discussion - https://github.com/MetacoSA/NBitcoin/issues/138#issuecomment-1503402357.

### DIFF
--- a/NBitcoin/Payment/BitcoinUrlBuilder.cs
+++ b/NBitcoin/Payment/BitcoinUrlBuilder.cs
@@ -1,191 +1,33 @@
 #nullable enable
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Net;
-#if !NOHTTPCLIENT
-using System.Net.Http;
-using System.Net.Http.Headers;
-#endif
-using System.Text;
-using System.Threading.Tasks;
-using System.Web;
-using System.Web.NBitcoin;
-using System.Runtime.ExceptionServices;
 
 namespace NBitcoin.Payment
 {
-	/// <summary>
-	/// https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki
-	/// </summary>
-	public class BitcoinUrlBuilder
+	[Obsolete("Use BitcoinUriBuilder instead")]
+	public class BitcoinUrlBuilder : BitcoinUriBuilder
 	{
-		[Obsolete("Use BitcoinUrlBuilder(Network) instead")]
-		public BitcoinUrlBuilder()
+		[Obsolete("Use BitcoinUriBuilder(Network) instead")]
+		public BitcoinUrlBuilder() :
+			base(Bitcoin.Instance.Mainnet)
 		{
-			Network = Bitcoin.Instance.Mainnet;
-			scheme = "bitcoin";
-		}
-		public BitcoinUrlBuilder(Network network)
-		{
-			Network = network;
-			scheme = network.UriScheme;
-		}
-		public BitcoinUrlBuilder(Uri uri, Network network)
-			: this(uri.AbsoluteUri, network)
-		{
-			if (uri == null)
-				throw new ArgumentNullException(nameof(uri));
 		}
 
-		public Network Network { get; }
-		string scheme;
-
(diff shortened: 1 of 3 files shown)


NBitcoin/Payment/BitcoinUrlBuilder.cs
```diff
@@ -1,204 +1,29 @@
...
-			}
-			_UnknownParameters = parameters;
-			var reqParam = parameters.Keys.FirstOrDefault(k => k.StartsWith("req-", StringComparison.OrdinalIgnoreCase));
-			if (reqParam != null)
-				throw new FormatException("Non compatible required parameter " + reqParam);
-		}
-
-		private readonly Dictionary<string, string> _UnknownParameters = new Dictionary<string, string>();
-		public IReadOnlyDictionary<string, string> UnknownParameters
+		public BitcoinUrlBuilder(Network network) :
```
knocte, line 14: this needs another Obsolete attribute

NBitcoin/Payment/BitcoinUrlBuilder.cs
```diff
@@ -1,204 +1,29 @@
...
-				}
-
-				WriteParameters(parameters, builder);
-
-				return new System.Uri(builder.ToString(), UriKind.Absolute);
-			}
 		}
 
-		private static void WriteParameters(Dictionary<string, string> parameters, StringBuilder builder)
+		public BitcoinUrlBuilder(Uri uri, Network network) :
```
knocte, line 19: @kristapsk and this too

NBitcoin/Payment/BitcoinUrlBuilder.cs
```diff
@@ -1,204 +1,29 @@
...
-				else
-					builder.Append("&");
-				builder.Append(parameter.Key);
-				builder.Append("=");
-				builder.Append(System.Web.NBitcoin.HttpUtility.UrlEncode(parameter.Value));
-			}
 		}
 
-		public override string ToString()
+		public BitcoinUrlBuilder(string uri, Network network) :
```
knocte, line 24: @kristapsk and this

knocte:
utACK, however, @kristapsk unfortunately this PR needs to be rebased

kristapsk:
Rebased.

Tests pass, but, please, review again.

Kukks (approved):

NicolasDorier:
I am on the fence for this one. This might requires a major release as it's a breaking change.

knocte:
> as it's a breaking change.

Are you sure? Previous consumers of the API will still work because he left a type called BitcoinUrlBuilder there. Only things that NBitcoin consumers will notice is some Obsolete warnings.

augustoproiete:
> > as it's a breaking change.
> 
> Are you sure? Previous consumers of the API will still work because he left a type called BitcoinUrlBuilder there. Only things that NBitcoin consumers will notice is some Obsolete warnings.

It's a **binary** breaking change for sure. The old class now inherits from the new class, so it's no longer the same type across different versions of NBitcoin, so if this were to be merged, it should be released with a major version bump.

An easy way to solve this would be remove the inheritance and use encapsulation instead... Just forwarding all the methods in the old class, to the new class... That would allow for a smooth minor version bump.

NicolasDorier:
@kristapsk can you take simpler approach, not breaking anything: Just leave the `BitcoinUrlBuilder` as it was before, just add the `[Obsolete]` attribute like you did.

Then just duplicate the code for `BitcoinUriBuilder`. I will eventually remove `BitcoinUrlBuilder` in a major release.

Would fix breaking change @augustoproiete pointed out.

kristapsk:
@NicolasDorier Ok, will do that!

NicolasDorier:
https://github.com/MetacoSA/NBitcoin/pull/1308#issuecomment-4703573874

PR #1182, opened 2023-08-31
If one output has amount below dust, throw exception instead of removing it

NicolasDorier:
While it shouldn't really matter economically speaking to remove output with a value that is too small, it is surprising behavior for users, and give rise to several edge case.

Assuming 540 of dust, here is a summary of the changes of behavior:

| Case |      Before |  Now |
|----------|-------------|------|
| you send bob 600 satoshi, but substract 400 sats, the output left is 200 sats. |  Silently remove output | Throw exception |
| you send bob 200 satoshi (without substract fee) |  Silently remove output |   Throw exception |
| you sent bob 600 satoshi, substracted 700 sats of fee | Throw exception |    Throw exception |

Before, we would silently remove such outputs in the built transaction. Now the builder will always throw an exception.

### DIFF
--- a/NBitcoin/TransactionBuilder.cs
+++ b/NBitcoin/TransactionBuilder.cs
@@ -1189,23 +1189,18 @@ namespace NBitcoin
 				{
 					var fee = ctx.CurrentGroupContext.Fee.GetAmount(Money.Zero);
 					txout.Value -= fee;
-
-					var minimumTxOutValue = (parent.DustPrevention ? parent.GetDust(txout.ScriptPubKey) : Money.Zero);
-					if (txout.Value < Money.Zero)
-					{
-						throw new OutputTooSmallException("Can't substract fee from this output because the amount is too small",
-						ctx.Group.Name,
-						-txout.Value
-						);
-					}
 					ctx.CurrentGroupContext.FeePaid = true;
-					if (txout.Value < minimumTxOutValue)
-					{
-						// Between zero and dust, should strip this output.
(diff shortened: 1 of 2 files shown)


NBitcoin/TransactionBuilder.cs
```diff
@@ -1189,25 +1189,20 @@ public void Build(TransactionBuildingContext ctx)
...
-						return;
-					}
 					ctx.CurrentGroupContext.FeeTxOut = txout;
 				}
 
+				var minimumTxOutValue = (parent.DustPrevention ? parent.GetDust(txout.ScriptPubKey) : Money.Zero);
+				if (txout.Value < minimumTxOutValue)
+				{
+					// If the txout is below dust, they throw an exception
+					throw new OutputTooSmallException("Can't substract fee from this output because the amount is too small",
```
farukterzioglu, line 1200: this error is out of `if (SubstractFee` block. But it says `Can't substract fee` 

This error will be thrown even user tries to send w/o SubstractFee, if the amount is less than dust. Kinda mis-leading.

farukterzioglu:
How can we differentiate the cases; 
- User tries to send lower than dust (w/ or w/o fee subtract)
- User tries to send bigger than dust,  but the amount after fee subtraction is less than dust

If we can differentiate the second case, user can try to send it again after fees dropped. 
When the fees are lower, subtracted amount will be less and remaining amount could be bigger than dust. 

But the first case will be always fail. 

Right now both of them throws `OutputTooSmallException` exception, so no way to handle them differently

PR #1183, opened 2023-09-04
Add details to OutputTooSmallException

NicolasDorier:
Ping @farukmeta https://github.com/MetacoSA/NBitcoin/pull/1182#issuecomment-1702470698

### DIFF
--- a/NBitcoin/TransactionBuilder.cs
+++ b/NBitcoin/TransactionBuilder.cs
@@ -340,10 +340,20 @@ namespace NBitcoin
 
 	public class OutputTooSmallException : NotEnoughFundsException
 	{
-		public OutputTooSmallException(string message, string? group, IMoney missing) : base(message, group, missing)
+		public enum ErrorType
 		{
-
+			TooSmallBeforeSubstractedFee,
+			TooSmallAfterSubstractedFee
 		}
+		public OutputTooSmallException(string message, string? group, IMoney missing, ErrorType reason, TxOut output) : base(message, group, missing)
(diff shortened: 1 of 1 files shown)


farukterzioglu:
in [here](https://github.com/MetacoSA/NBitcoin/pull/1182), you said; 

Case | Before | Now
-- | -- | --
you send bob 200 satoshi (without substract fee) | Silently remove output | Throw exception

But right now it looks like it is still silently removed. 
I tried this test but it fails;
```
[Fact]
public void DoNotGenerateTransactionWithDust()
{	
	var k = new Key();
	var scriptCoin = RandomCoin(Money.Coins(0.0003m), k.PubKey.ScriptPubKey, true);
	
	var builder = Network.CreateTransactionBuilder();
	
	var dust = builder.GetDust(new Key().GetScriptPubKey(ScriptPubKeyType.Legacy));

	var ex = Assert.Throws<OutputTooSmallException>(() => {
		builder
			.AddCoins(scriptCoin)
			.Send(new Key(), dust - Money.Satoshis(1))
			.SetChange(new Key())
			.BuildTransaction(false);
	}
	);
	Assert.Equal(OutputTooSmallException.ErrorType.TooSmallBeforeSubstractedFee, ex.Reason);
}
```
It is not throwing any exception if the amount being sent is less than dust. it is being removed (as Before)

@NicolasDorier can you confirm before/after in the table is correct

NicolasDorier:
wtf, seems like a bug to me. I check

PR #1191, opened 2023-10-21
`OutPoint`: Use `CompareTo` in `operator==`

kiminuo:
I have not measured the change but it seems that for many cases it should be faster.


lontivero:
I don't think it improves the perf. It should be measured. 
Also, I think N follows the Benford's law. Most Ns are 0 then 2, then 3 and so on.

PR #1195, opened 2023-12-22
Remove the need of reflection of p2p message (de)serialization

lontivero:
Reflection makes code stripping impossible and that's an impediment for fully supporting AOT. 

This PR is a first step in the removal of reflection and here we start from removing it from the peer-to-peer message serialization/deserialization.

Note: sorry for the eol change in NBitcoin/Protocol/Payloads/AddrPayload.cs, the truth is that NBitcoin has a problem of inconsistency where half of the files are CRLF while the other half are LF so, it is always painful for any contributor.


lontivero:
Ping @NicolasDorier

lontivero:
I reordered the messages based on my experience but I have no numbers to tell me which messages are more frequent.

PR #1204, opened 2024-03-13
Several changes from PKT Cash project

UCIS:
This pull request contains several changes which were needed to use NBitcoin with PKT Cash. The actual network details are currently implemented in a private repository as this is still incomplete. This PR is primarily for comparison and to suggest some improvements, it does not need to be accepted as-is.

1. The filter data of a cfilter payload is a var_str or CompactSize followed by a variable number of bytes, not a zero-size buffer: https://en.bitcoin.it/wiki/BIP_0157#cfilter
2. The unknown payload data was not correctly deserialized because it would read a zero-size buffer
3. The witness marker is specified as two bytes (0001), using an empty TxInList seems excessive and hides the actual purpose
4. Some coins need different transaction validation rules, in our case the maximum supply (hardcoded in MAX_MONEY) is different


NicolasDorier:
Reverting dummy change, it matches what bitcoin core does so if there is one day a change of logic there, we won't crash. It's not big deal.

NicolasDorier:
Mmmh actually not reverting, this is only for writing.

PR #1208, opened 2024-04-15
Fixing that Id should be increment when batching requests with certain RPC providers.

dlebee:
We are currently transitioning our RPCs to Chainstack. When multiple requests are combined into a single HTTP request, we ensure that each RPC request is assigned a unique ID.

### DIFF
--- a/NBitcoin/RPC/RPCClient.cs
+++ b/NBitcoin/RPC/RPCClient.cs
@@ -809,8 +809,11 @@ namespace NBitcoin.RPC
 			if (batches == null)
 				throw new InvalidOperationException("This RPCClient instance is not a batch, use PrepareBatch");
 			_BatchedRequests = null;
+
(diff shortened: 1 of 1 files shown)


marc-aurele-besner (approved):
LGTM 🚀 😂

NicolasDorier:
@dlebee can you confirm if it works? Maybe the results come back unordered and something else need to change (unsure)

dlebee:
I only confirmed its crashing, we had to use child class and reflection to fix it for now, so I can confirm tomorrow.

right now the batch code assumes the response come in the same order they are sent, so I assume the order should remain the same?

NicolasDorier:
I think the assumption that the order is the same hold with a real bitcoin core, but fail with Chainstack. In such case, you should just order the responses based on the id.

PR #1211, opened 2024-05-20
Remove net6.0 build, replace by 8.0

NicolasDorier:


knocte:
Why? .net8 apps can reference .net6/.netstandard2.0 libraries just fine.

NicolasDorier:
@knocte because 6.0 is out of support and I don't have it on my machine anymore to build.

NicolasDorier:
Good point. Then I'll wait a bit.

knocte:
> Then I'll wait a bit.

That sounds sensible!

BTW what is definitely out of support is the .NET3.1 builds, I would remove those.

NicolasDorier:
@knocte already done in this PR

knocte:
Then how about a PR that replaces .NETCore3.1 with .NET8, and we leave .NET6 for a while? (even a while after Nov2024 please 🙏 ). I would be willing to create such PR.

Carti-it:
.NET 6 is now out of support.

NicolasDorier:
Already on master.

Issue #1210, opened 2024-05-12
Support for Testnet4


claytantor:
TestNet4 is coming and it would be awesome if there was an easy way to configure Network.TestNet to use testnet3, testnet4, or whatever. 

I see something like `Network.GetNetwork("testnet3")` in some tests, is it possible to self configure for "testnet4"

in the implementation of `Network.TestNet` I see the following:

```csharp
_TestNet = new Network("TestNet",
				Encoders.Hex.DecodeData("0100000000000000000000000000000000000000000000000000000000000000000000003ba3edfd7a7b12b27ac72c3e67768f617fc81bc3888a51323a9fb8aa4b1e5e4adae5494dffff001d1aa4ae180101000000010000000000000000000000000000000000000000000000000000000000000000ffffffff4d04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73ffffffff0100f2052a01000000434104678afdb0fe5548271967f1a67130b7105cd6a828e03909a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112de5c384df7ba0b8d578a4c702b6bf11d5fac00000000"),
				0x0709110B,
				null,
				NBitcoin.Bitcoin.Instance);
```				

If I got some pointers/breadcrumbs on the appropriate encoding values for `testnet4` I could probably implement it in a branch...

https://github.com/bitcoin/bitcoin/pull/29775

jlopp:
Here's the raw hex of the (current) testnet4 genesis block; it's subject to change until such time as the PR is merged though hopefully it won't.

0100000000000000000000000000000000000000000000000000000000000000000000004e7b2b9128fe0291db0693af2ae418b767e657cd407e80cb1434221eaea7a07a046f3566ffff001dbb0c78170101000000010000000000000000000000000000000000000000000000000000000000000000ffffffff5504ffff001d01044c4c30332f4d61792f323032342030303030303030303030303030303030303030303165626435386332343439373062336161396437383362623030313031316662653865613865393865303065ffffffff0100f2052a010000002321000000000000000000000000000000000000000000000000000000000000000000ac00000000

claytantor:
Do you have any idea what `0x0709110B` represents?

claytantor:
[...]

PR #1216, opened 2024-06-20
Add TestNet4 for Bitcoin

turbolay:
Fixes #1216

This PR doesn't replace `TestNet` by `TestNet4`, it simply adds to `Network` and `Bitcoin` the field `TestNet4`. Most interesting file to look at is [Network.cs](https://github.com/MetacoSA/NBitcoin/pull/1216/files#diff-f8a547ef70a0b0926b695ed245d377c50c5052ca773f4c418a8b334d0cee929d)

I also replicated some of the tests we have for `TestNet` to `TestNet4`

I used @jlopp 's [value](https://github.com/MetacoSA/NBitcoin/issues/1210#issuecomment-2107253830) for network magic.

I however don't know what to use as `vSeeds`:

https://github.com/MetacoSA/NBitcoin/blob/828b7d9ffbc6255b06c40cdf7e46d68d6bd04ca9/NBitcoin/Network.cs#L2288-L2295

### DIFF
--- a/NBitcoin/Bitcoin.cs
+++ b/NBitcoin/Bitcoin.cs
@@ -17,6 +17,7 @@ namespace NBitcoin
 		{
 			CreateMainnet();
 			CreateTestnet();
+			CreateTestnet4();
 			CreateRegtest();
 			CreateSignet();
 			CreateMutinyNet();
--- a/NBitcoin.Tests/NetworkTests.cs
+++ b/NBitcoin.Tests/NetworkTests.cs
@@ -18,6 +18,7 @@ namespace NBitcoin.Tests
 			Assert.Equal(Network.GetNetwork("regtest"), Network.RegTest);
 			Assert.Equal(Network.GetNetwork("testnet"), Network.TestNet);
 			Assert.Equal(Network.GetNetwork("testnet3"), Network.TestNet);
+			Assert.Equal(Network.GetNetwork("testnet4"), Bitcoin.Instance.Testnet4);
 			Assert.Equal(Network.GetNetwork("signet"), Bitcoin.Instance.Signet);
 			Assert.Equal(Network.GetNetwork("mutinynet"), Bitcoin.Instance.Mutinynet);
 			Assert.Null(Network.GetNetwork("invalid"));
--- a/NBitcoin/Bitcoin.Testnet4.cs
+++ b/NBitcoin/Bitcoin.Testnet4.cs
@@ -1,0 +1,116 @@
+using System;
+using NBitcoin.DataEncoders;
+
+namespace NBitcoin
+{
+	public partial class Bitcoin
+	{
+		static readonly ChainName Testnet4Name = new("Testnet4");
+
+		public Network Testnet4 => _Networks[Testnet4Name];
+
+		private Network CreateTestnet4()
+		{
+			NetworkBuilder builder = new NetworkBuilder();
+			builder.SetGenesis("0100000000000000000000000000000000000000000000000000000000000000000000004e7b2b9128fe0291db0693af2ae418b767e657cd407e80cb1434221eaea7a07a046f3566ffff001dbb0c78170101000000010000000000000000000000000000000000000000000000000000000000000000ffffffff5504ffff001d01044c4c30332f4d61792f323032342030303030303030303030303030303030303030303165626435386332343439373062336161396437383362623030313031316662653865613865393865303065ffffffff0100f2052a010000002321000000000000000000000000000000000000000000000000000000000000000000ac00000000");
+			builder.SetMagic(0x283F161C);
+			builder.SetMaxP2PVersion(Network.BITCOIN_MAX_P2P_VERSION);
+			builder.SetUriScheme(null);
+			builder.SetChainName(Testnet4Name);
+			builder.SetName("TestNet4");
+			builder.AddAlias("testnet4");
+			builder.AddAlias("test4");
+			builder.AddAlias("testnet4");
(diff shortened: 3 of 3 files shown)


mathis1337:
These are the only two I am aware of for the vSeeds right now:

```
vSeeds.emplace_back("seed.testnet4.bitcoin.sprovoost.nl."); // Sjors Provoost
vSeeds.emplace_back("seed.testnet4.wiz.biz."); // Jason Maurice
```

You can verify this here: https://github.com/fjahr/bitcoin/blob/2024-04-testnet-4-fix/src/kernel/chainparams.cpp#L350
and the official PR: https://github.com/bitcoin/bitcoin/pull/29775/files#diff-e20339c384d6f19b519ea2de7f4ba4fed92a36d66a80f0339b09927c3fa38d6dR350

mathis1337:
I added a PR to yours, so you can merge in if you want.
https://github.com/turbolay/NBitcoin/pull/1

turbolay:
Thanks!! Not sure why I missed that.
We will review this PR next week with @lontivero and ping NDorier when we will be sure that it's mergeable

lontivero:
ACK. It connects to the p2p network and interacts with it flawlessly. What else should I test? @NicolasDorier ?

NicolasDorier:
Do not add `ChainName.Testnet4`. This is something specific to Bitcoin.
You can use a static variable `Testnet4ChainName` in the `Bitcoin` type, and handle the case in the `GetNetwork`.

NBitcoin/Bitcoin.cs
```diff
@@ -88,10 +88,13 @@ private static uint GetSignetMagic()
 
 		public Network Testnet => Network.TestNet;
 
+		public Network Testnet4 => Network.TestNet4;
+
 		public Network Regtest => Network.RegTest;
 
 		public string CryptoCode => "BTC";
 
+		public static readonly ChainName TestNet4Name = new ChainName("TestNet4");
```
turbolay, line 97: This change is why I took so long to push: I'm not sure if it's correct to have this `public`. I tried another approach and move `TestNet4` to `Bitcoin` class but ended up with a similar (worse probably) problem, [see](https://github.com/MetacoSA/NBitcoin/commit/abc2aec594d02a15e2e254fec8858002e0c95ab7#diff-f8a547ef70a0b0926b695ed245d377c50c5052ca773f4c418a8b334d0cee929dR1911)

turbolay:
@lontivero Can you check at last commit please? 9b4a9aa
I also made [this](https://github.com/MetacoSA/NBitcoin/commit/abc2aec594d02a15e2e254fec8858002e0c95ab7#diff-f8a547ef70a0b0926b695ed245d377c50c5052ca773f4c418a8b334d0cee929dR1911) approach for the PR, not sure which is best as they both look bad.

lontivero:
I don't have context. Why is not okay to have a ChainName for testnet4?

turbolay:
@farukterzioglu I think it's OK, even if the solution doesn't look like the best
Please also considerate this alternative, that makes other kind of compromises: https://github.com/turbolay/NBitcoin/commit/abc2aec594d02a15e2e254fec8858002e0c95ab7#diff-12355e5baf374196ed9c5a0708f16632568b4bfc9139c8daff9aca066fd9f8e1

Feel free to request changes or take over the PR.

farukterzioglu:
@NicolasDorier, above [lontivero](https://github.com/lontivero) confirmed it can connect to the new testnet. And it is refactored as you suggested above. 
Does the rest look good? Would you have additional comment?

Kukks:
Can you rebase? We moved things around to get everything cleaner

turbolay:
> Can you rebase? We moved things around to get everything cleaner

Sorry I am busy lately, thank you for the refactor, it looks good indeed. I will rebase tonight.

NBitcoin/JsonConverters/NetworkJsonConverter.cs
Kukks, line 1: Wow, this is a horrible converter with so many hardcoded stuff. It should use Network natively..

NBitcoin/Network.cs
```diff
@@ -252,6 +252,13 @@ public static Consensus TestNet
 				return Network.TestNet.Consensus;
 			}
 		}
+		public static Consensus TestNet4
```
Kukks, line 255: Remove.  Just use Bitcoin.Instance.Testnet4. testnet4 is a bitcoin-only concept.

NBitcoin/Network.cs
```diff
@@ -807,6 +814,8 @@ public ChainName ChainName
 
 		public static Network TestNet => Bitcoin.Instance.Testnet;
 
+		public static Network TestNet4 => Bitcoin.Instance.Testnet4;
```
Kukks, line 817: Remove.  Just use Bitcoin.Instance.Testnet4. testnet4 is a bitcoin-only concept.

Kukks (changes requested):

NBitcoin/Network.cs
turbolay: Done

NBitcoin/Network.cs
turbolay: Done

amccool:
I know this is late, but was
 `RegisterDefaultCookiePath(Network.TestNet4, testnet);`

missed?

turbolay:
Yes it was, we had the same issue in Wasabi.

amccool:
@turbolay is there an issue for this yet?

PR #1224, opened 2024-09-04
Implement BIP322

Kukks:


NicolasDorier:
>  I would suggest to avoid try{}catch{} with empty catch because it makes the developers' live miserable, any bug is hidden and you will never understand why.

Sadly not possible when parsing a witscript and a transaction without massive refactoring. The try/catch is however quite contained to parsing them.

knocte:
> Sadly not possible when parsing a witscript and a transaction without massive refactoring. The try/catch is however quite contained to parsing them.

But cannot you know the type of the exception that would be thrown at least?

PR #1222, opened 2024-08-29
Support mutinynet

Kukks:
This PR refactors Network and its Bitcoin networks, in that it pulls them to the NetworkSet class to reduce its coupling and make their instantiation similar to how every other network is created. It is IMO a much cleaner result.

Additioanlly, it adds support for MutinyNet, a custom signet network with 30s blocks and many covenant soft forks activated. Most of LN testing infra runs on Mutinynet, and is very useful.

PR #1226, opened 2024-09-06
Adds the ability to register custom Bitcoin signets

sipsorcery:
With testnet full of spam a custom sigent has become more appealing.

This PR is an attempt at allowing NBitcoin to initialise a signet with a custom challenge (no more flakey faucets).

I got this approach to work with nbxplorer to index my custom signet blockchain but despite that I'm not sure if this is the ideal approach or not. Happy to take any advice on alternatives.

### DIFF
--- a/NBitcoin.Tests/NetworkTests.cs
+++ b/NBitcoin.Tests/NetworkTests.cs
@@ -4,6 +4,7 @@ using System.Threading;
 using NBitcoin;
 using Xunit;
 using System;
+using NBitcoin.Crypto;
 
 namespace NBitcoin.Tests
 {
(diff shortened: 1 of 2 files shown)


Kukks:
Does https://github.com/MetacoSA/NBitcoin/pull/1222 help you?

sipsorcery:
Thanks for the follow-up.

> Does #1222 help you?

I don't think so. Isn't MutinyNet just hard coding another signet variation? 

This PR is so the hard coded parameters are not required and NBitcoin can be used with any arbitrary signet challenge.

NicolasDorier:
@sipsorcery concept ACK, but can you rebase as I merged https://github.com/MetacoSA/NBitcoin/pull/1222 which moved some stuff around

sipsorcery:
@Kukks I was looking at the signet and mutinynet initialisation and the only differences are:

- chain name and aliases
- POW target spacing
- network seeds
- challenge

One thing not different, that I suspect should be, is the genesis block. I think NBXplorer uses it to detect the start of the chain. MutinyNet can't be using the same genesis block as the default signet chain given they have different block signing keys. Might not matter if NBXplorer is not being used.

I could add another partial class to initialise a custom signet but it would seem better to create a single bitcoin signet class that can be initialised with different options for the parameters above. wdyt?

dangershony:
ACK good idea, we also use a custom signet.

Kukks:
I believe this should be possible now without this PR. Could you check?

sipsorcery:
@Kukks unless I've missed something there's still no way to register a `custom` signet provider.

I'm running a set of signet nodes that aren't using the default signet or mutinynet signing key. The idea bheind this PR was to be bale to register a network with an arbitrary signet challenge and genesis block.

The approach in this PR is outdated. It would now be better to either use a new Bitcoin.SignetCustom class or avoid a lot of repeated code and combine Bitcoin.Signet, Bitcoin.MutinyNet and BitcoinSignetCustom into a common class with some way to set the different properties, e.g. block spacing, name, challenge and genesis block.

NicolasDorier:
@sipsorcery I think that would be nice yes.

NicolasDorier:
@sipsorcery there seems to have still conflict

knocte:
@sipsorcery but CI is broken in that one.

PR #1228, opened 2024-09-22
Fix: GenerateNonce(ECPubKey) wasn't random as expected

NicolasDorier:
Because an empty `sessionId` was passed instead of null, the nonce wouldn't be randomly generated, using a constant zero byte `sessionId` leading to nonce reuse...


real-or-random:
Note that MuSig2 PR to libsecp256k1 got rid of the harmless sounding name `session_id32` (https://github.com/bitcoin-core/secp256k1/pull/1479#discussion_r1516062204). It's called `session_secrand32` now.

PR #1242, opened 2024-12-17
Fix dotnetcore31 CI lane

webwarrior-ws:
- Set DOTNET_SYSTEM_GLOBALIZATION_INVARIANT environment variable to 1
- Install specific libssl version


NicolasDorier:
> Good work, but I'd vote to remove 3.1 already, it's too old.

I will do during next LTS I think

PR #1252, opened 2025-02-06
Make musig2 private nonce loadable

Kukks:


NicolasDorier:
why you need this?
I avoided to do it to discourage nonce reuse.

Kukks:
> why you need this? I avoided to do it to discourage nonce reuse.

sometimes you need to pre-generate nonces and finish the session later async

PR #1263, opened 2025-04-23
Fix: Incorrect Fee Estimation Due to Omitted estimate_mode in Bitcoin Core 28.0

### LINKED ISSUE #1262: Incorrect Fee Estimation Due to Omitted estimate_mode in Bitcoin Core 28.0
### Background
When using NBitcoin (via NBXplorer) connected to a Bitcoin Core node version 28.0, the returned fee estimates do not align with the expected conservative values, even when `EstimateSmartFeeMode.Conservative` is specified. This appears to be due to a change in Bitcoin Core's default fee estimation behavior and the way NBitcoin constructs the `estimatesmartfee` RPC call.

---

### Bitcoin Core 28.0 Change
According to the [[release notes for Bitcoin Core 28.0](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-28.0.md)](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-28.0.md):

> The default mode for the estimatesmartfee RPC has been updated from conservative to economical...

This means that if the `estimate_mode` is not explicitly provided, Bitcoin Core 28.0 returns the economical fee by default.

---

### Issue in NBitcoin
In [`[RPCClient.cs](https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/RPC/RPCClient.cs)`](https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/RPC/RPCClient.cs), the `EstimateSmartFeeImplAsync` method sets the default `estimate_mode` to Conservative:

```csharp
private async Task<EstimateSmartFeeResponse> EstimateSmartFeeImplAsync(int confirmationTarget, EstimateSmartFeeMode estimateMode = EstimateSmartFeeMode.Conservative, CancellationToken cancellationToken = default)
```

However, due to the following conditional check:

```csharp
if (estimateMode != EstimateSmartFeeMode.Conservative)
{
    parameters.Add(estimateMode.ToString().ToUpperInvariant());
}
```

the `estimateMode` is *not* added to the RPC parameters when it is set to Conservative. This causes Bitcoin Core 28.0 to fall back to its new default: economical.

---

### Suggested Fix
Update the conditional check to always include the `estimate_mode` parameter, regardless of its value:

```csharp
parameters.Add(estimateMode.ToString().ToUpperInvariant());
```

Alternatively, implement version-aware logic to preserve backward compatibility with older Bitcoin Core versions, while ensuring explicit conservative mode behavior for version 28.0 and beyond.

juanmigdr:
Fixes #1262

### DIFF
--- a/NBitcoin/RPC/RPCClient.cs
+++ b/NBitcoin/RPC/RPCClient.cs
@@ -396,7 +396,7 @@ namespace NBitcoin.RPC
 				return;
 			}
 #endif
-			var address = new Key().GetAddress(type, rpc.Network);
+				var address = new Key().GetAddress(type, rpc.Network);
 			if (address == null)
 			{
 				setResult(false);
@@ -1084,7 +1084,7 @@ namespace NBitcoin.RPC
 			return ms;
 		}
 
-		#region P2P Networking
+#region P2P Networking
 #if !NOSOCKET
 		public PeerInfo[] GetPeersInfo()
 		{
@@ -1242,9 +1242,9 @@ namespace NBitcoin.RPC
 		}
 #endif
 
-		#endregion
+#endregion
 
-		#region Block chain and UTXO
+#region Block chain and UTXO
 
 		public async Task<BlockchainInfo> GetBlockchainInfoAsync(CancellationToken cancellationToken = default)
 		{
@@ -1899,13 +1899,13 @@ namespace NBitcoin.RPC
 			return GetTransactions(GetBlockHash(height));
 		}
 
-		#endregion
+#endregion
 
-		#region Coin generation
+#region Coin generation
 
-		#endregion
+#endregion
 
-		#region Raw Transaction
+#region Raw Transaction
 
 		public Transaction DecodeRawTransaction(string rawHex)
 		{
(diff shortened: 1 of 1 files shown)


farukterzioglu:
LGTM

juanmigdr:
I have now made changes through [this commit](https://github.com/MetacoSA/NBitcoin/pull/1263/commits/d76d4bccabec081a51fbe63a4ff19d24a317e2dc) which should cater for the following scenarios:

If the user does not set `estimateMode`:
- If node < 0.28, NO estimateMode is sent in RPC -> We use default "conservative" mode - **SAME**
- if node >= 0.28, 'NO estimateMode is sent in RPC -> We use default "economical" mode - **SAME**

If the user sets `estimateMode` to `conservative`:
- If node < 0.28, NO estimateMode is sent in RPC -> We use default "conservative" mode - **SAME OUTCOME - we just pass the `estimateMode=conservative` in the request**
- if node >= 0.28, 'NO estimateMode is sent in RPC -> We use default "economical" mode - **DIFFERENT OUTCOME - we pass the `estimateMode=conservative` in the request which is then used as the estimationMode**

If the user sets `estimateMode` to `economical`:
- If node < 0.28, `economical` estimateMode is sent in RPC -> We use "economical" mode - **SAME**
- if node >= 0.28, `economical` estimateMode is sent in RPC -> We use "economical" mode - **SAME**

PR #1264, opened 2025-04-24
Incorrect fee estimation - Backport to v7.0.42

farukterzioglu:


juanmigdr:
Maybe worth updating the .csproject version to match v7.0.42.1?

PR #1269, opened 2025-05-15
Fix improper parsing of transaction with some nVersion

NicolasDorier:
The coinbase transaction [889d4fed1ec4a775d02082b5ff727f48d93013469db709d8d435e15b173118f3](https://mempool.space/tx/889d4fed1ec4a775d02082b5ff727f48d93013469db709d8d435e15b173118f3) wasn't properly parsed by NBitcoin, resulting in code crashing when parsing block 896727.

The root cause seems to be that I based my implementation on code that was done pre-release of segwit https://github.com/bitcoin/bitcoin/commit/e37b16a75c1ff86d515ef3050d2c76432ea71225 .

### DIFF
--- a/NBitcoin/Transaction.cs
+++ b/NBitcoin/Transaction.cs
@@ -1530,9 +1530,6 @@ namespace NBitcoin
 			}
 		}
 
-		//Since it is impossible to serialize a transaction with 0 input without problems during deserialization with wit activated, we fit a flag in the version to workaround it
-		protected const uint NoDummyInput = (1 << 27);
-
 #region IBitcoinSerializable Members
 
 		public virtual void ReadWrite(BitcoinStream stream)
@@ -1547,11 +1544,8 @@ namespace NBitcoin
 				/* Try to read the vin. In case the dummy is there, this will be read as an empty vector. */
 				stream.ReadWrite(ref vin);
 				vin.Transaction = this;
-				var hasNoDummy = (nVersion & NoDummyInput) != 0 && vin.Count == 0;
-				if (witSupported && hasNoDummy)
-					nVersion = nVersion & ~NoDummyInput;
 
-				if (vin.Count == 0 && witSupported && !hasNoDummy)
+				if (vin.Count == 0 && witSupported)
 				{
 					/* We read a dummy or an empty vin. */
 					stream.ReadWrite(ref flags);
@@ -1580,8 +1574,11 @@ namespace NBitcoin
 				{
 					/* The witness flag is present, and we support witnesses. */
 					flags ^= 1;
-					Witness wit = new Witness(Inputs);
-					wit.ReadWrite(stream);
+					if (Inputs.Count != 0)
+					{
+						Witness wit = new Witness(Inputs);
+						wit.ReadWrite(stream);
+					}
 				}
 				if (flags != 0)
 				{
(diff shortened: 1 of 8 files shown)


NBitcoin/Transaction.cs
```diff
@@ -1547,11 +1547,8 @@ public virtual void ReadWrite(BitcoinStream stream)
 				/* Try to read the vin. In case the dummy is there, this will be read as an empty vector. */
 				stream.ReadWrite(ref vin);
 				vin.Transaction = this;
-				var hasNoDummy = (nVersion & NoDummyInput) != 0 && vin.Count == 0;
-				if (witSupported && hasNoDummy)
-					nVersion = nVersion & ~NoDummyInput;
 
-				if (vin.Count == 0 && witSupported && !hasNoDummy)
+				if (vin.Count == 0 && witSupported)
```
Roasbeef, line 1548: Not super familiar with the codebase, but the version based check is present below when writing out a transaction. As a result, I think if you tried to write out a segwit transaction with that version that tripped things up in the first place, it may end up writing it out as a non-segwit txn.

Roasbeef: Either way, I'd recommend treating the flag market bytes and the version as distinct components during parsing.

NicolasDorier: This whole dummy things shouldn't have been in the source anymore... I'm now copying the Bitcoin Core implementation to make sure it match.

But this broke my tests, due to parsing for transactions without input... Normally this can't happen in the network, but internally in NBitcoin it happens time to time. So I'm just adapting for this case only.

NBitcoin/Transaction.cs
```diff
@@ -1591,13 +1588,12 @@ public virtual void ReadWrite(BitcoinStream stream)
...
 			{
-				var version = (witSupported && (vin.Count == 0 && vout.Count > 0)) ? nVersion | NoDummyInput : nVersion;
-				stream.ReadWrite(ref version);
+				stream.ReadWrite(ref nVersion);
 
 				if (witSupported)
 				{
 					/* Check whether witnesses need to be serialized. */
-					if (HasWitness)
+					if (HasWitness || Inputs.Count == 0)
```
Roasbeef, line 1596: I don't think you need to gate setting the flag value on if there're inputs or not. Instead, if writing a witness encoding is requested and _any_ of the inputs has a witness, then the market byte should be written.

Roasbeef: From BIP 144: 
> If the witness is empty, the old serialization format must be used.

https://github.com/bitcoin/bips/blob/master/bip-0144.mediawiki#user-content-Serialization:~:text=If%20the%20witness%20is%20empty%2C%20the%20old%20serialization%20format%20must%20be%20used.

Roasbeef: bitcoind also has a check for failing parsing if the witness encoding is used, but no witnesses are found in the transaction.

NicolasDorier:
Ok so I think I will push this, even if tests aren't passing. There are trouble for transactions without inputs which are used internally in some part of NBitcoin. But I guess I will be able to fix that somehow later, I want at least to unblock NBX.

NBitcoin/Transaction.cs
Roasbeef: Ah gotcha, maybe it's used for serializing PSBTs where the template is known but inputs haven't been added? 

I think you should still be able to tighten up this parsing, while still allowing for writing and reading transactions that actually have no inputs.

NicolasDorier: After further analysis, it was only used in tests, and in a single place to estimate the size of a transaction... So I will just forbid to serialize a tx without any input from now.

PR #1304, opened 2026-06-05
Node: Add `VersionHandshakeAsync`

Carti-it:
`Node.VersionHandshake` currently generates unobserved exceptions:

```
2026-06-04 15:54:06.585 [ 1] TRC | Program.cs:138                 | System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (The I/O operation has been aborted because of either a thread exit or an application request.)
 ---> System.Net.Sockets.SocketException (995): The I/O operation has been aborted because of either a thread exit or an application request.
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource<System.Int32>.GetResult(Int16 token)
   at System.Threading.Tasks.ValueTask`1.ValueTaskSourceAsTask.<>c.<.cctor>b__4_0(Object state)
   --- End of inner exception stack trace ---
```

The reason for it is because there are calls like this: 

https://github.com/MetacoSA/NBitcoin/blob/399e7b221ef4f41a44d81256ebb2d81f74942027/NBitcoin/Protocol/Node.cs#L1085

which is not awaited.

This PR is the least intrusive way to fix it by copying code of `Node.VersionHandshake` and making `Node.VersionHandshake` obsolete. 

I hope that this will positively impact: https://github.com/WalletWasabi/WalletWasabi/pull/14651 where I could just close the PR.

@NicolasDorier if this PR makes sense to you, I could try to fix tests to use the new method.

### DIFF
--- a/NBitcoin.Tests/ProtocolTests.cs
+++ b/NBitcoin.Tests/ProtocolTests.cs
@@ -326,7 +326,7 @@ namespace NBitcoin.Tests
 					Eventually(() =>
 					{
 						Assert.NotEmpty(group.ConnectedNodes);
-						Assert.All(group.ConnectedNodes, connectedNode => 
+						Assert.All(group.ConnectedNodes, connectedNode =>
 							Assert.True(connectedNode.RemoteSocketEndpoint.IsEqualTo(node.NodeEndpoint)));
 					});
 				}
@@ -955,12 +955,10 @@ namespace NBitcoin.Tests
 			using (var tester = new NodeServerTester())
 			{
 				tester.Server2.Nonce = tester.Server1.Nonce;
-				var ex =  Assert.ThrowsAny<Exception>(() =>
+				Assert.Throws<InvalidOperationException>(() =>
 				{
 					tester.Node1.VersionHandshake();
 				});
-
-				Assert.True((ex is InvalidOperationException) || (ex is OperationCanceledException));
 			}
 		}
 
--- a/NBitcoin.Tests/ProtocolTests.cs
+++ b/NBitcoin.Tests/ProtocolTests.cs
@@ -955,7 +955,7 @@ namespace NBitcoin.Tests
 			using (var tester = new NodeServerTester())
 			{
 				tester.Server2.Nonce = tester.Server1.Nonce;
-				Assert.Throws<InvalidOperationException>(() =>
+				Assert.Throws<OperationCanceledException>(() =>
 				{
 					tester.Node1.VersionHandshake();
 				});
--- a/NBitcoin.Tests/ProtocolTests.cs
+++ b/NBitcoin.Tests/ProtocolTests.cs
@@ -955,10 +955,12 @@ namespace NBitcoin.Tests
 			using (var tester = new NodeServerTester())
 			{
 				tester.Server2.Nonce = tester.Server1.Nonce;
-				Assert.Throws<OperationCanceledException>(() =>
+				var ex =  Assert.ThrowsAny<Exception>(() =>
 				{
 					tester.Node1.VersionHandshake();
 				});
+
+				Assert.True((ex is InvalidOperationException) || (ex is OperationCanceledException));
 			}
 		}
 
--- a/NBitcoin/Protocol/Node.cs
+++ b/NBitcoin/Protocol/Node.cs
@@ -589,7 +589,7 @@ namespace NBitcoin.Protocol
 				int socksFail = 0;
 				while (true)
 				{
-					
+
 					if (groupFail > 50 || socksFail > 50)
 					{
 						parameters.ConnectCancellation.WaitHandle.WaitOne((int)TimeSpan.FromSeconds(60).TotalMilliseconds);
@@ -930,7 +930,7 @@ namespace NBitcoin.Protocol
 		/// Send a message to the peer asynchronously
 		/// </summary>
 		/// <param name="payload">The payload to send</param>
-		/// <param name="System.OperationCanceledException">The node has been disconnected</param>
+		/// <param name="System.InvalidOperationException">The node has been disconnected</param>
 		public Task SendMessageAsync(Payload payload)
 		{
 			if (payload is null)
@@ -943,7 +943,7 @@ namespace NBitcoin.Protocol
 #endif
 			if (!IsConnected)
 			{
-				completion.SetException(new OperationCanceledException("The peer has been disconnected"));
+				completion.SetException(new InvalidOperationException("The peer has been disconnected"));
 				return completion.Task;
 			}
 
@@ -1140,7 +1140,7 @@ namespace NBitcoin.Protocol
 #nullable disable
 
 		/// <summary>
-		/// 
+		///
 		/// </summary>
 		/// <param name="cancellation"></param>
 		public void RespondToHandShake(CancellationToken cancellation = default(CancellationToken))
--- a/NBitcoin/Protocol/Node.cs
+++ b/NBitcoin/Protocol/Node.cs
@@ -1092,6 +1092,10 @@ namespace NBitcoin.Protocol
 		{
 			VersionHandshakeAsync(requirements, cancellationToken).GetAwaiter().GetResult();
 		}
+		public Task VersionHandshakeAsync(CancellationToken cancellationToken = default(CancellationToken))
+		{
+			return VersionHandshakeAsync(null, cancellationToken);
+		}
 
 #nullable enable
 		public async Task VersionHandshakeAsync(NodeRequirement? requirements, CancellationToken cancellationToken = default(CancellationToken))
@@ -1106,7 +1110,7 @@ namespace NBitcoin.Protocol
 
 				await SendMessageAsync(MyVersion).ConfigureAwait(false);
 				#if !NO_CHANNELS
-				var version = await listener.ReceivePayloadAsync<VersionPayload>(cancellationToken);
+				var version = await listener.ReceivePayloadAsync<VersionPayload>(cancellationToken).ConfigureAwait(false);
 				#else
 				var version = listener.ReceivePayload<VersionPayload>(cancellationToken);
 				#endif
@@ -1141,7 +1145,7 @@ namespace NBitcoin.Protocol
 				await SendMessageAsync(new VerAckPayload()).ConfigureAwait(false);
 
 				#if !NO_CHANNELS
-				await listener.ReceivePayloadAsync<VerAckPayload>(cancellationToken);
+				await listener.ReceivePayloadAsync<VerAckPayload>(cancellationToken).ConfigureAwait(false);
 				#else
 				listener.ReceivePayload<VerAckPayload>(cancellationToken);
 				#endif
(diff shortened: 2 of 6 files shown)


NBitcoin/Protocol/Node.cs
```diff
@@ -1131,7 +1134,66 @@ public VersionPayload PeerVersion
...
+		{
+			if (State == NodeState.HandShaked)
+				throw new InvalidOperationException("Already handshaked");
+			requirements = requirements ?? new NodeRequirement();
+			using (var listener = CreateListener()
+									.Where(p => p.Message.Payload is VersionPayload ||
+												p.Message.Payload is VerAckPayload))
+			{
+
+				await SendMessageAsync(MyVersion).ConfigureAwait(false);
```
Carti-it, line 1111: Added missing `await` here

NBitcoin/Protocol/Node.cs
```diff
@@ -1131,7 +1134,66 @@ public VersionPayload PeerVersion
...
+				{
+					Disconnect("The peer does not support the required services requirement");
+					return;
+				}
+
+				// As a courtesy we do not send sendaddr to nodes that do not support it.
+				if (ProtocolCapabilities.SupportAddrv2)
+				{
+					// Signal ADDRv2 support (BIP155).
+					await SendMessageAsync(new SendAddrV2Payload()).ConfigureAwait(false);
```
Carti-it, line 1142: Added missing `await` here

NBitcoin/Protocol/Node.cs
```diff
@@ -1131,7 +1134,66 @@ public VersionPayload PeerVersion
...
+				}
+
+				await SendMessageAsync(new VerAckPayload()).ConfigureAwait(false);
```
Carti-it, line 1145: Added missing `await` here

NBitcoin/Protocol/Node.cs
```diff
@@ -1131,7 +1134,66 @@ public VersionPayload PeerVersion
...
+				listener.ReceivePayload<VerAckPayload>(cancellationToken);
+
+				State = NodeState.HandShaked;
 
+				if (Advertize)
+				{
+					if (MyVersion.AddressFrom is IPEndPoint iPEndPoint && !iPEndPoint.Address.IsRoutable(true))
+						return;
+
+					await SendMessageAsync(new AddrPayload(new NetworkAddress(MyVersion.AddressFrom)
```
Carti-it, line 1160: Added missing `await` here

NBitcoin/Protocol/Node.cs
```diff
@@ -1131,7 +1134,66 @@ public VersionPayload PeerVersion
 			}
 		}
 
+#nullable enable
```
Carti-it, line 1100: I marked `NodeRequirement? requirements` (one line below) to be nullable because that's how it works (see line 1074)

NBitcoin/Protocol/Node.cs
```diff
@@ -1068,10 +1068,13 @@ public VersionPayload PeerVersion
 			}
 		}
 
+		[Obsolete("Use VersionHandshakeAsync instead")]
```
Carti-it: Alternatively I can just remove this line. If it is too intrusive.

NBitcoin/Protocol/Node.cs
```diff
@@ -1068,10 +1068,13 @@ public VersionPayload PeerVersion
...
 		public void VersionHandshake(CancellationToken cancellationToken = default(CancellationToken))
 		{
 			VersionHandshake(null, cancellationToken);
 		}
+
+		[Obsolete("Use VersionHandshakeAsync instead")]
 		public void VersionHandshake(NodeRequirement requirements, CancellationToken cancellationToken = default(CancellationToken))
 		{
 			if (State == NodeState.HandShaked)
```
Carti-it: This method can potentially contain just:

```cs
VersionHandshakeAsync.GetAwaiter().GetResult();
```

Carti-it:
@NicolasDorier What do you think please?

NicolasDorier:
Do not obsolete the sync version, we use it everywhere, and forward to `.GetAwaiter().GetResult();`.
 
In `VersionHandshakeAsync`, make sure every `await` has `.ConfigureAwait(false)`, else the sync over async might deadlock. (check also that this is the case in the methods where you call await)

Carti-it:
Thank you for reviewing

> Do not obsolete the sync version, we use it everywhere, and forward to `.GetAwaiter().GetResult();`.

I did it in 069358853dbc86723f51dae38ef50d7a532848ba

> In `VersionHandshakeAsync`, make sure every `await` has `.ConfigureAwait(false)`, else the sync over async might deadlock. 

Yes, I added `ConfigureAwait(false).

> (check also that this is the case in the methods where you call await)

`SendMessageAsync` does not await anything, so I think it's OK hopefully.

NBitcoin.Tests/ProtocolTests.cs
```diff
@@ -955,7 +955,7 @@ public void CantConnectToYourself()
 			using (var tester = new NodeServerTester())
 			{
 				tester.Server2.Nonce = tester.Server1.Nonce;
-				Assert.Throws<InvalidOperationException>(() =>
+				Assert.Throws<OperationCanceledException>(() =>
```
Carti-it: Node.cs [contains](https://github.com/MetacoSA/NBitcoin/blob/399e7b221ef4f41a44d81256ebb2d81f74942027/NBitcoin/Protocol/Node.cs#L934-L963): 

```cs
public Task SendMessageAsync(Payload payload)
{
	if (payload is null)
		throw new ArgumentNullException(nameof(payload));

#if NO_RCA
	TaskCompletionSource<bool> completion = new TaskCompletionSource<bool>();
#else
	TaskCompletionSource<bool> completion = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
#endif
	if (!IsConnected)
	{
		// !!!!!!!!!!!!!!!!!!!!! THIS IS WHAT IS THROWN. !!!!!!!!!!!!!!!!
		completion.SetException(new OperationCanceledException("The peer has been disconnected"));
		return completion.Task;
	}

	var activity = Guid.NewGuid();

	Action final = () =>
	{
		_Connection.Messages.Add(new SentMessage()
		{
			Payload = payload,
			ActivityId = activity,
			Completion = completion
		});
	};
	OnSendingMessage(payload, final);
	return completion.Task;
}
```

Carti-it: @NicolasDorier This is the only test that failed for me. But running CI tests would help.

NicolasDorier:
Take a look at `CantConnectToYourself` test, it fails to pass it seems with `OperationCancelledException` my guess is that it timeout rather than crash fast like before. Can you take a look?

Carti-it:
> Take a look at `CantConnectToYourself` test, it fails to pass it seems with `OperationCancelledException` my guess is that it timeout rather than crash fast like before. Can you take a look?

There are two scenarios. The first scenario ends up with `InvalidOperationException`:

- [`ProtocolTests.CantConnectToYourself`] `tester.Node1.VersionHandshake()` is [called](https://github.com/Carti-it/NBitcoin/blob/c459a9acead77f1796bd7c0034453ea3d87569a1/NBitcoin.Tests/ProtocolTests.cs#L960)
    - [`Node.VersionHandshakeAsync`] `var version = listener.ReceivePayload<VersionPayload>(cT);` is [called](https://github.com/Carti-it/NBitcoin/blob/c459a9acead77f1796bd7c0034453ea3d87569a1/NBitcoin/Protocol/Node.cs#L1093)
        - [`NodeListener.ReceivePayload`] `throw new InvalidOperationException("The node is not in a connected state")`  is [called](https://github.com/Carti-it/NBitcoin/blob/c459a9acead77f1796bd7c0034453ea3d87569a1/NBitcoin/Protocol/NodeListener.cs#L68)

The second scenario ends up with `OperationCancelledException`:

- [`ProtocolTests.CantConnectToYourself`] `tester.Node1.VersionHandshake()` is [called](https://github.com/Carti-it/NBitcoin/blob/c459a9acead77f1796bd7c0034453ea3d87569a1/NBitcoin.Tests/ProtocolTests.cs#L960)
    - [`Node.VersionHandshakeAsync`] `await SendMessageAsync(new SendAddrV2Payload()).ConfigureAwait(false);` is [called](https://github.com/Carti-it/NBitcoin/blob/c459a9acead77f1796bd7c0034453ea3d87569a1/NBitcoin/Protocol/Node.cs#L1119)
        - [`Node.SendMessageAsync`] `completion.SetException(new OperationCanceledException("The peer has been disconnected"))` is [called](https://github.com/Carti-it/NBitcoin/blob/c459a9acead77f1796bd7c0034453ea3d87569a1/NBitcoin/Protocol/Node.cs#L946)

Basically, `NodeServer` disconnects the peer in this test either here or here:

* https://github.com/Carti-it/NBitcoin/blob/49c34dbebdee76c91232f7dee90ffb834a9b88c2/NBitcoin/Protocol/NodeServer.cs#L281
* https://github.com/Carti-it/NBitcoin/blob/49c34dbebdee76c91232f7dee90ffb834a9b88c2/NBitcoin/Protocol/NodeServer.cs#L304

and it takes certain amount of time to disconnect the node. Consequently, either `InvalidOperationException` or `OperationCancelledException` can be thrown based on how fast the disconnect is.

**tl;dr: `VersionHandshakeAsync` contains always the pair `SendMessageAsync` + `ReceivePayload` like:**

```cs
await SendMessageAsync(MyVersion).ConfigureAwait(false);
var version = listener.ReceivePayload<VersionPayload>(cancellationToken);
```

On `master`, only the second command can throw (because `SendMessageAsync` is not awaited) and the exception is `InvalidOperationException`.

In this PR, both `SendMessageAsync` (can throw `OperationCanceledException` because `SendMessageAsync`) and `ReceivePayload` (throws `OperationCanceledException`) can throw.

If my last commit is fine, then there is nothing else to do. 

Alternatively, one could change:

```cs
await SendMessageAsync(MyVersion).ConfigureAwait(false);
var version = listener.ReceivePayload<VersionPayload>(cancellationToken);
```

to

```cs
try
{
	await SendMessageAsync(MyVersion).ConfigureAwait(false);
}
catch (OperationCanceledException)
{
	if (!IsConnected)
		throw new InvalidOperationException("The node is not in a connected state");

	throw;
}

var version = listener.ReceivePayload<VersionPayload>(cancellationToken);
```

to avoid introducing a new exception in this specific case (`VersionHandshake` that is).

Carti-it:
@NicolasDorier Could you please rerun tests?

NicolasDorier:
You changed the test rather than fixing the bug... did it

NicolasDorier:
Pushed on 10.0.6

PR #1305, opened 2026-06-07
Enforce CompactSize canonical representation

### LINKED ISSUE #1303: `GolombRiceFilter(byte[])` slices filter data at the wrong offset for non-minimal `VarInt(N)`
Opus 4.8 found this bug while I was working on something else. I think this might affect multiple projects that use NBitcoin. Although, I have not tested it.

When a serialized filter is parsed, the number-of-elements prefix N is read with VarInt.StaticRead, which accepts non-minimal VarInt encodings. The constructor then computes the data offset by re-serializing N (n.ToBytes().Length), which always yields the minimal length. For a non-minimally-encoded N, the re-serialized length is smaller than the bytes actually consumed, so Data is sliced at the wrong offset and the leftover VarInt bytes are left as garbage at the front of the Golomb-Rice stream. The filter then decodes incorrectly, producing false negatives (a light client silently misses entries that are actually in the filter).

```C#
public GolombRiceFilter(byte[] data, byte p, uint m)
{
    P = p;
    M = m;
    var n = new VarInt();
    var stream = new BitcoinStream(data);
    stream.ReadWrite(ref n);          // StaticRead consumes the ACTUAL bytes (may be non-minimal)
    N = (int)n.ToLong();              // (secondary) ulong -> int truncation
    var l = n.ToBytes().Length;       // re-serializes MINIMALLY -> wrong length
    Data = data.SafeSubarray(l);      // slices at wrong offset -> corrupt stream
}
```

## Cause

- VarInt.StaticRead ([Protocol/VarInt.cs:130-158](https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/Protocol/VarInt.cs)) does not enforce minimal encoding: e.g. the value 5 is accepted as FD 05 00 (3 bytes) or FF 05 00 00 00 00 00 00 00 (9 bytes).
- VarInt.ToBytes() / StaticWrite always emit the minimal form (05, 1 byte).
- So l = n.ToBytes().Length ≠ bytes consumed from data whenever the input VarInt is non-minimal.

Honest producers (including NBitcoin's own GolombRiceFilterBuilder / ToBytes(), which writes new VarInt((ulong)N).ToBytes()) always emit minimal VarInts, so this never triggers by accident — only with a malformed/adversarial filter.

## Reproduction

Key-independent proof — same N, different Data:

```
using System;
using System.Linq;
using NBitcoin.BIP158;

// Any valid filter produced by NBitcoin (minimal VarInt(N) || data), N < 0xFD for brevity
var good = new GolombRiceFilterBuilder()
    .SetKey(uint256.Zero)
    .AddEntries(new[] { new byte[]{1,2,3}, new byte[]{4,5,6} })
    .Build();

byte[] serialized = good.ToBytes();          // [N][ ...data... ]
byte n = serialized[0];                       // minimal 1-byte N
byte[] dataPart = serialized.Skip(1).ToArray();

// Re-encode the SAME N non-minimally as a 3-byte VarInt: 0xFD <ushort LE>
byte[] evilBytes = new byte[]{ 0xFD, n, 0x00 }.Concat(dataPart).ToArray();

var f1 = new GolombRiceFilter(serialized);
var f2 = new GolombRiceFilter(evilBytes);

Console.WriteLine(f1.N == f2.N);                                   // True  (both N)
Console.WriteLine(f1.Data.SequenceEqual(f2.Data));                // False (BUG: f2.Data has 2 garbage leading bytes)
```
Consequence at the matching layer (false negative) — using the same key the filter was built with (BIP158: first 16 bytes of the block hash):

```
byte[] key = uint256.Zero.ToBytes().Take(16).ToArray();
Console.WriteLine(f1.Match(new byte[]{1,2,3}, key));   // True
Console.WriteLine(f2.Match(new byte[]{1,2,3}, key));   // False  <-- silently misses an entry that IS in the filte
```

## Impact

- A light client that parses a GCS filter from an untrusted/buggy source gets silent false negatives: it fails to match scripts that the filter actually contains, so it can miss blocks paying its own addresses (incorrect balance / missed incoming funds / privacy-relevant scan gaps), with no exception and no recovery short of a rescan.
- This is directly reachable for any consumer fetching filters over BIP157 P2P (raw cfilter bytes from arbitrary peers fed straight into new GolombRiceFilter(bytes)).
- Secondary: N = (int)n.ToLong() truncates ulong → int; a crafted large N becomes negative and corrupts the modulus ((ulong)N * M in ConstructHashedSet), another false-negative path.

lontivero:
This PR enforces CompactSize (VarInt) canonical decoding to prevent array size manipulation. I think the non-canonical representation enables manipulations like the one reported in #1303 and makes no sense or at least it has no utility other than to contradict the purpose of VarInt.

If this is not okay, I can implement a more local fix.

Close https://github.com/MetacoSA/NBitcoin/issues/1303

### DIFF
--- a/NBitcoin.Tests/script_tests.cs
+++ b/NBitcoin.Tests/script_tests.cs
@@ -1105,17 +1105,15 @@ namespace NBitcoin.Tests
 			}
 
 			// Test on non canonic values
-			testCases = new (byte[] Bytes, ulong Value)[]
+			var nonCanonicalTestCases = new byte[][]
 			{
-				(Bytes: new byte[] { 0xFD, 0x01, 0x00 }, Value: 0x01),
-				(Bytes: new byte[] { 0xFE, 0x01, 0x00, 0x00, 0x00 }, Value: 0x01),
-				(Bytes: new byte[] { 0xFF, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }, Value: 0x01)
+				new byte[] { 0xFD, 0x01, 0x00 },
+				new byte[] { 0xFE, 0x01, 0x00, 0x00, 0x00 },
+				new byte[] { 0xFF, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 },
 			};
-			foreach (var testCase in testCases)
-			{
-				var stream = new BitcoinStream(testCase.Bytes);
-				Assert.Equal(testCase.Value, VarInt.StaticRead(stream));
-			}
+			Assert.All(nonCanonicalTestCases,
+				testCase => Assert.Throws<InvalidDataException>(
+					()=> VarInt.StaticRead(new BitcoinStream(testCase))));
 		}
 
 		[Fact]
--- a/NBitcoin/Protocol/VarInt.cs
+++ b/NBitcoin/Protocol/VarInt.cs
@@ -141,18 +141,24 @@ namespace NBitcoin.Protocol
 			{
 				var value = (ushort)0;
 				bs.ReadWrite(ref value);
+				if (value < 0xFD)
+					throw new InvalidDataException("non-canonical representation for varint");
 				return value;
 			}
 			else if (prefix == 0xFE)
 			{
 				var value = (uint)0;
 				bs.ReadWrite(ref value);
+				if (value < 0x10000)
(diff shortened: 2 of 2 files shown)


NicolasDorier:
Pushed 10.0.5

PR #1306, opened 2026-06-09
Modernize `Stream.ReadEx` 

Carti-it:
On Wasabi side, I can still see some unobserved exception. I suspect that `Stream.ReadEx` might be another potential source of unobserved exceptions so I tried to modernize it.

### DIFF
--- a/NBitcoin/Utils.cs
+++ b/NBitcoin/Utils.cs
@@ -315,24 +315,15 @@ namespace NBitcoin
 			if (offset > buffer.Length - count)
 				throw new ArgumentOutOfRangeException("count");
 
-#if NET8_0_OR_GREATER
-			try
-			{
-				stream.ReadExactlyAsync(buffer, offset, count, cancellation).GetAwaiter().GetResult();
-				return count;
-			}
-			catch (EndOfStreamException)
-			{
-				return 0;
-			}
-#else
 			int totalReadCount = 0;
 
 			while (totalReadCount < count)
 			{
-				cancellation.ThrowIfCancellationRequested();
-
+#if NET8_0_OR_GREATER
+				int currentReadCount = stream.ReadAsync(buffer, offset + totalReadCount, count - totalReadCount, cancellation).GetAwaiter().GetResult();
+#else
 				int currentReadCount;
+				cancellation.ThrowIfCancellationRequested();
 
 				//Big performance problem with BeginRead for other stream types than NetworkStream.
 				//Only take the slow path if cancellation is possible.
@@ -356,6 +347,7 @@ namespace NBitcoin
 					//IO interruption not supported in this path.
 					currentReadCount = stream.Read(buffer, offset + totalReadCount, count - totalReadCount);
 				}
+#endif
 
 				if (currentReadCount == 0)
 					return 0;
@@ -364,7 +356,6 @@ namespace NBitcoin
 			}
 
 			return totalReadCount;
-#endif
 		}
 
 #if HAS_SPAN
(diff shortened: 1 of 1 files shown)


NBitcoin/Utils.cs
```diff
@@ -302,7 +302,6 @@ public static IEnumerable<List<T>> Partition<T>(this IEnumerable<T> source, Func
 			}
 		}
 
-#if !NETSTANDARD1X
```
Carti-it, line 305: It is my understanding that .NET Standard support 1.x was removed in #1243 (commit https://github.com/MetacoSA/NBitcoin/commit/4780a7f38189444e9961fa952a0ecea29721c90b).

NBitcoin/Utils.cs
```diff
@@ -355,44 +329,6 @@ public static IEnumerable<List<T>> Partition<T>(this IEnumerable<T> source, Func
 
 			return totalReadCount;
 		}
-#else
```
Carti-it, line 358: Given that .NET Standard 1.x support was removed, the following code is dead code in my opinion.

NBitcoin/Utils.cs
```diff
@@ -320,32 +319,7 @@ public static IEnumerable<List<T>> Partition<T>(this IEnumerable<T> source, Func
...
-					cancellation.ThrowIfCancellationRequested();
-
-					currentReadCount = stream.EndRead(ar);
-				}
-				else
-				{
-					//IO interruption not supported in this path.
-					currentReadCount = stream.Read(buffer, offset + totalReadCount, count - totalReadCount);
-				}
+				int currentReadCount = stream.ReadAsync(buffer, offset + totalReadCount, count - totalReadCount, cancellation).GetAwaiter().GetResult();
```
Carti-it, line 322: https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=net-10.0 mentions that this overload is support on:

| Technology       | Versions |
|------------------|----------|
| .NET Core        | 1.0, 1.1, 2.0, 2.1, 2.2, 3.0, 3.1, 5, 6, 7, 8, 9, 10, 11 |
| .NET Framework   | 4.5, 4.5.1, 4.5.2, 4.6, 4.6.1, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8, 4.8.1 |
| .NET Standard    | 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 2.0, 2.1 |
| UWP              | 10.0 |

My understanding is that those are all supported platforms for NBitcoin. To be sure, I checked https://www.nuget.org/packages/NBitcoin#supportedframeworks-body-tab as well. I don't see anything missing. 


btw: I considered `stream.Read()` method but it does not support cancellation so I use `stream.ReadAsync()`.

NBitcoin/Utils.cs
Carti-it: #1307 attempts to remove the remaining .NET Standard 1.x code

NicolasDorier:
This code saw some shit, so I would prefer not touching what is not broken. 

Why do you thing here is a source of unobserved exception?

Carti-it:
> Why do you thing here is a source of unobserved exception?

I found this PR https://github.com/WalletWasabi/WalletWasabi/pull/8081 and https://github.com/MetacoSA/NBitcoin/pull/1097 which suggests so. However, that's all I know.

Carti-it:
I will try to dig more into it and if I find an actual proof, I will reopen the PR.

PR #1307, opened 2026-06-09
Remove .NET Standard 1.x leftovers

Carti-it:


NicolasDorier:
ACK, but I'd prefer you remove https://github.com/MetacoSA/NBitcoin/pull/1306, because I don't think we should touch it.

Carti-it:
> ACK, but I'd prefer you remove #1306, because I don't think we should touch it.

I removed it.

PR #1308, opened 2026-06-10
Add `BitcoinUriBuilder` and obsolete `BitcoinUrlBuilder`

Carti-it:
Replaces #1168

As @kristapsk says:

> BIP21 contains [URIs](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier), not [URLs](https://en.wikipedia.org/wiki/URL), this is correct terminology (every URL is URI, but not every URI is URL).
>
> Original discussion - https://github.com/MetacoSA/NBitcoin/issues/138#issuecomment-1503402357.

This PR was done according to this comment: https://github.com/MetacoSA/NBitcoin/pull/1168#issuecomment-1555915375

cc @knocte (https://github.com/MetacoSA/NBitcoin/issues/138#issuecomment-1503424978)

### DIFF
--- a/NBitcoin.Tests/PaymentTests.cs
+++ b/NBitcoin.Tests/PaymentTests.cs
@@ -1,96 +1,81 @@
-﻿#if !NOFILEIO
-using NBitcoin.DataEncoders;
-using NBitcoin.Payment;
+﻿using NBitcoin.Payment;
 using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using System.Net;
-using System.Net.Http;
-using System.Security.Cryptography.X509Certificates;
-using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
 using Xunit;
 
-namespace NBitcoin.Tests
+namespace NBitcoin.Tests;
+
+//https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki
+//Their examples are broken
+public class PaymentTests
 {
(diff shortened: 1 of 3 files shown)


NBitcoin.Tests/PaymentTests.cs
```diff
@@ -1,96 +1,81 @@
...
-using System.Net;
-using System.Net.Http;
-using System.Security.Cryptography.X509Certificates;
-using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
 using Xunit;
 
-namespace NBitcoin.Tests
+namespace NBitcoin.Tests;
```
Carti-it, line 5: It's easier to review this file with https://github.com/MetacoSA/NBitcoin/pull/1308/changes?w=1

NBitcoin.Tests/PaymentTests.cs
```diff
@@ -1,96 +1,81 @@
-﻿#if !NOFILEIO
```
Carti-it, line 1: `NOFILEIO` was removed in https://github.com/MetacoSA/NBitcoin/commit/d99cc4ee#diff-b0a54aaf2fe58a54697a57a54b10a61c5c8d30febfb4382c85388c37a0e92f37

NicolasDorier:
@Carti-it I will close this because it is not worth bothering users for such a small thing.
That said, the NOFILEIO cleanup is a good idea.

PR #1314, opened 2026-06-14
`ConcurrentChain`: Optimize `Load` method

Carti-it:
This PR optimizes `ConcurrentChain(byte[] bytes)` method by removing LINQ use and using a simple while loop. On my machine, it translates to ~14% improvement when using the constructor.

## Benchmark

I started Wasabi (https://github.com/WalletWasabi/WalletWasabi/commit/b28faafed239e7a6c72b92b4864e6623f6abb6cb) and it generates `WalletWasabi/Client/BitcoinP2pNetwork/BlockHeadersMain.dat` (~104 MB) and used that file for benchmarking purposes (benchmark was added in 920bb363 and removed in 71169786):

```cs
using BenchmarkDotNet.Attributes;
using System;
using System.IO;

namespace NBitcoin.Bench;

public class ConcurrentChainBench
{
	byte[] Bytes; 

	[GlobalSetup]
	public void Setup()
	{
		var appDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "WalletWasabi", "Client", "BitcoinP2pNetwork", "BlockHeadersMain.dat");
		Bytes = File.ReadAllBytes(appDataPath);
	}

	[Benchmark]
	public void WriteToString()
	{
		var chain = new ConcurrentChain(Bytes, Network.Main.Consensus.ConsensusFactory);
	}
}
```

```
cd NBitcoin.Bench
dotnet run -c Release -- --runtimes net8.0 --filter *ConcurrentChainBench*
```

master:

| Method        | Mean    | Error    | StdDev   |
|-------------- |--------:|---------:|---------:|
| WriteToString | 1.507 s | 0.0298 s | 0.0306 s |

PR:

| Method        | Mean    | Error    | StdDev   |
|-------------- |--------:|---------:|---------:|
| WriteToString | 1.299 s | 0.0214 s | 0.0189 s |

So it's about 200 ms faster, or 14% faster.

In a way, I believe that Wasabi should optimize the [header chain loading](https://github.com/WalletWasabi/WalletWasabi/issues/14670) differently. However, given I spent some time with this, I figured it may be worth a PR.


NicolasDorier:
Checking... but I think that people should avoid having to get the `ConcurrentChain`, because it is taking way too much memory in general.

PR #1316, opened 2026-06-14
Optimize SHA computations for .NET 8+

Carti-it:
Stacked on #1315 to avoid conflicts

master:

| Method | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------- |---------:|---------:|---------:|-------:|----------:|
| SHA1   | 546.5 ns | 10.91 ns | 25.06 ns | 0.0353 |     224 B |
| SHA256 | 578.6 ns | 11.50 ns | 19.52 ns | 0.0381 |     240 B |
| SHA512 | 679.1 ns | 13.48 ns | 24.30 ns | 0.0477 |     304 B |

PR:

| Method | Mean     | Error   | StdDev   | Median   | Gen0   | Allocated |
|------- |---------:|--------:|---------:|---------:|-------:|----------:|
| SHA1   | 300.7 ns | 6.01 ns | 10.68 ns | 305.7 ns | 0.0076 |      48 B |
| SHA256 | 365.5 ns | 7.29 ns | 10.46 ns | 370.5 ns | 0.0086 |      56 B |
| SHA512 | 445.5 ns | 4.86 ns |  3.79 ns | 444.7 ns | 0.0138 |      88 B |

The CPU time improvements are in the range of 35% - 45% and memory improvements are 70%+.

### DIFF
--- a/NBitcoin/Crypto/Hashes.cs
+++ b/NBitcoin/Crypto/Hashes.cs
@@ -106,6 +106,8 @@ namespace NBitcoin.Crypto
 			byte[] rv = new byte[20];
 			sha1.DoFinal(rv, 0);
 			return rv;
+#elif NET8_0_OR_GREATER
+			return System.Security.Cryptography.SHA1.HashData(data.AsSpan(offset, count));
 #else
 			using (var sha1 = System.Security.Cryptography.SHA1.Create())
 			{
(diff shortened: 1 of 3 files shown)


NBitcoin/Crypto/Hashes.cs
```diff
@@ -650,18 +636,24 @@ public static byte[] SHA256(byte[] data)
 #if HAS_SPAN
 		public static byte[] SHA256(ReadOnlySpan<byte> data)
 		{
+#if NET8_0_OR_GREATER
+			return System.Security.Cryptography.SHA256.HashData(data);
```
Carti-it, line 640: This is to avoid `.ToArray()` copy.

PR #1329, opened 2026-08-13
Enforce BIP94 (TestNet4) difficult adjustment logic

### LINKED ISSUE #1328: TestNet4: header sync fails PoW validation (Node.GetChain throws); Core 31.1 accepts the same chain
**Version:** NBitcoin 10.0.7 (as shipped in Wasabi Wallet 2.8.1), Windows x64

**Repro:**
```csharp
var net = Network.GetNetworks().First(n => n.Name.EndsWith("TestNet4", StringComparison.OrdinalIgnoreCase));
var node = Node.Connect(net, new IPEndPoint(IPAddress.Parse("80.253.94.252"), 48333)); // any current testnet4 peer
node.VersionHandshake();
var chain = node.GetChain(new SynchronizeChainOptions(), cts.Token);
```
**Actual:** `ProtocolException: An header which does not pass proof of work verification has been received`
**Expected:** full header chain (~148k as of 2026-08-12).

**Notes:** A consumer (Wasabi 2.8.1) building the chain through this path wedges at exactly height 30,239 across restarts. 30,240 is a retarget boundary (15 x 2016). Testnet4 is roughly 85-90% min-difficulty CPU-mined blocks, so this looks like the min-difficulty/retarget interplay in TestNet4's difficulty validation. Bitcoin Core 31.1 syncs and validates the same chain without complaint.

lontivero:
Closes https://github.com/MetacoSA/NBitcoin/issues/1328


NBitcoin/Bitcoin.Testnet4.cs
```diff
@@ -65,6 +65,7 @@ private Network CreateTestnet4()
 				SupportTaproot = true,
 				SupportSegwit = true,
 				CoinType = 1,
+				EnforceBIP94 = true, // Testnet4 a different difficulty adjustment rule
```
lontivero, line 68: This is for TestNet4 and only for TestNet4. TestNet5 proposal is different so, I copied the name for bitcoin core but it is a flags that makes no much sense because it is true for only one very specific network.

PR #1332, opened 2026-09-06
Optimize `Encoders.Hex.EncodeData(ReadOnlySpan<byte> data)`

Carti-it:
By default

```cs
public virtual string EncodeData(ReadOnlySpan<byte> data)
{
	return this.EncodeData(data.ToArray());
}
```

is called. However, this allocates an array so it beats any improvement that `ReadOnlySpan<byte>` provides.

This PR adds:

```cs
#if NET8_0_OR_GREATER
		public override string EncodeData(ReadOnlySpan<byte> data)
		{
			// TODO: Convert.ToHexStringLower (.NET 9+) once available.
			return Convert.ToHexString(data).ToLowerInvariant();
		}
#endif
```

to avoid copying span data.

### DIFF
--- a/NBitcoin/DataEncoders/HexEncoder.cs
+++ b/NBitcoin/DataEncoders/HexEncoder.cs
@@ -52,6 +52,14 @@ namespace NBitcoin.DataEncoders
 		private static readonly string[] HexTbl = Enumerable.Range(0, 256).Select(v => v.ToString("x2")).ToArray();
 #endif
 
+#if NET8_0_OR_GREATER
+		public override string EncodeData(ReadOnlySpan<byte> data)
+		{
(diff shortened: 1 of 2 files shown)


NicolasDorier:
I believe we should start shipping NBitcoin version for .NET10 and just use the `ToHexStringLower` there.
Can you do this? Also, you will need to update the CI so that we replace .NET8.0 tests with .NET10.0.

Carti-it:
> I believe we should start shipping NBitcoin version for .NET10 and just use the `ToHexStringLower` there.
> Can you do this? Also, you will need to update the CI so that we replace .NET8.0 tests with .NET10.0.

Yes, I have already done it locally. I will try to publish it today or tomorrow.

Carti-it:
I opened #1333

PR #1333, opened 2026-09-07
Migrate to .NET 10 from .NET 8

Carti-it:
One thing that I cannot really explain is this:

### master (net8.0)

```cs
cd NBitcoin.Bench
dotnet run -f net8.0 -c Release -- --runtimes net8.0 --filter *HexBench*
```

| Method             | hexString            | Mean        | Error     | StdDev     | Gen0   | Allocated |
|------------------- |--------------------- |------------:|----------:|-----------:|-------:|----------:|
| DecodeData         | ?                    |    25.11 ns |  0.434 ns |   0.385 ns | 0.0089 |      56 B |
| DecodeDataLongSpan | ?                    | 4,030.49 ns | 79.167 ns | 102.940 ns | 1.6289 |   10264 B |
| EncodeData         | ?                    |    64.81 ns |  1.324 ns |   2.583 ns | 0.0485 |     304 B |
| DecodeDataSpan     | 00000(...)fffff [64] |    32.56 ns |  0.679 ns |   1.171 ns | 0.0089 |      56 B |
| DecodeDataSpan     | 0000(...)ffff [128]  |    47.32 ns |  0.967 ns |   1.588 ns | 0.0140 |      88 B |
| DecodeDataSpan     | 0000(...)ffff [192]  |    64.95 ns |  1.325 ns |   3.780 ns | 0.0191 |     120 B |

### PR (net10.0)

```cs
cd NBitcoin.Bench
dotnet run -f net10.0 -c Release -- --runtimes net10.0 --filter *HexBench*
```

| Method             | hexString            | Mean        | Error      | StdDev     | Gen0   | Allocated |
|------------------- |--------------------- |------------:|-----------:|-----------:|-------:|----------:|
| DecodeData         | ?                    |    40.50 ns |   0.704 ns |   0.659 ns | 0.0089 |      56 B |
| DecodeDataLongSpan | ?                    | 9,580.75 ns | 166.571 ns | 155.811 ns |      - |         - |
| EncodeData         | ?                    |    17.18 ns |   0.276 ns |   0.258 ns | 0.0242 |     152 B |
| DecodeDataSpan     | 00000(...)fffff [64] |    35.94 ns |   0.402 ns |   0.356 ns |      - |         - |
| DecodeDataSpan     | 0000(...)ffff [128]  |    64.46 ns |   1.107 ns |   0.981 ns |      - |         - |
| DecodeDataSpan     | 0000(...)ffff [192]  |    99.97 ns |   1.786 ns |   1.671 ns |      - |         - |

(This was measured with the latest BenchmarkDotNet version ([v0.15.8](https://github.com/dotnet/BenchmarkDotNet/releases/tag/v0.15.8)) and I did the same for .NET 8. So that's not the reason.)

Decoding HEX on .NET 10 seems to be slower than on .NET 8. I think that all other perf improvements of .NET 10 outweighs this one but still I don't really know what the issue might be. I kept playing with it for a while but I don't know. 

Perhaps 

* it's a really a genuine regression in .NET 10, 
* or it needs to be reproduced on a different machine than mine
* or perhaps something like `dotnet run -f net8.0 -c Release -- --runtimes net10.0,net8.0 --filter *HexBench*` would be more fair comparison. 

I don't know.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -6,47 +6,47 @@ env:
   DOTNET_CLI_TELEMETRY_OPTOUT: 'true'
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  net80:
-    name: '8.0'
+  net100:
+    name: '10.0'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - run: ./Build/CI/install-env.sh
         env:
           DOTNET_INSTALL_DIR: /usr/share/dotnet
-          ADDITIONAL_RUNTIME: 8.0.25
+          ADDITIONAL_RUNTIME: 10.0.11
       - run: ./Build/CI/tests.sh
         env:
           BUILD_ARGS: /p:AdditionalDefineConstants=SECP256K1_VERIFY
-          Framework: net8.0
-  dotnetcore80standard20:
-    name: '8.0 with netstandard2.0'
+          Framework: net10.0
+  dotnetcore100standard20:
+    name: '10.0 with netstandard2.0'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - run: ./Build/CI/install-env.sh
         env:
           DOTNET_INSTALL_DIR: /usr/share/dotnet
-          ADDITIONAL_RUNTIME: 8.0.25
+          ADDITIONAL_RUNTIME: 10.0.11
       - run: ./Build/CI/tests.sh
         env:
           BUILD_ARGS: /p:TargetFrameworkOverride=netstandard2.0
-          Framework: net8.0
-  dotnetcore80macos:
-    name: '8.0 on Mac-OS'
+          Framework: net10.0
+  dotnetcore100macos:
+    name: '10.0 on Mac-OS'
     runs-on: macOS-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - run: ./Build/CI/install-env.sh
         env:
-          ADDITIONAL_RUNTIME: 8.0.25
+          ADDITIONAL_RUNTIME: 10.0.11
       - run: ./Build/CI/tests.sh
         env:
-          Framework: net8.0
-  dotnetcore80winfx:
-    name: '8.0 on Windows NetFramework472'
+          Framework: net10.0
+  dotnetcore100winfx:
+    name: '10.0 on Windows NetFramework472'
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - run: dotnet clean -c Release ./NBitcoin.Tests/NBitcoin.Tests.csproj && dotnet nuget locals all --clear
       - run: dotnet test -c Release -v n ./NBitcoin.Tests/NBitcoin.Tests.csproj --filter "RestClient=RestClient|RPCClient=RPCClient|Protocol=Protocol|Core=Core|UnitTest=UnitTest|Altcoins=Altcoins|PropertyTest=PropertyTest" -p:ParallelizeTestCollections=false -f net472
--- a/NBitcoin.Bench/NBitcoin.Bench.csproj
+++ b/NBitcoin.Bench/NBitcoin.Bench.csproj
@@ -2,14 +2,14 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net8.0</TargetFrameworks>
+    <TargetFrameworks>net10.0</TargetFrameworks>
     <DebugType>pdbonly</DebugType>
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="BenchmarkDotNet" Version="0.14.0" />
-    <PackageReference Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.14.0" />
+    <PackageReference Include="BenchmarkDotNet" Version="0.15.8" />
+    <PackageReference Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.15.8" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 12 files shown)


NBitcoin/Protocol/AddressManager.cs
```diff
@@ -334,9 +334,9 @@ public static AddressManager LoadPeerFile(string filePath, Network expectedNetwo
 			using (var fs = File.Open(filePath, FileMode.Open, FileAccess.Read))
 			{
 				data = new byte[fs.Length - 32];
-				fs.Read(data, 0, data.Length);
```
Carti-it, line 337: This reports an analyzer warning because it can return after a partial read.

NBitcoin/NBitcoin.csproj
```diff
@@ -20,7 +20,7 @@
 		<EmbedUntrackedSources>true</EmbedUntrackedSources>
 	</PropertyGroup>
 	<PropertyGroup>
-    <TargetFrameworks>net472;netstandard2.1;netstandard2.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net10.0;net472;netstandard2.1;netstandard2.0</TargetFrameworks>
```
Carti-it, line 23: `net10.0` TFM was moved everywhere to be the first entry in `<TargetFrameworsk>` to unify it in code.

NBitcoin/PriorityQueue.cs
```diff
@@ -1,4 +1,4 @@
-#if !NET6_0_OR_GREATER && !NO_TUPLE
+#if !NET10_0_OR_GREATER && !NO_TUPLE
```
Carti-it, line 1: Easier to find in the future

.github/workflows/CI.yml
```diff
@@ -6,47 +6,47 @@ env:
   DOTNET_CLI_TELEMETRY_OPTOUT: 'true'
```
Carti-it, line 6: Done based on https://github.com/MetacoSA/NBitcoin/commit/fa4e67701a193c92fe08f403e4056f022c13a393

NBitcoin.Bench/NBitcoin.Bench.csproj
```diff
@@ -2,14 +2,14 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net8.0</TargetFrameworks>
+    <TargetFrameworks>net10.0</TargetFrameworks>
```
Carti-it, line 5: Done based on https://github.com/MetacoSA/NBitcoin/commit/dc5118ac65f2517c2d81fbc4c8a1efd9b4b90caa

NBitcoin.Bench/NBitcoin.Bench.csproj
```diff
@@ -2,14 +2,14 @@
...
-    <TargetFrameworks>net8.0</TargetFrameworks>
+    <TargetFrameworks>net10.0</TargetFrameworks>
     <DebugType>pdbonly</DebugType>
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="BenchmarkDotNet" Version="0.14.0" />
-    <PackageReference Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.14.0" />
+    <PackageReference Include="BenchmarkDotNet" Version="0.15.8" />
```
Carti-it, line 11: Required for .NET 10 support

NBitcoin/BIP32/ExtPubKey.cs
```diff
@@ -112,8 +112,6 @@ public ExtPubKey(byte[] bytes, int offset, int length)
 		/// </summary>
 		public ExtPubKey(ReadOnlySpan<byte> bytes)
 		{
-			if (bytes == null)
```
Carti-it, line 115: Warning: span cannot be null

NBitcoin/DataEncoders/HexEncoder.cs
```diff
@@ -52,14 +52,20 @@ public class HexEncoder : DataEncoder
 		private static readonly string[] HexTbl = Enumerable.Range(0, 256).Select(v => v.ToString("x2")).ToArray();
 #endif
 
+#if NET10_0_OR_GREATER
+		public override string EncodeData(ReadOnlySpan<byte> data)
```
Carti-it, line 56: Added to avoid `byte[]` allocations.

NBitcoin.Tests/NBitcoin.Tests.csproj
```diff
@@ -19,22 +19,22 @@
...
 		</PackageReference>
-		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
+		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.9.0" />
 		<ProjectReference Include="..\NBitcoin\NBitcoin.csproj" />
 		<ProjectReference Include="..\NBitcoin.Altcoins\NBitcoin.Altcoins.csproj" />
 		<ProjectReference Include="..\NBitcoin.TestFramework\NBitcoin.TestFramework.csproj" />
 		<PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
-		<PackageReference Include="xunit" Version="2.9.2" />
-		<PackageReference Include="xunit.runner.visualstudio" Version="2.4.1">
+		<PackageReference Include="xunit" Version="2.9.3" />
```
Carti-it, line 36: `xunit` package is [deprecated](https://www.nuget.org/packages/xunit).

https://github.com/WalletWasabi/WalletWasabi/pull/14864 - upgrade to xUnit v3 is not that hard

NicolasDorier:
@Carti-it I find it weird that 8.0 was faster.
Did you try just bumping 10.0 without the call to `Convert.ToHexStringLower`? Might it be the code path isn't used?

Carti-it:
> @Carti-it I find it weird that 8.0 was faster.

Yes. It's unexpected. I would expect better times, or at worst the same times. Certainly not _worse_.

> Did you try just bumping 10.0 without the call to `Convert.ToHexStringLower`?

Yes. It was also slower. About the same as now.

> Might it be the code path isn't used?

I run tests with the debugger on and the .NET 10 code paths get hit. So I don't think this is the issue.

Carti-it:
@NicolasDorier Perhaps you can try to measure it on your machine.

Carti-it:
https://github.com/dotnet/runtime/discussions/133350

NicolasDorier:
that was fast, even pinpointed on dotnet/runtime wow

PR #1338, opened 2026-09-18
`GetBlockVerbosity`: Add missing value 3

Carti-it:
This verbosity 3 option was introduced in Bitcoin Core 23.0 and its [release notes](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-23.0.md#updated-rpcs) mentions:

> The getblock RPC command now supports verbosity level 3 containing transaction inputs' prevout information. The existing /rest/block/ REST endpoint is modified to contain this information too. Every vin field will contain an additional prevout subfield describing the spent output. prevout contains the following keys:
>
> * generated - true if the spent coins was a coinbase.
> * height
> * value
> * scriptPubKey

RPC documentation: https://bitcoincore.org/en/doc/31.0.0/rpc/blockchain/getblock/

edit: My goal is to add support for verbosity level 3 here: https://github.com/MetacoSA/NBitcoin/blob/7b238c9a972be6d2aa6752716c01df4a7364573e/NBitcoin/RPC/RPCClient.cs#L1507.


Carti-it:
Closing in favor of #1339 which adds support for verbosity 3 and contains this commit.

